PR #5, opened 2015-03-11
Correct copyright notice

TheDoubleK:


LICENCE
```diff
@@ -1,6 +1,4 @@
-The MIT License (MIT)
-
-Copyright (c) 2015 GOV.UK
+Copyright (C) 2015 Crown Copyright (Government Digital Service)
```
tombye: Think the `(C)` should be `(c)` here (according to usage across other GDS projects).

TheDoubleK: I'm not sure it makes a lot of difference as either is supposed to be a substitute for the copyright sign.

Whitehall has a capital C, which is where I copied this from: https://github.com/alphagov/whitehall/blob/master/LICENCE

PR #6, opened 2015-04-07
Changes for elasticbeanstalk deployment

allait:
Adds build and run_app scripts, .ebextensions and environment variable names matching the common convention

### DIFF
--- a/.ebextensions/packages.config
+++ b/.ebextensions/packages.config
@@ -1,0 +1,3 @@
+packages:
+  yum:
+    libffi-devel: []
--- a/application.py
+++ b/application.py
@@ -4,7 +4,9 @@ import os
 from app import create_app
 from flask.ext.script import Manager, Server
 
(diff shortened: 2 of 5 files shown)


quis:
This conflicts with the changes that @TheDoubleK has made in #4, specifically around where we handle exceptions when environment variables aren't set. And it's different again from [how the admin app does it](https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/master/app/main/helpers/service.py). Would be good to have some consistency or sharing of code here.

TheDoubleK:
Agreed that things should be done the same way across all three frontend apps.  Is it something we could pull into a shared library?  (I have a feeling that might be ugly)

allait:
@quis yeah, we didn't have supplier-frontend in fourth wall, so I completely missed the existing PRs. I've merged them now and rebased without conflicting changes.

PR #12, opened 2015-04-24
Fix static path to match the /suppliers URL prefix

allait:


robyoung:
Sorry, I can't merge this because my micro pause dialog is blocking the button.

PR #14, opened 2015-04-27
Adding flask login

minglis:
This pull request is to complete a number of pivotal stories:

https://www.pivotaltracker.com/story/show/92213656

https://www.pivotaltracker.com/story/show/92213744

It provides log in functionality for the supplier application based on the API auth features.

**Flask-Login**

https://flask-login.readthedocs.org/en/latest/

Uses Flask-Login for the log in functionality - this uses existing flask session features and wraps it in a number of convenience methods to handle log in, log out, path protection and so on.

Currently the URLs in the "services" package are protected - this is the dashboard page and the service page. All other pages are open.

We will need to set up some beanstalk properties in production for KEYs, domains and so .

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -7,10 +7,12 @@ from flask._compat import string_types
 
 from .main import main as main_blueprint
 from .status import status as status_blueprint
+from .flask_api_client.api_client import ApiClient
+
+api_client = ApiClient()
 
 
 def create_app(config_name):
-
     application = Flask(__name__,
                         static_folder='static/',
                         static_url_path=config[config_name].STATIC_URL_PATH)
@@ -28,6 +30,8 @@ def create_app(config_name):
     main_blueprint.config = {
         'BASE_TEMPLATE_DATA': application.config['BASE_TEMPLATE_DATA']
     }
+
+    api_client.init_app(application)
 
     return application
 
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -5,8 +5,6 @@ from flask import Flask
 from config import config
 from flask._compat import string_types
 
-from .main import main as main_blueprint
-from .status import status as status_blueprint
 from .flask_api_client.api_client import ApiClient
 
 api_client = ApiClient()
@@ -17,12 +15,17 @@ def create_app(config_name):
                         static_folder='static/',
                         static_url_path=config[config_name].STATIC_URL_PATH)
 
+    from .main import main as main_blueprint
+    from .status import status as status_blueprint
+
+    application.config.from_object(config[config_name])
+
     for name in config_attrs(config[config_name]):
         if name in os.environ:
             application.config[name] = convert_to_boolean(os.environ[name])
 
-    application.config.from_object(config[config_name])
     config[config_name].init_app(application)
+    api_client.init_app(application)
 
     application.register_blueprint(status_blueprint)
     application.register_blueprint(main_blueprint,
@@ -30,8 +33,6 @@ def create_app(config_name):
     main_blueprint.config = {
         'BASE_TEMPLATE_DATA': application.config['BASE_TEMPLATE_DATA']
     }
-
-    api_client.init_app(application)
 
     return application
 
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -2,12 +2,13 @@ import os
 import re
 
 from flask import Flask
+from flask_login import LoginManager
 from config import config
 from flask._compat import string_types
-
 from .flask_api_client.api_client import ApiClient
 
 api_client = ApiClient()
+login_manager = LoginManager()
 
 
 def create_app(config_name):
@@ -26,7 +27,8 @@ def create_app(config_name):
 
     config[config_name].init_app(application)
     api_client.init_app(application)
-
+    login_manager.init_app(application)
+    login_manager.login_view = '/suppliers/login'
     application.register_blueprint(status_blueprint)
     application.register_blueprint(main_blueprint,
                                    url_prefix='/suppliers')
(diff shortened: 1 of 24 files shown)


app/flask_api_client/api_client.py
```diff
@@ -0,0 +1,72 @@
...
+            headers=self.headers()
+        )
+        if res.status_code is 200:
+            return self.user_json_to_user(res.json())
+        elif res.status_code is 400:
+            # TODO log bad request
+            return None
+        else:
+            # TODO log error
+            return None
```
robyoung: We do have logging in, logs could be added with `current_app.logger.error("my error %s", "message")`.

app/flask_api_client/api_client.py
```diff
@@ -0,0 +1,72 @@
...
+        )
+        if res.status_code is 200:
+            return self.user_json_to_user(res.json())
+        elif res.status_code is 400:
+            # TODO log bad request
+            return None
+        elif res.status_code is 403:
+            # TODO log unauthorized
+            return None
+        elif res.status_code is 404:
```
robyoung: If we're returning the same thing could these be collapsed and then add a generic log message?

```python
elif res.status_code in [400, 403, 404]:
  current_app.logger.error("failed users auth [%s]", res.status_code)
else:
```

robyoung: Also, do we get the error message in the response?

app/flask_api_client/api_client.py
```diff
@@ -0,0 +1,72 @@
...
+            # TODO log not found
+            return None
+        else:
+            # TODO log error
+            return None
+
+    @staticmethod
+    def user_json_to_user(user_json):
```
robyoung: I think the model class should either be part of this module (part of the API interface) or not connected with this module at all. It makes this client library much less portable.

app/main/errors.py
```diff
@@ -1,5 +1,5 @@
 from flask import render_template
-from . import main
+from app.main import main
```
robyoung: should be a relative import

app/flask_api_client/api_client.py
```diff
@@ -0,0 +1,72 @@
+import requests
+from flask import json
+from app.model import User
```
robyoung: Should be a relative import. Also, in other modules we are separating imports from 3rd party libraries from application imports with a blank line.

app/flask_api_client/api_client.py
robyoung: Either way I think the from / to JSON functions should be on that object. eg.
```python
user  = User.from_json(user_json)
user.to_json()
```

app/main/login.py
```diff
@@ -1,20 +1,52 @@
...
+    form = LoginForm()
+    template_data = main.config['BASE_TEMPLATE_DATA']
+    if form.validate_on_submit():
+
+        user = api_client.users_auth(
+            form.email_address.data,
+            form.password.data)
+
+        if not user:
+            flash(
```
TheDoubleK: We can add categories to `flash` to distinguish between success and error messages. e.g.: 
```
flash('Sorry, we couldn't find a user with that username and password.', 'error')
```
and then in the template use:
```
{% with messages = get_flashed_messages(with_categories=true) %}
```

app/flask_api_client/api_client.py
```diff
@@ -0,0 +1,72 @@
...
+        elif res.status_code is 404:
+            # TODO log not found
+            return None
+        else:
+            # TODO log error
+            return None
+
+    @staticmethod
+    def user_json_to_user(user_json):
+        return User(
```
TheDoubleK: I might have missed something really obvious because my head is quite full of cold, but I don't see why we need our own `User` model at all here? Once the login is confirmed and we set the user as logged-in in the session then it looks like Flask has it's `current_user` method that should give us what we need without having to implement our own `User` class?  Or does Flask depend on us defining our own `User` for its auth stuff to work?  Apologies if I've completely missed the point here.

app/__init__.py
```diff
@@ -35,6 +43,11 @@ def create_app(config_name):
     return application
 
 
+@login_manager.user_loader
+def load_user(user_id):
```
TheDoubleK: Something not 100% clear to me from the docs is when this `load_user` method is called?  Is it only for initial login and cookie setting, or for every page that requires authentication?  If the latter then we'll be calling the auth API for every page load.

minglis: This is true - swithered over this for a while. Argument is any changes to the user should log that user out immediatly. Which wouldn't be possible with a cookie cache. We can optimise this by either:
- adding a timestamp into the session and checking on that expiry - say every 10mins not every request, This logic would be in the load_user method
- basing the user_id of the functionliaty of alternative tokens - problem there is you need to reload the user from the token, so it would need to be two way encryption. But we could use the email and the last password changed date or something.

Was thinking could be it;s own story

app/main/login.py
minglis: Nice, did not know.

app/flask_api_client/api_client.py
minglis: Not sure how good our error messaging is right now. I split these as a placeholder for myself to handle the conditions when logging was in. Happy to collapse.

app/__init__.py
TheDoubleK: Ok, happy to go with this for now then, supplier user traffic will generally be pretty low so the API calls probably not too much of an issue.  We'll need some sort of solution by the last week of SSP though!

app/flask_api_client/api_client.py
minglis: Rob is picking this up as part of move to utils package for shared code.

app/main/errors.py
minglis: done

app/flask_api_client/api_client.py
minglis: done

PR #16, opened 2015-04-29
Do not activate venv if already in a venv

robyoung:


TheDoubleK:
I already did this in my branch, but doesn't hurt to get it in as its own thing.

PR #17, opened 2015-04-30
Forgotten password emails

TheDoubleK:
This fleshes out the "forgotten password" route - it now sends emails and updates passwords for real. See this story: https://www.pivotaltracker.com/story/show/92213710
Also: https://www.pivotaltracker.com/story/show/92215856
Also: https://www.pivotaltracker.com/story/show/92216072
Also: https://www.pivotaltracker.com/story/show/92215834
Also: https://www.pivotaltracker.com/story/show/92213688
Also: https://www.pivotaltracker.com/story/show/92213724

You will need a Mandrill key if you want to see it send real emails (see me if you want it). The app fails quietly if you don't have a valid API key set (it logs a Mandrill error but doesn't break the app), and you can still test the whole flow without the risk of emailing people you didn't mean to as follows:
* start up the data API locally (hopefully you have some users in your database already, otherwise run the import_users script)
* start the app
* visit http://localhost:5003/suppliers/login and hit "forgotten password"
* type in an email address (preferably one that that exists in your local users table - you can try out different behaviour for supplier vs buyer accounts) and hit the button
* check your app logging for a line that starts "Generated reset URL:" - here you can find the URL that would have been sent in an email if you had a valid Mandrill key set.
* Paste the full URL into your browser (this is like clicking the link in the email that would have been sent)
* Reset the password for the account
* Log in to the app using the email address and new password. (Or not if you used a buyer account.)


robyoung:
Could we add the mandril key to the credentials repository?

TheDoubleK:
I asked this in Slack the other day - I'm not sure if Credentials might be too public?

app/main/helpers/email.py
```diff
@@ -0,0 +1,63 @@
...
+
+with open("./email_templates/forgotten_password_email.html", "r") as f:
+    forgot_password_email_body = f.read()
+
+
+def send_password_email(user_id, email_address):
+    try:
+        mandrill_client = mandrill.Mandrill(main.config['MANDRILL_API_KEY'])
+        url = generate_reset_url(user_id, email_address)
+        body = forgot_password_email_body.format(url)
```
robyoung: This feels like it should be using Flask's template rendering. http://flask.pocoo.org/docs/0.10/quickstart/#rendering-templates

TheDoubleK: I did think about this, but it didn't feel to me like the email template belonged in the same place as our page templates so I deliberately didn't bundle it in with the page templates.

app/main/helpers/email.py
```diff
@@ -0,0 +1,63 @@
...
+                           'type': 'to'}],
+                   'important': False,
+                   'track_opens': None,
+                   'track_clicks': None,
+                   'auto_text': True,
+                   'tags': ['password-resets'],
+                   'headers': {'Reply-To': main.config['FORGOT_PASSWORD_EMAIL_FROM']},  # noqa
+                   'recipient_metadata': [
+                       {'rcpt': email_address,
+                        'values': {'user_id': 123456}}]
```
robyoung: What's this number? Should it be configurable?

app/main/helpers/email.py
```diff
@@ -0,0 +1,63 @@
...
+        }
+        result = mandrill_client.messages.send(message=message, async=False,
+                                               ip_pool='Main Pool')
+        '''
+        [{'_id': 'abc123abc123abc123abc123abc123',
+          'email': 'recipient.email@example.com',
+          'reject_reason': 'hard-bounce',
+          'status': 'sent'}]
+        '''
```
robyoung: What does this string relate to?

app/main/helpers/email.py
robyoung: I think putting them in a separate subdirectory would be separate enough. If we want to keep them completely separate we could use Jinja2 directly. I think using the same templating language throughout makes template authoring simpler.

app/main/helpers/email.py
TheDoubleK: It's an example of a result from the mandrill client.  We probably don't need it in there.

app/main/helpers/email.py
TheDoubleK: Good spot - this was a placeholder, I meant to replace it with user id from the API, although I'm not sure we need it all really.

app/main/helpers/email.py
TheDoubleK: OK I'll have a play and see how it looks.

TheDoubleK: Rendering Flask templates creates an HTTP response, but all we need here is a string to pass in to the Mandrill client.

TheDoubleK:
Latest version uses Flask template rendering and inserts user_id where it should have been before.

app/main/helpers/email.py
```diff
@@ -0,0 +1,51 @@
...
+        # Mandrill errors are thrown as exceptions
+        current_app.logger.error("A mandrill error occurred: %s", e)
+        return
+    current_app.logger.info("Sent password email: %s", result)
+
+
+def generate_reset_url(user_id, email_address):
+    encoded_string = str(user_id) + " " + email_address
+    ts = URLSafeTimedSerializer(main.config["SECRET_KEY"])
+    token = ts.dumps(encoded_string, salt=main.config["SECRET_SALT"])
```
allait: According to http://pythonhosted.org/itsdangerous/#the-salt salt acts more like a namespace than a hash secret. It should make it impossible to reuse the "reset password" code for other purposes (in case the encoded data is the same), so I think we should rename `SECRET_SALT` into e.g. `RESET_PASSWORD_SALT`

app/main/helpers/email.py
```diff
@@ -0,0 +1,51 @@
...
+                                               ip_pool='Main Pool')
+    except mandrill.Error as e:
+        # Mandrill errors are thrown as exceptions
+        current_app.logger.error("A mandrill error occurred: %s", e)
+        return
+    current_app.logger.info("Sent password email: %s", result)
+
+
+def generate_reset_url(user_id, email_address):
+    encoded_string = str(user_id) + " " + email_address
```
allait: I think `itsdangerous` can work with lists/dictionaries too, so we should be able to do `.dumps({"user": user_id, "email": email})`

app/main/login.py
```diff
@@ -60,24 +63,66 @@ def forgotten_password():
...
+        return Response("Error: %s" % str(e), 500)
+
+
+@main.route('/change-password/<token>', methods=["GET"])
+def change_password(token):
+    try:
+        decoded = email.decode_email(token)
+        user_id = decoded.split()[0]
+        email_address = decoded.split()[1]
+    except Exception as e:
```
allait: Should catch the specific exception we're interested in

app/main/login.py
```diff
@@ -60,24 +63,66 @@ def forgotten_password():
...
+            current_app.logger.info(
+                "Password reset request for invalid supplier email address %s",
+                email_address
+            )
+            # TODO: Add to count in "forgotten password - invalid" metric
+
+        flash('If that Digital Marketplace supplier account exists, you will '
+              'be sent an email containing a link to reset your password.')
+        return redirect(url_for('.forgotten_password'))
+    except Exception as e:
```
allait: Why do we need this? Flask will catch any uncaught exceptions for us and we'll end up with the same 500.
And I think the 500 response page is coming from EB apache right now anyway.

TheDoubleK: Good point - I'll drop the try-catch

app/main/helpers/email.py
TheDoubleK: Nice - yes it does.

scripts/run_app.sh
```diff
@@ -9,6 +9,9 @@ fi
 export DM_SUPPLIER_FRONTEND_API_AUTH_TOKEN=${DM_SUPPLIER_FRONTEND_API_AUTH_TOKEN:=myToken}
 export DM_API_URL=${DM_API_URL:=http://localhost:5000}
 export DM_API_AUTH_TOKEN=${DM_API_AUTH_TOKEN:=myToken}
+export DM_MANDRILL_API_KEY=${DM_MANDRILL_API_KEY:=not_a_real_key}
```
allait: Do we need to add the new env vars to the digitalmarketplace-aws stacks.yml before merging this or will this still work without them?

scripts/run_app.sh
```diff
@@ -9,6 +9,9 @@ fi
...
+export DM_PASSWORD_SECRET_KEY=${DM_PASSWORD_SECRET_KEY:=verySecretKey}
+export DM_DM_SECRET_SALT=${DM_SECRET_SALT:=itsSalty}
```
allait: DM_DM_ is probably a typo

TheDoubleK: Yep - think this is fixed already.

scripts/run_app.sh
TheDoubleK: I'll do that - hold off merging for now, just to be sure.

app/main/login.py
```diff
@@ -60,24 +64,69 @@ def forgotten_password():
...
+        email_address = decoded["email"]
+    except SignatureExpired:
+        current_app.logger.info("Password reset attempt with expired token.")
+        flash('The token supplied has expired. Password reset links are only'
+              ' valid for 24 hours. You can generate a new one using the form'
+              ' below.', 'error')
+        return redirect(url_for('.forgotten_password'))
+    except BadSignature as e:
+        current_app.logger.info("Error changing password: %s", e)
+        flash('The token supplied was invalid. You can generate a new one'
```
minglis: Not sure about having flash messages hard coded in python code. Do we need a messages file or some other model?

TheDoubleK:
Pull-request in for adding the environment variables - this should merge before this one:
https://github.com/alphagov/digitalmarketplace-aws/pull/73

app/main/login.py
TheDoubleK: OK I'll have a look at a neat way to do this. I guess config vars is equally ugly.

TheDoubleK:
Fingers crossed for Travis... this should be good to go now.

PR #18, opened 2015-04-30
Updated `_status` endpoint

pcraig3:
Successful response:
```json
{
  "api_status": {
    "db_version": "10_adding_supplier_to_user_table",
    "status": "ok",
    "version": null
  },
  "status": "ok",
  "version": null
}
```

API not connected to DB
```json
{
  "api_status": {
    "message": "Error connecting to database",
    "status": "error",
    "version": null
  },
  "message": "Error connecting to the (Data) API.",
  "status": "error",
  "version": null
}
```

No response from API
```json
{
  "api_status": null,
  "message": "Error connecting to the (Data) API.",
  "status": "error",
  "version": null
}
```


app/status/views.py
```diff
@@ -2,8 +2,38 @@
...
 
 @status.route('/_status')
 def status():
-    return jsonify(status="ok", version=utils.get_version_label())
+
+    api_response = utils.return_response_from_api_status_call(
+        api_client.status
+    )
+
+    apis_wot_got_errors = []
```
TheDoubleK: Please change the variable name

app/status/views.py
```diff
@@ -2,8 +2,38 @@
...
+
+    if api_response is None or api_response.status_code is not 200:
```
TheDoubleK: Use == for int comparison.

tests/app/status/test_status.py
```diff
@@ -0,0 +1,66 @@
...
+
+import mock
+from nose.tools import assert_equal, assert_in
+
+
+class TestStatus(BaseApplicationTest):
+    def setup(self):
+        super(TestStatus, self).setup()
+
+        self._api_response = mock.patch(
```
TheDoubleK: These tests work fine as they are and I'm not saying we should re-do them or anything, but it's worth taking a look at the requests_mock package used here: https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/tests/app/flask_api_client/test_api_client.py
I think it's a bit easier and clearer than building up the Response objects as you have here - worth bearing in mind for the future. (Or have a play with it here it the urge takes you).

pcraig3:
Added `dmutils.apclient` in latest commit; means we now have two API clients in the app.

app/flask_api_client/api_client.py
```diff
@@ -22,6 +22,11 @@ def headers(self):
             "Authorization": "Bearer {}".format(self.token)
         }
 
+    def status(self):
```
allait: Don't need this anymore

PR #28, opened 2015-05-07
Make `url_for` use the correct scheme (http/https)

TheDoubleK:
Addresses this bug: https://www.pivotaltracker.com/story/show/94091430

By adding ProxyFix the headers are correctly passed through to the app, so that it will know
if the original request was from an http or https context.

This is the same way it is done in the data API project for generating URLs.


TheDoubleK:
I believe this will work, but can't test for sure until deployed to an https environment.

PR #29, opened 2015-05-07
Display and edit status of a service

quis:
This adds examples for what the edit service page should look like if:
- The service is visible on the Marketplace (suppliers can change its status)
- Nicole has removed the service

It shows both of these things on one page, where in reality it would be either/or.

Removed | Visible or hidden
--- | ---
| image  | image |

image

It also adds the service status to the supplier's page (formerly the 'dashboard'):
image


TheDoubleK:
I don't know if we're planning on displaying expired services at all?  As far as I understood it we're just keeping them for future reference if needed to look up by us for legal reasons (e.g. to check T&Cs for ongoing contracts from expired frameworks, as we had recently).  As far as the frontend apps are concerned though I don't think they exist - there is no API endpoint that will return an expired service.

ralph-hawkins:
Using ‘edit’ to view a service that has been removed (and can't be put back by the user) sets up the wrong expectation. We could use a different term when its been removed, e.g. ‘details’ for removed services (keep ‘edit’ when you can change the status of a service).

ralph-hawkins:
I think it would be clearer if it said ‘*Name of Service* has been removed'. Possibly with a reason or by whom eg. ‘*Name of Service* has been removed by Admin because it contained the word ‘Agile’’. And date it was removed also.

quis:
@ralph-hawkins We can't do date/reason at the moment. And I don't think _Name of service_ makes sense when that line's going to appear right underneath the service name.

app/assets/javascripts/_selection-buttons.js
```diff
@@ -0,0 +1,16 @@
+(function(GOVUK, GDM) {
+
+  GDM.selectionButtons = function() {
```
TheDoubleK: Not for this PR necessarily, but would this kind of script sit happily in the frontend toolkit?

TheDoubleK:
I think 'enabled' services need bolding and some colour - the same as Published and Disabled. Maybe some shade of amber, like traffic lights?

Although Disabled should be a rare thing for suppliers to see - so maybe user red/green for the main published/unpublished distinction, and do something else for Disabled?

I think the term 'enabled' will be very misleading for suppliers - we should decide some different terminology for the frontend to display.

screen shot 2015-05-11 at 11 15 51

TheDoubleK:
Happy to merge this now and iterate these things later though, if there is user research input needed to settle things.

app/assets/javascripts/_selection-buttons.js
quis: It's on mine and Tom's backlogs to think about how we do dependencies/modules for all the apps. TL;DR you're right, but we haven't decided what to do about it yet.

quis:
@TheDoubleK addressed that unstyled text in ab19455. I think the plan is to get this stuff working, and then do a database migration so that the correct terminology is used from the API through to the frontend. This three statuses that @superroz and I came up with were:
- Public
- Private
- Removed

ralph-hawkins:
@quis I'm not suggesting  '*name of service* has been removed' appears under the service name but we look at that page including the heading and make it read a bit more easily

pcraig3:
Waiting on this pull request to be merged in `digitalmarketplace-utils`

https://github.com/alphagov/digitalmarketplace-utils/pull/31

pcraig3:
#### A Supplier's Services are listed in the Supplier dashboard.  `Removed` Services have a _Details_ link instead of an _Edit_ Link.

screen shot 2015-05-12 at 15 58 21

***
#### Changing a Service's status is fairly straightforward.  Buttons have been repositioned horizontally.

screen shot 2015-05-12 at 15 58 43

***

#### Trying to sneakily change your Service's status to one that isn't permitted returns an error.

screen shot 2015-05-12 at 15 59 44

***

#### After successfully updating a Service's status, user is returned to the dashboard view and there's a notification near the top confirming the change that was just made.

screen shot 2015-05-12 at 15 59 57

pcraig3:
Sorry for the dumb formatting of that last comment -- the screenshots aren't too clearly separated from GitHub's interface.

quis:
Looking excellent to me now. Would be good to get this incorporated before merging: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/67

TheDoubleK:
I don't think the frontend needs to echo the database values for status? Our DB statuses are for consumption by the apps, not for showing to users.  I'm not sure "Public" vs "Private" will make sense to suppliers - why would they want a Private service?  Published/Unpublished still seems better to me from a supplier point of view, though still not ideal.

Just to check - is this still a WIP or is it ready to go?

app/assets/scss/.service-status-removed
```diff
@@ -0,0 +1,7 @@
+.service-status-removed {
+  color: $red;
+}
+
+.service-status-expired {
```
TheDoubleK: Is this used anywhere now?  Or is expired gone from everywhere else in the frontend?

app/assets/scss/_service-status.scss
```diff
@@ -0,0 +1,23 @@
...
+  color: $red;
+}
+
+.service-status-published,
+.service-status-public {
+  @extend %service-status;
+  color: $button-colour;
+}
+
+.service-status-unpublished,
```
TheDoubleK: This triplet of indecisiveness doesn't look great to me, but I'll hopefully never look at a `.scss` file so probably don't care that much.

app/assets/scss/.service-status-removed
quis: @TheDoubleK No idea what this file is. Removed in a15f5af.

quis:
This is no longer WIP.

TheDoubleK:
I can't get this to run locally - I've got master API running, have `pip install`ed, `npm install`ed, run the frontend build, etc. but when trying to update a service status I get a 400 error at a weird redirect URL e.g: `http://localhost:5003/suppliers/services/%5Bu%275-G4-0149-213%27%5D` - looks like the URL encoding is going wrong somewhere? 
screen shot 2015-05-12 at 17 43 02

Have tried it with a G6 service and that gives the same error at this weird URL (service id=1122334455667788): `http://localhost:5003/suppliers/services/%5B%271122%27%2C%20%273344%27%2C%20%275566%27%2C%20%277788%27%5D`

quis:
@TheDoubleK Fixed in 4450c4f

app/main/services.py
```diff
@@ -16,14 +19,79 @@ def dashboard():
...
+    if status in translate_frontend_to_api.keys():
+        status = translate_frontend_to_api[status]
+    else:
+        return _update_service_status_error(
+            service_id,
+            "Sorry, but '{}' is not a valid status.".format(status)
+        )
+
+    try:
+        updated_service = data_api_client.update_service_status(
```
TheDoubleK: There should be a condition here to stop suppliers from changing the status if the current status is `disabled` - at the moment they could CURL a request and un-disable their disabled services.

app/templates/services/dashboard.html
```diff
@@ -2,58 +2,86 @@
...
-
-
-<table class="summary-item-body">
+  {% if updated_service_id %}
+    <div class="grid-row">
+    <div class="column-one-whole">
+      <div class="banner-success-without-action">
+        <p class="banner-message">
+          <a href="{{ url_for('.update_service_status', service_id=updated_service_id ) }}">{{ updated_service_name }}</a>
+           is now {{ updated_service_status }}
```
TheDoubleK: The value here needs to be mapped to Public/Private - at the moment it is echoing the DB values enabled/published.

TheDoubleK: There should also be a test for this - i.e. that suppliers can't un-disable a disabled service.

app/main/services.py
TheDoubleK: There should also be a test for this - i.e. that suppliers can't un-disable a disabled service.

pcraig3: I don't see how they could enable (or publish) a disabled service based on this logic.  It doesn't allow statuses that aren't in the dictionary to be sent to the server.  

But anyway, writing the test is probably useful.

TheDoubleK:
I've found another bug here - a logged-in supplier can change the status of any service, whether or not the service belongs to them.

To reproduce:
* find a service ID for a service belonging to "Supplier 1" 
* log in to the supplier app as "Supplier 2"
* go to /suppliers/services/<supplier-1-servcie-ID>
* turn off your rival's service!

To fix this the GET and POST endpoints for `/services/<string:service_id>` both need to check that the service ID belongs to the logged-in supplier.  If it doesn't then it should probably return a 404 (but I could be persuaded 403 if you prefer).

app/main/services.py
TheDoubleK: It could - the dictionary here allows a service to be set to 'public' or 'private'.  So if Nicole has set a service to 'removed' because it is in breach of something, the supplier could hit this endpoint to change the status back to 'public' or 'private'.  The dictionary will prevent services from being 'removed', but doesn't stop them being turned back on again.
There needs to be a check on the current status of the service, and if it is removed/disabled then the update should not be done.

TheDoubleK: To be clear - they wouldn't be able to do it via the interface because the form isn't there for disabled services.  But they could curl the POST endpoint with a request that would update the service status.

app/assets/scss/_service-status.scss
pcraig3: I think leave it until a final decision about statuses has been made.

app/main/services.py
pcraig3: Right, thanks for the explanation -- quite a good catch actually.
Guess I'm not as apt to thinking as deviously as you are, Kev.

pcraig3:
Still gonna write some tests, but tried to address your concerns, Kev!

TheDoubleK:
I've added some tests for the admin app update status routes here (in a pull-request) - I'm not saying they're the greatest ever, but might be worth a look at to get ideas for the update status tests on this story. 

https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/admin_deactivate_a_service/tests/app/main/test_views.py

pcraig3:
Finished all my functional tests and rebased off of the master branch so if we like the look of the tests we should be good to go.

app/templates/services/service.html
```diff
@@ -0,0 +1,106 @@
...
+        <fieldset class="question">
+          <legend class="question-heading question-heading-with-hint ">
+            Choose service status
+          </legend>
+          <p class="hint">
+            Private services don't appear in search results and don't have a
+            <acronym title="Uniform resource locator">URL</acronym>
+          </p>
+          <label class="selection-button selection-button-boolean">
+            <input type="radio" name="service_status" id="service_status_published" value="public" {{'checked=checked' if service_data['status'] == 'published' }} />
```
TheDoubleK: Attribute values should be in quotes.  I know it works anyway, but still nice to be consistent.  I'll fix this now and then merge it.

TheDoubleK:
:+1:

PR #31, opened 2015-05-11
Make service link relative

robyoung:
On the dashboard, make the services link relative so that it points to
the correct environment.

Fixes: [94162812](https://www.pivotaltracker.com/story/show/94162812)


quis:
This isn't a fix for the [linked card](https://www.pivotaltracker.com/n/projects/997836/stories/94162812) because that is about the admin app. And is itself a duplicate of [94069990](https://www.pivotaltracker.com/n/projects/997836/stories/94069990) which is currently blocked.

That said, this is a worthwhile change so imma merge it.

PR #33, opened 2015-05-12
Add password rules and tidy up login forms

TheDoubleK:
See: https://www.pivotaltracker.com/story/show/92215106

As well as adding the requirement for passwords between 10 and 50 characters long this PR tidies up the login forms to all use Flask's WTForms.  This means that validation rules can all be set in the Form objects, and removes validation checking of individual rules from the views. It also makes error messaging look consistent between pages.

Both the `main.` qualifier added to the `change_password` url_for call, and the `SERVER_NAME` added to the Test config are required for the call to `url_for` to work in tests.

After discussion with @tombye: Validation errors on a single field are shown in-line only.  Pages with multiple validation errors also get the "Errors on the following fields:" box at the top of the page.


robyoung:
Why do we need an upper limit on password length?

TheDoubleK:
It was specified in the story.  We don't necessarily need one.

app/main/login.py
```diff
@@ -59,27 +59,35 @@ def forgotten_password():
...
+        return redirect(url_for('.forgotten_password'))
     else:
-        message = "login.reset-email.invalid-email: " \
-                  "Password reset request for invalid supplier email %s"
-        current_app.logger.info(message, email_address)
-
-    flash('email_sent')
-    return redirect(url_for('.forgotten_password'))
+        template_data = main.config['BASE_TEMPLATE_DATA']
+        return render_template("auth/forgotten-password.html",
```
robyoung: Do we want to be logging login failure here?

tests/app/main/test_login.py
```diff
@@ -1,24 +1,23 @@
...
+email_empty_error = "Email can not be empty"
+email_invalid_error = "Please enter a valid email address"
+email_sent_message = "If that Digital Marketplace supplier account exists, " \
+                     "you will be sent an email containing a link to reset " \
+                     "your password."
+password_empty_error = "Please enter your password"
+password_invalid_error = "Passwords must be between 10 and 50 characters"
+password_mismatch_error = "The passwords you entered do not match"
+new_password_empty_error = "Please enter a new password"
+new_password_confirm_empty_error = "Please confirm your new password"
```
robyoung: These should probably be all caps.

tests/app/main/test_login.py
```diff
@@ -72,29 +71,130 @@ def test_should_return_a_403_for_invalid_login(self):
...
-            in self.strip_all_whitespace(res.get_data(as_text=True)))
+            self.strip_all_whitespace(password_empty_error)
+            in content)
 
     def test_should_be_validation_error_if_invalid_email(self):
         res = self.client.post("/suppliers/login", data={
             'email_address': 'invalid',
             'password': '1234567890'
         })
+        content = self.strip_all_whitespace(res.get_data(as_text=True))
```
robyoung: Ooh, `as_text` is exactly what I needed earlier. Thanks!

robyoung:
Fair enough about the 50 character limit, I'll raise it with Cath or @minglis.

app/main/login.py
TheDoubleK: Probably.  I haven't put any logging in; I wasn't sure if we wanted to add it in a separate story once we decide exactly what we want in the logs, to ensure consistency of messaging across apps, etc.  I can put something in for now if you reckon best to have it there to be changed later if necessary.

tests/app/main/test_login.py
TheDoubleK: OK will do - I was just modifying the constants already there, which were lower to start with.

app/main/login.py
robyoung: No, that's a fair point. I've created a story for the logging in pivotal and added this as an area to consider https://www.pivotaltracker.com/story/show/94490624

TheDoubleK: I don't think we need to log validation errors - it isn't a login failure here, just a mismatch of new password vs. confirm password, or a too short/too long password.

PR #35, opened 2015-05-15
[#94292928] Integrate custom proxy fix

robyoung:
AWS ELBs set the X-Forwarded-Proto to HTTP if they're configured port
80 -> 80. Because the apps are now internal and served over HTTP we lose
the X-Forwarded-Proto that is sent by Nginx. This change allows the HTTP
protocol to be configured per environment while still making use of
flasks ProxyFix middleware.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ Flask==0.10.1
 Flask-Login==0.2.11
 Flask-Script==2.0.5
 Flask-WTF==0.11
--e git+https://github.com/alphagov/digitalmarketplace-utils.git@0.10.2#egg=digitalmarketplace-utils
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -2,7 +2,7 @@ Flask==0.10.1
 Flask-Login==0.2.11
 Flask-Script==2.0.5
 Flask-WTF==0.11
--e git+https://github.com/alphagov/digitalmarketplace-utils.git@0.10.2#egg=digitalmarketplace-utils
+-e git+https://github.com/alphagov/digitalmarketplace-utils.git@0.11.0#egg=digitalmarketplace-utils
```
TheDoubleK: We're up to 0.12.0 now.

PR #37, opened 2015-05-18
Remove ability for Suppliers to edit status of their Services.

pcraig3:
After talking with Cath and Nicole today, it was felt that giving Suppliers the ability to make their Services 'private' wasn't really valuable.  
The contention was that Suppliers would see the 'Edit' link and -- thinking it would allow them to edit all of the fields -- become disappointed when all they would actually be able to do would be remove Services from public view.

Solution was to:
* Remove 'Edit'  / 'Details' links on the Supplier dashboard
* Disable routes to view/change a Service's status in the app
* Refactor the tests

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/94826380)

***

### New Dashboard

screen shot 2015-05-18 at 15 03 53

### DIFF
--- a/tests/app/main/test_services.py
+++ b/tests/app/main/test_services.py
@@ -126,53 +126,10 @@ class TestDashboardLogin(BaseApplicationTest):
                      '?next=%2Fsuppliers%2F')
 
 
-class TestServicesLogin(BaseApplicationTest):
-    def test_should_show_dashboard_if_logged_in(self):
-        with self.app.test_client():
-            data_api_client.authenticate_user = Mock(
-                return_value=(self.user(123, "email@email.com", 1234, 'name')))
-
-            data_api_client.get_user = Mock(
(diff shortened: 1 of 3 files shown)


tests/app/main/test_services.py
```diff
@@ -126,53 +127,10 @@ def test_should_redirect_to_login_if_not_logged_in(self):
                      '?next=%2Fsuppliers%2F')
 
 
-class TestServicesLogin(BaseApplicationTest):
```
pcraig3: This class was only checking 2 things:
1. That, once logged in, a user could _see_ an 'Edit Service Status' page
2. That, if not logged in, a user would be redirected.

Since there was a lot of overlap, I just absorbed both checks into my `TestSupplierUpdateService` class.

quis:
Spoke to @minglis and @TheDoubleK about whether we should really disable this feature. Going to close this pull request for now until that discussion is complete.

PR #38, opened 2015-05-19
No trailing slashes on URLs

quis:
Duplicates https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/52

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -1,6 +1,6 @@
 import re
 
-from flask import Flask
+from flask import Flask, request, redirect
 from flask_login import LoginManager
 from flask._compat import string_types
 from dmutils import apiclient, logging, config, proxy_fix
@@ -34,6 +34,11 @@ def create_app(config_name):
                                    url_prefix='/suppliers')
     login_manager.login_view = 'main.render_login'
     main_blueprint.config = application.config.copy()
+
+    @application.before_request
+    def remove_trailing_slash():
+        if request.path.endswith('/'):
+            return redirect(request.path[:-1])
 
     return application
 
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -38,7 +38,7 @@ def create_app(config_name):
     @application.before_request
     def remove_trailing_slash():
         if request.path.endswith('/'):
-            return redirect(request.path[:-1])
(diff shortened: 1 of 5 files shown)


app/__init__.py
```diff
@@ -35,6 +35,11 @@ def create_app(config_name):
     login_manager.login_view = 'main.render_login'
     main_blueprint.config = application.config.copy()
 
+    @application.before_request
+    def remove_trailing_slash():
+        if request.path.endswith('/'):
+            return redirect(request.path[:-1])
```
allait: `redirect` defaults to 302, so we should probably add `code=301` for a permanent redirect.
Also, this will lose query parameters. I'm not sure if that's a problem, but e.g. https://preview.development.digitalmarketplace.service.gov.uk/search/?q=email ends up on `/search` with a 404 right now.

quis: The Grails app makes URLs like `https://www.digitalmarketplace.service.gov.uk/search?keyword=cloud`. So I think we're safe. Not ideal for the URL to not redirect; is there any easy way to keep the query parameters?

quis: And I've added the `code=301` in 7958fe5.

app/__init__.py
```diff
@@ -35,6 +35,11 @@ def create_app(config_name):
     login_manager.login_view = 'main.render_login'
     main_blueprint.config = application.config.copy()
 
+    @application.before_request
+    def remove_trailing_slash():
+        if request.path.endswith('/'):
+            return redirect(request.path[:-1], code=301)
```
allait: We could add `'?{}'.format(request.query_string)` to preserve the query string. Alternatively, if we don't expect any existing links/bookmarks with trailing slash and don't really care about the redirect it looks like there's a way to make flask handle links without slash and with slash the same way without any redirects. http://librelist.com/browser/flask/2011/2/27/re-automatic-removal-of-trailing-slashes/#043b1a0b6e841ab8e7d38bd7374cbb58

quis: @allait I think these comments are relevant to the buyer app only at the moment. Separate pull request here: https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/53

PR #41, opened 2015-05-22
Error pages

minglis:
Bringing the error pages into line with those on the buyer app.

- Please note there is an issue currently where API errors cause an internal server error to be displayed NOT the error template.

- This is because whilst rendering a template Flask-Login tries to confirm the user - fails (API error) and throws an error - preventing the template rendering.

- Propose we merge this (as this issue is preexisting) and I'll add a story to fix in another way.

### DIFF
--- a/app/templates/errors/404.html
+++ b/app/templates/errors/404.html
@@ -6,5 +6,12 @@
   <h1>Page could not be found</h1>
 </header>
 
+<p>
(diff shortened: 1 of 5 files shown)


minglis:
https://www.pivotaltracker.com/story/show/95235942

TheDoubleK:
I broke stuff and it looks nice.

PR #43, opened 2015-05-27
Matched footer links with those on buyer app

minglis:
Matched the footer in supplier app to the new buyer app footers.

Note cookie page is hosted on the buyer app.

### DIFF
--- a/app/templates/_base_page.html
+++ b/app/templates/_base_page.html
@@ -34,7 +34,7 @@
     <ul>
         <li><a href="/terms-and-conditions" class="terms-and-conditions">Terms and conditions</a></li>
         <li><a href="/cookies">Cookies</a></li>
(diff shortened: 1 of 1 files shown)


app/templates/_base_page.html
```diff
@@ -30,7 +30,12 @@
 {% endblock %}
 
 {% block footer_support_links %}
-  <a href="/terms-and-conditions" class="terms-and-conditions">Terms and conditions</a>
+  <h2 class="visuallyhidden">Support links</h2>
+    <ul>
+        <li><a href="/terms-and-conditions" class="terms-and-conditions">Terms and conditions</a></li>
+        <li><a href="/cookies">Cookies</a></li>
+        <li>Built by the <a href="https://www.gov.uk/government/organisations/government-digital-service">Government Digital Service</a></li>
```
TheDoubleK: Is there a reason not to include the ` rel="external"` on the link here? Seems odd to have it in the supplier app but not the buyer one?

PR #44, opened 2015-05-28
Update reset password content

robyoung:
This pull request updates the copy and routes for the password reset flow so that they all use the same language; reset password. Below are a bunch of screen shots showing the changes. The routes have also been changed so that they follow the same language. No redirects have been set up because we are assuming that these pages will not have been bookmarked.

One thing that has been highlighted by @ralph-hawkins but not addressed is the form elements being too wide. A new story has been opened for this [#95618752](https://www.pivotaltracker.com/story/show/95618752).

cc/ @tombye 

### Old
supplierlogin-old

### New
supplierlogin-new

### Old
requestreset-old

### New
screen shot 2015-05-29 at 09 57 36

### Old
resetpassword-old

### New
resetpassword-new


superroz:
Looks good to me.

TheDoubleK:
The line "You can request a new password if your reset link has expired" doesn't really make sense to me. 
You can request a new link, not a new password.

superroz:
Hmm. Interesting. I wonder if we really need that last sentence at all. If we do, suggest changing to:

Enter your email address and we'll send you a link so you can reset your password. Password reset links are only valid for 24 hours. If your reset link has expired, you can request a new one.

TheDoubleK:
I think we can do without the last sentence at all, but revised version good if we really need to spell it out.

app/main/login.py
```diff
@@ -102,19 +102,19 @@ def change_password(token):
...
     template_data = main.config['BASE_TEMPLATE_DATA']
-    return render_template("auth/change-password.html",
+    return render_template("auth/reset-password.html",
                            form=ChangePasswordForm(email_address=email_address,
                                                    user_id=user_id,),
                            **template_data), 200
 
 
-@main.route('/change-password', methods=["POST"])
+@main.route('/update-password', methods=["POST"])
```
TheDoubleK: I'm not sure about having a different URL here from the /reset_password/ - nice to have GET and POST at the same URL to make clear that the form from the GET request POSTs to here.

TheDoubleK: OK I see what you've done with the other endpoint.  How about the `/reset-password/<token>` GET could change to `/update-password/<token>`?

app/assets/scss/_text.scss
```diff
@@ -0,0 +1,9 @@
+
+.lede {
```
TheDoubleK: Is this 'lede' a standard thing from the toolkit?  What does it stand for?

robyoung: It's a literature thing meaning a leading paragraph http://en.wikipedia.org/wiki/Lead_paragraph. Was a new word to me too but apparently it's used elsewhere in GDS.

PR #45, opened 2015-05-28
Incorporate Feature Flags.

pcraig3:
##### Note: the code in here has changed quite significantly, so refer to newer comment(s) below.

***

<del>Added Flask-FeatureFlag package and added back Suppliers editing their own service statuses if app runs in Test mode or Dev mode.  Links don't appear in other modes, and the relevant routes throw 404 errors.</del>

Added [Flask-FeatureFlag](https://github.com/trustrachel/Flask-FeatureFlags) package to dmutils, and made a few changes in here to incorporate it.

[>> Story on Pivotal](https://www.pivotaltracker.com/n/projects/997836)

#### Note
<del>Haven't demoed this to Cath or Martyn yet, so don't merge it.  But by all means see if the code looks good.</del>

Showed it to Martyn and we're good to go.

#### Note 2
Feature strings are are defined as dictionaries in `config.py`, but it's also possible to have them as inline parameters or even hook them up to a database, if we wanted.  For example:
```python
# default format
FEATURE_FLAGS = {
    "cool_feature": True,
    "another_cool_feature": False,
}

# inline format
FEATURE_FLAGS_COOL_FEATURE = True
FEATURE_FLAGS_ANOTHER_COOL_FEATURE = False
```

<del>Do we have opinions about this?  (I kind of like them as they are -- seems like they'd scale better.)</del>
Decided on inline flags.  More consistent with how we currently do things.


quis:
Can this be closed now we have feature flags in utils?

quis:
Or if this now incorporates utils, can we remove the [Do not merge]?

pcraig3:
Moved most of the FeatureFlag initialisation code to digitalmarketplace-utils, and now have to refactor each of the apps to make use of it.

* Generalised some of the `create_app` stuff in `dmutils` that we share across all of our apps.
* Moved the `get_version_label()` function into `dmutils`.
* Add initiate `flask_featureflags.FeatureFlag()` in our `__init__` file
* Include a `get_flags()` function in our /_status endpoint tells us about the flags in our config variables and what they're set to in the current app.

app/__init__.py
```diff
@@ -3,31 +3,32 @@
...
+        configs[config_name],
+        data_api_client=data_api_client,
+        feature_flags=feature_flags,
+        login_manager=login_manager,
+    )
 
     from .main import main as main_blueprint
     from .status import status as status_blueprint
 
-    proxy_fix.init_app(application)
```
TheDoubleK: Is the proxy fixing being done some other way now?  I can't see anything that looks like it would replace this?

config.py
```diff
@@ -61,10 +66,15 @@ class Test(Config):
     WTF_CSRF_ENABLED = False
     SERVER_NAME = 'localhost'
 
+    FEATURE_FLAGS_EDIT_SERVICE_PAGE = enabled_since('2015-06-03')
```
TheDoubleK: Should this be False until we are ready to turn on editing services? i.e. once we have monitoring of changes in the admin app sorted out?

app/__init__.py
pcraig3: It's being done in [digitalmarketplace-utils](https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/__init__.py#L25) since this past Friday.

PR #48, opened 2015-06-08
Edit supplier and contact information page

allait:
Pivotal: [#95432162](https://www.pivotaltracker.com/story/show/95432162)

### Add supplier dashboard view and template (`/suppliers`)
Includes supplier and account information and links to service list and supplier info edit pages.

There's no service counts since API doesn't return them with supplier data yet.

screen shot 2015-06-08 at 16 01 30

### Move supplier services list to /suppliers/services
Renames services.dashboard view and template to list_services.
Adds a breadcrumb link to the new supplier dashboard.

### Add supplier info edit page (`/suppliers/edit`)
Allows editing supplier and main contact details. There's no validation apart from form field limits, data is sent to the API and the response "error" message is displayed if the request fails or returns 400.

Things that are missing compared to dm-prototype design: last edit date, since it's not returned by the API yet
screen shot 2015-06-08 at 16 01 42

### Remove SECRET_KEY default value from app config
Missing SECERET_KEY env variable should always prevent the app from working, otherwise errors in production configuration can mean the app is silently using the default value, which is unsafe.

### Enable CSRF protection for all app endpoints
Flask-WTF will automatically store a CSRF token in the user session and require that each POST request contains a matching value in the submitted form data. Forms still need to either explicitly define
the CSRF token field or mark the view as csrf.exempt.

### Add forms for supplier and contact information
Forms allow us to validate basic field requirements before making a request to the API, since the API doesn't return per-field errors at the moment.

Supplier and contact information have separate forms, since they require separate API requests to save the data and might end up with separate views in the future.

### Use forms for editing supplier and contact information
Forms are populated with data from API response or the previous form submits. Only editable fields are populated.

Form classes are not used for HTML rendering. Instead, input fields are created by Jinja macros and only use data and errors values from the Form field objects.

Contact information form fields are prefixed with `contact_` to avoid possible field name clashes and adding a prefix to the hidden contact `id` field.

Submitted form with errors:
screen shot 2015-06-08 at 16 02 42

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -2,6 +2,7 @@ from flask import render_template, request, redirect, url_for, abort
 from flask_login import login_required, current_user
 
 from dmutils.apiclient import APIError
+from dmutils import flask_featureflags
 
 from ...main import main
 from ... import data_api_client
@@ -10,6 +11,8 @@ from ..forms.suppliers import EditSupplierForm, EditContactInformationForm
 
 @main.route('')
 @login_required
+@flask_featureflags.is_active_feature('SUPPLIER_DASHBOARD',
+                                      redirect='.list_services')
 def dashboard():
     template_data = main.config['BASE_TEMPLATE_DATA']
 
@@ -30,6 +33,7 @@ def dashboard():
 
 @main.route('/edit', methods=['GET'])
 @login_required
+@flask_featureflags.is_active_feature('SUPPLIER_DASHBOARD')
 def edit_supplier(supplier_form=None, contact_form=None, error=None):
     template_data = main.config['BASE_TEMPLATE_DATA']
 
@@ -61,6 +65,7 @@ def edit_supplier(supplier_form=None, contact_form=None, error=None):
 
 @main.route('/edit', methods=['POST'])
 @login_required
+@flask_featureflags.is_active_feature('SUPPLIER_DASHBOARD')
 def update_supplier():
     # FieldList expects post parameter keys to have number suffixes
     # (eg client-0, client-1 ...), which is incompatible with how
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -25,3 +25,54 @@ def dashboard():
         supplier=supplier,
         **template_data
     ), 200
+
+
+@main.route('/edit', methods=['GET'])
+@login_required
+def edit_supplier(error=None):
+    template_data = main.config['BASE_TEMPLATE_DATA']
+
+    try:
+        supplier = data_api_client.get_supplier(
+            current_user.supplier_id
+        )['suppliers']
+        supplier['contact'] = supplier['contactInformation'][0]
+    except APIError as e:
+        abort(e.status_code)
+
+    return render_template(
+        "suppliers/edit_supplier.html",
+        supplier=supplier,
+        error=error,
+        **template_data
+    ), 200
+
+
+@main.route('/edit', methods=['POST'])
+@login_required
+def update_supplier():
+    try:
+        data_api_client.update_supplier(
+            current_user.supplier_id,
+            {
+                "description": request.form['description'],
+                "clients": request.form.getlist('clients'),
+            },
+            current_user.email_address,
+            "Update supplier info"
+        )
+
+        data_api_client.update_contact_information(
+            current_user.supplier_id,
+            request.form['contactId'],
+            {k: request.form[k] for k in [
+                "address1", "address2", "city", "country", "postcode",
+                "website", "phoneNumber", "email", "contactName",
+            ]},
+            current_user.email_address,
+            "Update supplier info"
+        )
+    except APIError as e:
+        return edit_supplier(error=e.message)
+
+    return redirect(url_for(".dashboard"))
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -5,6 +5,7 @@ from dmutils.apiclient import APIError
 
 from ...main import main
 from ... import data_api_client
+from ..forms.suppliers import EditSupplierForm, EditContactInformationForm
 
 
 @main.route('')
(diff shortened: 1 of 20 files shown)


TheDoubleK:
I don't think the link to view Services is at all obvious - "Services" at the top looks like a heading with nothing below it.  It doesn't make sense that you have to click a small "Edit" link to be able to view the list of services.

allait:
@TheDoubleK the design prototype has service counts below the Services header (http://dm-prototype.herokuapp.com/dashboard). We're not doing this now since API doesn't return the supplier service counts yet. I've talked with Cath and we've agreed to release this as is for now, since it's going to be behind a feature flag.

TheDoubleK:
The supplier summary box allows you to save summaries OK when over the word limit - needs some validation on submission.

TheDoubleK:
POSTing too many clients in the clients list results in a 500 (although this isn't the end of the world, it might be nice to catch the error and return a validation message rather than a generic 500).

TheDoubleK:
Phone number allows letters to be saved - not sure if we should do some validation for only digits, +, - and maybe space?

TheDoubleK:
Missing address lines (which are allowed) are rendered as "None" in the edit form, and then saved as the string "None" subsequently.  Null fields should be left empty when pre-filling the form.

TheDoubleK:
Although in rendering addresses on the summary page it would be nice if empty address fields didn't give you an empty line, e.g: 
screen shot 2015-06-09 at 11 16 32

app/main/views/suppliers.py
```diff
@@ -0,0 +1,100 @@
...
+        supplier_form = EditSupplierForm(
+            description=supplier['description'],
+            clients=supplier['clients']
+        )
+        contact_form = EditContactInformationForm(
+            prefix='contact_',
+            **supplier['contactInformation'][0]
+        )
+
+    return render_template(
```
TheDoubleK: Nice - passing in two forms to one template - neat!

TheDoubleK:
Clicking the "Digital Marketplace" breadcrumb here:
screen shot 2015-06-09 at 11 43 02

Results in this:
screen shot 2015-06-09 at 11 41 57

Not sure if this is a separate bug to be added in Pivotal, but the breadcrumbs in preview work OK, so it may be introduced by this pull request.

app/templates/macros/forms.html
```diff
@@ -0,0 +1,79 @@
+{% macro question_input(name, label, value='', hint='', errors=None) -%}
```
TheDoubleK: Are these macros the kind of thing that should live in the Frontend Toolkit? This pull-request from @quis has at least some of these elements in it: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/70

tests/app/main/test_suppliers.py
```diff
@@ -0,0 +1,216 @@
...
+            self.login()
+
+            res = self.client.get("/suppliers")
+            assert_equal(res.status_code, 200)
+
+            data_api_client.get_supplier.assert_called_once_with(1234)
+
+            resp_data = res.get_data(as_text=True)
+
+            assert_in("Supplier Description", resp_data)
```
TheDoubleK: Might it be worth testing that each field is in the right place on the page, rather than just somewhere on the page?  Could be a fiddly though.

tests/app/main/test_suppliers.py
```diff
@@ -0,0 +1,216 @@
...
+    @mock.patch("app.main.views.suppliers.data_api_client")
+    def test_shows_edit_buttons(self, data_api_client):
+        data_api_client.get_supplier.side_effect = get_supplier
+        with self.app.test_client():
+            self.login()
+
+            res = self.client.get("/suppliers")
+            assert_equal(res.status_code, 200)
+
+            assert_true("/suppliers/edit" in res.get_data(as_text=True))
```
TheDoubleK: This could test that these are the href values for a particular link - just having these strings in the response doesn't really guarantee that the edit buttons exist.

allait:
Thanks for all the amazing comments @TheDoubleK!
I've fixed the supplier summary and clients limit validations.
I've looked at the existing phone numbers in the suppliers export and some of the values are:
```
    Office <number hidden>    Mobile <number hidden>
    <number hidden> | <number hidden>
    <number hidden> or <number hidden>
    <number hidden>; <nuber hidden>
```
so it looks like there was no validation before and we'd need to sanitise the data before adding any (otherwise suppliers won't be able to save existing data when editing). I think if we still want to do this it should be a separate story.

The breadcrumb is linking to the root page, which belongs to the buyers app, so it's a local issue. There is an infinite redirect when running the app locally, which only affects the '/' root (normally unaccessible through nginx). I was considering fixing it by adding a check to the trailing slash removal middleware, but it's an additional bit of logic running on every request for something that only affects development server, so I left it as is for now.

Re. tests: I don't want to go to deep into the template testing in the unit tests, it feels like something that's better served by functional tests. But I will think about ways to improve the checks, since they don't give me much confidence either. Not sure if this will be a part of this PR though.

Still **TODO** I'll integrate the new frontend toolkit forms instead of the macros and replace the address layout with the toolkit one (this should take care of the extra empty lines). I'll also add a feature flag wrapper for the whole dashboard since that's now merged.

allait:
I've added a feature flag to hide supplier dashboard in live environments.

I'm holding off on replacing form and address layout with toolkit includes since I've run into some issues with them that I'd like to talk with @quis about first.

I suggest we merge this without the toolkit for now and update all the templates that aren't using toolkit later.

TheDoubleK:
We need a CSRF token added to the edit service page template, i.e. the pages like `/suppliers/services/6040697471762432` - trying to update the status now gives a CSRF error.

TheDoubleK:
Agreed re: phone numbers - if there's letters in them already let's leave it.

TheDoubleK:
Agreed that refactoring all templates using the toolkit can be done as a later job - this looks good for now.

Happy to merge this once that extra CSRF token is in. :+1:

allait:
@TheDoubleK added CSRF token to the service page, nice catch!

PR #53, opened 2015-06-10
Remove DM_HOST variable

tombye:
It's not needed as explained in alphagov/digitalmarketplace-buyer-frontend#74 but it's useful to be able to specify the host when running the `scripts/run_app.sh` script so this also adds that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,12 @@ not already been set:
 ./scripts/run_app.sh
 ```
 
+To set the host the app will run on, use the `h` flag:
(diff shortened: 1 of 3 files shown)


allait:
Can we just pass in all the `$@` options to `run_app.sh` directly to `python application.py`? I don't think we need to do our own parsing since we don't have any additional options.

tombye:
Didn't know about `$@` so thanks for the tip. Now amended.

PR #56, opened 2015-06-10
Login session should expire

TheDoubleK:
Addresses this bug: https://www.pivotaltracker.com/story/show/96536586

It turns out that it is a feature of the Chrome browser that it does not delete temporary cookies and sessions in the way browsers are supposed to, so the Flask default of a temporary session that dies when the browser is closed will leave Chrome users permanently logged-in.

To fix this we can instead use a permanent session with a timeout.  This means that if a user closes and re-opens their browser within the timeout period they will still be logged in.  I have set the session to expire after 15 minutes of inactivity on the site - the clock is reset with every new request.  This should hopefully give a reasonable level of security without overly annoying users with repeated demands to log in.

I have also made it so that logged-out users who try to access a resource that they must be logged-in to access will be taken directly to that resource after logging in, rather than being always taken to the dashboard page after log in.


TheDoubleK:
Rebased against master for auto-merge again.

TheDoubleK:
Rebased against master for auto-merge again. This time fixing PEP8 errors created during rebasing.

PR #58, opened 2015-06-11
Sync secure cookie config with admin app

pcraig3:
Instead of having our secure cookies set to `False` and then explicitly set as `True` in a Live environment, it's better to assume secure cookies and explicitly unset them where needed.  
This is how it's been done [in the admin app](https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/66/files), at any rate.
Also added a test to confirm secure cookies are enabled.

### DIFF
--- a/tests/app/main/test_login.py
+++ b/tests/app/main/test_login.py
@@ -1,5 +1,5 @@
 from app.main import helpers
-from nose.tools import assert_equal, assert_true, assert_is_not_none
+from nose.tools import assert_equal, assert_true, assert_is_not_none, assert_in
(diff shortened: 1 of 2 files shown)


tests/app/main/test_login.py
```diff
@@ -33,6 +33,7 @@ def test_should_redirect_to_dashboard_on_login(self):
         })
         assert_equal(res.status_code, 302)
         assert_equal(res.location, 'http://localhost/suppliers')
+        assert_in('Secure;', res.headers['Set-Cookie'])
```
TheDoubleK: :+1: Nice one.

PR #63, opened 2015-06-12
Deny third-party sites from embedding in an iframe

pcraig3:
Upgrading to new verison of dmutils which initiates each app with the `X-Frame-Options` header set to `DENY`.
Also added test to confirm.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ Flask==0.10.1
 Flask-Login==0.2.11
 Flask-Script==2.0.5
 Flask-WTF==0.11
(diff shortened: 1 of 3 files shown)


quis:
Same comment as on the buyer's app, testing for the header would be better (I think) as its own separate test.

quis:
Also, good branch name.

minglis:
As per Chris comment, would the test fit in here?

https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/tests/app/test_application.py

pcraig3:
@minglis Sure, it's an app-wide setting, so seems logical.

@quis Branch name's a bit of a stretch, but I got there in the end.

PR #64, opened 2015-06-12
Render service from YAML

quis:
_In partial fulfilment of [#95430642](https://www.pivotaltracker.com/story/show/95430642)_

This pull request present the following fields on the service page:
- Service name
- Service description
- Features
- Benefits

It doesn't do any editing of these fields.

It tries, as much as possible to share code with the admin app.

image


app/templates/macros/answers.html
```diff
@@ -0,0 +1,66 @@
+{% macro textbox(value='') -%}
```
TheDoubleK: Will these macros eventually be shared between apps - in either the utils package or frontend toolkit?

app/templates/services/_edit-status.html
```diff
@@ -0,0 +1,34 @@
+{% if service_data.status != 'disabled' %}
+  <form action="{{ url_for('.update_service_status', service_id=service_data['id'] ) }}" method="POST">
```
TheDoubleK: The `url_for` seems to be broken here - I think it's encoding the "presented" version of the Service ID - e.g. :+1: `<form action="/suppliers/services/%5B%275837%27%2C%20%276279%27%2C%20%279545%27%2C%20%275488%27%5D" method="POST">`
Instead of `/suppliers/services/5837627995455488`

TheDoubleK:
Looks good to go once the update status form URL is fixed.

app/templates/macros/answers.html
quis: Undecided as yet. You _could_ argue that it's each app's responsibility to decide what the interface is—think of the buyer app which displays the description of a service differently to the admin/supplier apps.

quis:
@TheDoubleK Form URL is fixed in 2ce125e.

TheDoubleK:
:+1:

PR #65, opened 2015-06-15
Fail the build if SASS doesn't compile

quis:
We have had problems with builds going through and getting deployed without any stylesheets.

This commit addresses that issue by causing `./scripts/run_tests.sh` to fail if the SASS compilation doesn't complete.

This means that problems will be picked up locally or on Travis before they reach deployment.

--

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -68,6 +68,14 @@ var uglifyOptions = {
   }
 };
 
+var logErrorAndExit = function logErrorAndExit(err) {
+
+  // coloured text: https://coderwall.com/p/yphywg/printing-colorful-text-in-terminal-when-run-node-js-script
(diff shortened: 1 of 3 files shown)


gulpfile.js
```diff
@@ -68,6 +68,14 @@ var uglifyOptions = {
   }
 };
 
+var logErrorAndExit = function logErrorAndExit(err) {
+
+  // coloured text: https://coderwall.com/p/yphywg/printing-colorful-text-in-terminal-when-run-node-js-script
+  console.log('\x1b[41m\x1b[37m  Error: ' + err.message + '\x1b[0m');
```
TheDoubleK: Not quite as catchy as `bgcolor=red`, is it?

PR #68, opened 2015-06-16
Disable autocomplete on logins.

pcraig3:
Login forms and (non-hidden) input elements belonging to them are checked for `autocomplete=off` attribute, which, somewhat intuitively, prevents browsers from autofilling their content.

Demoed to Martyn and I'll have to do the same thing for the Admin App, so let me know if you have any suggestions for improvements before I do the other one.

[>> Comment ("5.21 disable autocomplete on logins") in story on Pivotal](https://www.pivotaltracker.com/story/show/96550136)

### DIFF
--- a/app/templates/auth/login.html
+++ b/app/templates/auth/login.html
@@ -47,7 +47,7 @@
     <h1>Supplier login</h1>
 </header>
 
-<form action="{{ url_for('.process_login', next=next) }}" method="POST">
+<form autocomplete="off" action="{{ url_for('.process_login', next=next) }}" method="POST">
 
     {{ form.hidden_tag() }}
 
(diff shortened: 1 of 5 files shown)


app/templates/auth/login.html
```diff
@@ -47,7 +47,7 @@ <h3 class="validation-masthead-heading" id="validation-masthead-heading">
     <h1>Supplier login</h1>
 </header>
 
-<form action="{{ url_for('.process_login', next=next) }}" method="POST">
+<form autocomplete="off" action="{{ url_for('.process_login', next=next) }}" method="POST">
```
TheDoubleK: Does it need to be on both the form and the individual fields?  Seems like either one should do on its own?

pcraig3: Not totally sure, [but it seemed like the safe way to go](http://stackoverflow.com/questions/2530/how-do-you-disable-browser-autocomplete-on-web-form-field-input-tag#comment4833209_2531).

PR #69, opened 2015-06-16
[WIP] Prevent unwanted reuse of password reset tokens.

pcraig3:
Password reset tokens, once issued, would be valid for 24 hours.
This meant that all generated tokens could be used for the full 24 period without discrimination.
Added some logic to disallow tokens that were created _before_ a user's password was last changed.
Also created a new method that isolates token verification logic.

Also also replaced calls to `Mock` with `mock.patched` objects for our login tests because apparently that's a better way to go.

[>> Task ("Password reset tokens can be used multiple times") in story on Pivotal](https://www.pivotaltracker.com/story/show/96550136)

### DIFF
--- a/app/main/helpers/email.py
+++ b/app/main/helpers/email.py
@@ -1,6 +1,7 @@
 from flask import url_for, current_app, render_template
 import mandrill
 from itsdangerous import URLSafeTimedSerializer
+from datetime import datetime
 
 from .. import main
 
@@ -49,5 +50,12 @@ def decode_email(token):
     ts = URLSafeTimedSerializer(main.config["SECRET_KEY"])
     decoded = ts.loads(token,
                        salt=main.config["RESET_PASSWORD_SALT"],
-                       max_age=86400)
+                       max_age=86400, return_timestamp=True)
     return decoded
+
+
+def token_created_before_password_last_changed(token_timestamp, user):
+
+        password_last_changed = datetime.strptime(
+            user['users']['passwordChangedAt'], '%a, %d %b %Y %H:%M:%S %Z')
+        return token_timestamp < password_last_changed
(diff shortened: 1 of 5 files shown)


app/main/helpers/email.py
```diff
@@ -49,5 +50,12 @@ def decode_email(token):
...
                        salt=main.config["RESET_PASSWORD_SALT"],
-                       max_age=86400)
+                       max_age=86400, return_timestamp=True)
     return decoded
+
+
+def token_created_before_password_last_changed(token_timestamp, user):
+
+        password_last_changed = datetime.strptime(
+            user['users']['passwordChangedAt'], '%a, %d %b %Y %H:%M:%S %Z')
```
TheDoubleK: I wonder how brittle this method of decoding the time is?  The API user serialize() method doesn't specify any particular format for the time string - perhaps it should do?  That way we can be sure the timestamp will be in the format expected by the supplier app here.

TheDoubleK: The Created At and Updated at fields in the API specify the following format: `DATE_FORMAT = "%Y-%m-%dT%H:%M:%S%Z"` - I think we should probably use the same format for all our datetime strings.

TheDoubleK:
I think it's worth standardising our datetime format in the API now, as I don't think anything else depends on the existing format at the moment, so shouldn't break anything to add it in.  Once that's done this looks good to go.

app/main/helpers/email.py
pcraig3: So should we enforce the correct format in this app (with a fallback if we actually get the format we expect) so that we can merge it and then correct the output from the API?

Also, I've used that date format string in two places -- there's one in the tests as well.  Would it be better to store it in a config variable?

TheDoubleK: As discussed - format should be in the Utils package and shared between apps.

pcraig3:
Added [WIP] tag while updating supplier app to work with new digitalmarketplace-utils.

allait:
#71 was merged, is there anything else blocking this?

pcraig3:
Closing pull request while #71 is closed.

PR #70, opened 2015-06-16
Pentest mitigations headers

minglis:
Adds no-cache header to supplier app


robyoung:
The comment on the first commit states that it adds a dependency on a more recent version of utils but it doesn't.

robyoung:
The first commit also sneakily brings in feature flags. Was that intended?

TheDoubleK:
The "files changed" overall only has the cache control header stuff though.

TheDoubleK:
The rebase conflicts in this are horrible so going to merge it with the commit history as-is.

robyoung:
Fair enough, it's not the end of the world.

PR #71, opened 2015-06-16
Make use of new dmutils features

quis:
Namely:

- Add present_all method to presenters
  https://github.com/alphagov/digitalmarketplace-utils/pull/61

- Make content loader handle dependencies…
  https://github.com/alphagov/digitalmarketplace-utils/pull/62

…which means that the app can update to the latest content structure, as changed in: https://github.com/alphagov/digital-marketplace-ssp-content/pull/42

This commit also makes use, wherever possible, of the templates from the toolkit, rather than local templates.

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -43,17 +43,11 @@ def edit_service(service_id):
         abort(404)
 
     template_data = main.config['BASE_TEMPLATE_DATA']
-
-    presented_service_data = {}
-    for key, value in service.items():
-        presented_service_data[key] = presenters.present(
-            value, content.get_question(key)
-        )
-
+    content.filter(service)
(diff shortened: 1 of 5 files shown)


app/main/views/services.py
```diff
@@ -43,17 +43,11 @@ def edit_service(service_id):
...
 
     template_data = main.config['BASE_TEMPLATE_DATA']
-
-    presented_service_data = {}
-    for key, value in service.items():
-        presented_service_data[key] = presenters.present(
-            value, content.get_question(key)
-        )
-
+    content.filter(service)
```
robyoung: Same thread safety issue here.

quis:
Closing this until issues are resolved.

PR #72, opened 2015-06-18
Edit existing service

quis:
_Fulfills [95430642](https://www.pivotaltracker.com/story/show/95430642)_

image

This pull request, [using the new dmutils features](https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/71) and templates from the toolkit:
- adds a `GET` endpoint for `/services/<id>/edit/<section>`
- renders form fields with prefilled values for the above four questions, using templates from the toolkit
- adds a `POST` endpoint for the same, which takes the edited values and
  - if the API returns successfully redirects back to the service summary page
  - if the API returns an error then it will be displayed back to the user

It doesn't, yet, do any per-field validation of edits.

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -112,11 +112,16 @@ def update_service_status(service_id):
 @login_required
 @flask_featureflags.is_active_feature('EDIT_SERVICE_PAGE')
 def edit_section(service_id, section):
-    service_data = data_api_client.get_service(service_id)['services']
+
+    service = data_api_client.get_service(service_id)['services']
+
+    if not _is_service_associated_with_supplier(service):
+        abort(404)
+
     return render_template(
         "services/edit_section.html",
-        section=content.get_section_filtered_by(section, service_data),
-        service_data=service_data,
+        section=content.get_section_filtered_by(section, service),
+        service_data=service,
         service_id=service_id,
         **main.config['BASE_TEMPLATE_DATA']
     )
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -149,13 +149,9 @@ def update_section(service_id, section):
     for key in request.form:
         item_as_list = request.form.getlist(key)
         list_types = ['list', 'checkboxes', 'pricing']
-        print("============================")
-        print(key)
-        print(content.get_question(key))
-        print(content.get_question(key).get("type"))
-        print("============================")
         if (
             key != 'csrf_token' and
+            # 'type' in content.get_question(key) and
             content.get_question(key)['type'] in list_types
         ):
             posted_data[key] = item_as_list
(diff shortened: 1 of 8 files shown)


TheDoubleK:
I'm getting a "None" validation error next to service description, eg:
screen shot 2015-06-18 at 15 42 57

TheDoubleK:
Hang on, let me run the frontend build. and try again...

TheDoubleK:
OK the "None" has gone, but the edit page seems to be missing some features & benefits, e.g. this service (with 5 features and 3 benefits):
screen shot 2015-06-18 at 15 47 47

When you click to edit only shows four and two (With the first one missing from each list):
screen shot 2015-06-18 at 15 48 02

Also, it allows saving of empty strings as a feature or benefit (although this will be fixed by future validation rather than for now, maybe):
screen shot 2015-06-18 at 15 51 17

app/main/views/services.py
```diff
@@ -113,6 +105,72 @@ def update_service_status(service_id):
...
 
 
+@main.route(
+    '/services/<string:service_id>/edit/<string:section>',
+    methods=['GET']
+)
+@login_required
+@flask_featureflags.is_active_feature('EDIT_SERVICE_PAGE')
+def edit_section(service_id, section):
+    service_data = data_api_client.get_service(service_id)['services']
```
TheDoubleK: This endpoint needs a check that the service belongs to the currently logged-in supplier.  Otherwise if I'm logged in I can try to edit anyone else's services (the POST will fail, but still it looks bad...)

quis: @TheDoubleK Check added in 55d954d

app/main/views/services.py
```diff
@@ -113,6 +105,77 @@ def update_service_status(service_id):
...
+    service = data_api_client.get_service(service_id).get('services')
+
+    if not _is_service_associated_with_supplier(service):
+        abort(404)
+
+    posted_data = dict(
+        list(request.form.items()) + list(request.files.items())
+    )
+
+    posted_data.pop('csrf_token', None)
```
TheDoubleK: I had a feeling that @allait had it set up so that the csrf token got checked and popped for all POST requests before being processed by the view.  But I can't seem to find that code now - did I dream it, or has it been removed?

app/main/views/services.py
```diff
@@ -113,6 +105,77 @@ def update_service_status(service_id):
...
+
+    # Turn responses which have multiple parts into lists
+    for key in request.form:
+        item_as_list = request.form.getlist(key)
+        list_types = ['list', 'checkboxes', 'pricing']
+        if (
+            key != 'csrf_token' and
+            # 'type' in content.get_question(key) and
```
TheDoubleK: Looks like this comment can be removed.

app/templates/services/edit_section.html
```diff
@@ -0,0 +1,66 @@
...
+
+    </div>
+  </div>
+  <form method="post" action="{{ url_for(".update_section", service_id=service_id, section=section.id) }}">
+
+    {% for question in section.questions %}
+      {{ forms[question.type](question, service_data, {}) }}
+    {% endfor %}
+
+    <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
```
TheDoubleK: Would it be worth having this in some sort of "posted form" base template? Or is this edit_section template-y enough?

TheDoubleK:
Reviewed - should be good to go once the features and benefits count is fixed.

quis:
@TheDoubleK Off by one error will require merging https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/85

app/main/views/services.py
quis: Removed in 6ad51ad

quis:
@TheDoubleK Off by one error fixed in  4c021ed

PR #74, opened 2015-06-18
Add framework service counts to the supplier dashboard

allait:
[#97275278](https://www.pivotaltracker.com/story/show/97275278)

Requires alphagov/digitalmarketplace-api#150 

Displays the service counts as reported by the API.
There are no framework dates since we don't return them from the API yet

screen shot 2015-06-18 at 16 11 22

G-Cloud 7 "now open" block will be a part of @robyoung's pull request.


quis:
Spoke with @ralph-hawkins, agree that 'view' instead of 'edit' would be better for the link text.

quis:
:+1:

PR #79, opened 2015-06-25
[# 97631008] Add draft documents S3 and prune config

robyoung:
This adds an environment variable for the S3 bucket we want to use for uploading draft documents. It does not actually do any uploading yet because that will use the existing S3 upload from dmutils.

When adding the config I noticed a few old configs that could be cleaned up as well.


TheDoubleK:
Are there corresponding changes to be made in the Cloudformation templates to add the new environment vars?

PR #80, opened 2015-06-29
Put all service attributes in the manifest for new services

quis:
This pull request:
- matches [the order and groupings used in the Grails app](https://github.gds/gds/digitalmarketplace-frontend/blob/c7a901c6cd9b7c53db21b4dbe3b6717c62e1e7b4/grails-app/views/marketplace/listing/_G6Listing.gsp) for displaying services
- does just enough to stop the above throwing template exceptions when viewing a service

--

It does not:
- display all the service data perfectly (it especially has trouble with questions that have an assurance component)
- do any integration with the draft services stuff

--

gov uk - the best place to find government services and information copy
longcat

### DIFF
--- a/app/new_service_manifest.yml
+++ b/app/new_service_manifest.yml
@@ -45,7 +45,7 @@
     - termsAndConditionsDocumentURL
 -
   name: Support
-  editable: true
+  editable: True
   questions:
     - supportTypes
     - supportForThirdParties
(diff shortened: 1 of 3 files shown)


TheDoubleK:
Weird how some booleans display OK but others don't? Or is that just the ones with assurance? I think I just answered my own question... as you were.

PR #81, opened 2015-06-29
[#97502388] Service updates: Respond to validation errors returned by the API

TheDoubleK:
**NOTE**: This requires validation error responses from the API, which are added in this pull request: https://github.com/alphagov/digitalmarketplace-api/pull/159

This prevents updates with validation errors from being saved, and displays validation errors on the form.

In-line messages are no properly mapped to all possible validation errors - it could be easily done for this small set of fields, but it will be better to harmonise with the messaging mechanism used for SSP flow once it exists.

For now the validation message shown is the last in the list of `validations` for the question in SSP content.

There seems to be some issue with the width of the "Edit service" form - @quis can you help with that at all?

screen shot 2015-06-29 at 15 03 15

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -159,15 +159,18 @@ def update_section(service_id, section):
         except HTTPError as e:
             errors_map = {}
             for error in e.response.json()['error'].keys():
-                id = existing_service_content.get_question(error)['id']
-                errors_map[id] = \
-                    {
-                    'input_name': error,
-                    'question': existing_service_content
-                        .get_question(error)['question'],
-                    'message': existing_service_content
-                        .get_question(error)['validations'][-1]['message']
-                    }
+                if error == '_form':
+                    abort(400, "Submitted data was not in a valid format")
+                else:
+                    id = existing_service_content.get_question(error)['id']
+                    errors_map[id] = \
+                        {
+                        'input_name': error,
+                        'question': existing_service_content
+                            .get_question(error)['question'],
+                        'message': existing_service_content
+                            .get_question(error)['validations'][-1]['message']
+                        }
 
             return render_template(
                 "services/edit_section.html",
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -157,6 +157,17 @@ def update_section(service_id, section):
                 "user",
                 "supplier app")
         except HTTPError as e:
+            errors_map = {}
+            for error in e.response.json()['error'].keys():
+                id = existing_service_content.get_question(error)['id']
+                errors_map[id] = {
+                    'input_name': error,
+                    'question': existing_service_content
+                    .get_question(error)['question'],
+                    'message': existing_service_content
+                    .get_question(error)['validations'][-1]['message']
+                }
+
             return render_template(
                 "services/edit_section.html",
                 section=content.get_section(section),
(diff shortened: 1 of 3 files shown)


quis:
@TheDoubleK Fixed The problem with the banner at the top not being full-width in 2816168


Not sure what's going on with the textbox—looks OK for me locally:
image

I wonder if it's because we're not stripping line breaks when edits are being made…?

TheDoubleK:
The summary is fine I think - yes it's multi-line, which we do keep at the moment.

app/main/views/services.py
```diff
@@ -157,6 +157,17 @@ def update_section(service_id, section):
                 "user",
                 "supplier app")
         except HTTPError as e:
+            errors_map = {}
+            for error in e.response.json()['error'].keys():
+                id = existing_service_content.get_question(error)['id']
+                errors_map[id] = {
+                    'input_name': error,
+                    'question': existing_service_content
+                    .get_question(error)['question'],
```
allait: It's a nitpick, but could we indent the `.get_question` lines so they stand out among the keys.

Side note: I think it's another example of how our line length code style check often doesn't make things better. Maybe it's worth discussing the idea of removing it.

app/templates/macros/toolkit_forms.html
```diff
@@ -17,7 +17,8 @@
     name=question_content.id,
     value=service_data[question_content.id],
     large=True,
-    max_length_in_words=question_content["maxLengthInWords"] or False
+    max_length_in_words=question_content["maxLengthInWords"] or False,
+    error=errors.get(question_content.id)['message'] or None
```
allait: Do we need the `or None` and `or False` here? Looking at the toolkit code (https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/toolkit/templates/forms/list-entry.html#L17) it just checks for `{% if error %}`, so there doesn't seem to be a need to replace any false value with `None`.

TheDoubleK: Good spot - they seem to be superfluous.

app/main/views/services.py
TheDoubleK: Fixed, but you're right about the line length.  The limit could be a bit more than 80 at least (maybe 120 chars)?

app/main/views/services.py
```diff
@@ -157,6 +157,18 @@ def update_section(service_id, section):
                 "user",
                 "supplier app")
         except HTTPError as e:
+            errors_map = {}
+            for error in e.response.json()['error'].keys():
+                id = existing_service_content.get_question(error)['id']
```
allait: I get a `KeyError` on this line when `'errors'` contains non-field keys, e.g. `_form`

TheDoubleK: How do you provoke a `_form` error here? Do you have to hack the posted data or is there input you can make on a form that will break it?

allait: I get `u"Additional properties are not allowed (u'lastUpdatedByEmail', u'lastCompletedByEmail', u'lastCompleted', u'lastUpdated' were unexpected)"` which is probably related to the recent changes to the api.

But I still think we should handle the `_form` or any other unexpected key here.

TheDoubleK:
I've put a catch in for `_form` errors that I think is enough. The way the errors are generated at the API the only other possible keys exist in the schema, so there's no point checking for additional keys.

Also worth bearing in mind that this method is likely to change once we sort out the validation messaging for the SSP in the content a bit more.

PR #82, opened 2015-06-29
[#997836] Kick off a new draft service

pcraig3:
<del>Forgot about the breadcrumbs on the template page.  Probably need to talk to @quis or @ralph-hawkins before merging this</del>
Talked with Chris about the breadcrumbs and we agreed they should look like this.
screen shot 2015-07-02 at 12 01 07

***

Added a GET route to a page where logged-in users can choose a lot for their ~agile~ new service, and a POST route that creates a new draft for them and redirects to the first SSP question based on their selected lot.  If you try to submit the form _without_ a lot selected, you're returned to the same form with an appropriate error message.

Lot selection content pulled in from [`lot.yml`](https://github.com/alphagov/digitalmarketplace-ssp-content/blob/master/g6/lot.yml) in the SSP content, very cool. 
Form is generated using [`Selection buttons`](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/forms/selection-buttons.html) in the toolkit, which is also very cool.

There's a bit of a funny thing going on where the SPP content stuff says [`radios`](https://github.com/alphagov/digitalmarketplace-ssp-content/blob/master/g6/lot.yml#L2) instead of `radio`, which is probably something that should eventually be fixed.

Mocking stuff in tests is a bit a nightmare right now, but they all work, so.

[>> Story on Pivotal](https://www.pivotaltracker.com/n/projects/997836)


quis:
@pcraig3 What's your question about the breadcrumbs?

app/main/views/services.py
```diff
@@ -246,6 +245,77 @@ def update_section_submission(service_id, section):
...
+@login_required
+@flask_featureflags.is_active_feature('CREATE_SERVICE_PAGE')
+def start_new_draft_service():
+    """
+    Page to kick off creation of a new (G7) service.
+    """
+    template_data = main.config['BASE_TEMPLATE_DATA']
+
+    lots = new_service_content.get_question('lot')
+    lots['type'] = 'radio'
```
TheDoubleK: I'm pretty sure there was a rationale behind `radios` instead of `radio` in the SSP content (@quis can probably shed some light).  It might be better to stick with `radios` for this macro too for consistency?

TheDoubleK:
The text describing the lots seems to be bunched over on the left - is that intentional?
screen shot 2015-07-02 at 09 29 30

app/main/views/services.py
```diff
@@ -246,6 +245,77 @@ def update_section_submission(service_id, section):
...
+    Hits up the data API to create a new draft (G7) service.
+    """
+    lot = request.form.get('lot', None)
+
+    if not lot:
+        return redirect(
+            url_for(".start_new_draft_service", error="Answer is required")
+        )
+
+    # hard-coding in a framework slug :/
```
TheDoubleK: Rather than hard coding we could make the slug a path parameter on the GET and POST endpoints - i.e. /submission/<string:framework_slug>/create.

pcraig3: @robyoung says it's not worth doing this, and I think he's right.

If we allow passed-in parameters then we have to validate them, and currently there's only one possibility that we would allow anyway and we already know what that possibility looks like.

app/main/views/services.py
pcraig3: All of the parameters in the SSP are passed directly into the [form-building elements](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/forms/selection-buttons.html), which *almost* works.

The SSP stuff says `type: radios`, which, when passed straight through to the form-building code, renders a broken `<input type="radios">` element.  

I'm not really sure what else depends on using the word `radios`, but it might be nothing.

PR #84, opened 2015-07-01
Role aware authentication and locked user accounts

robyoung:
## Only 'supplier' accounts should be able to sign into the supplier frontend
https://www.pivotaltracker.com/story/show/95620350

Now that the admin frontend authentication is handled by the data API we will start having user accounts that are not supplier accounts in there. As a result the frontend apps need to be aware of the role so that they can only let the right kind of user in.

## User accounts can now be locked
https://www.pivotaltracker.com/story/show/97201260
- Update the login failure message to make it clear that a user account can be locked and what the user must do to rectify it.
- Update the password reset email to tell the user if their account is locked and what they must do to rectify it.

This includes copy changes and changes to the authentication flow. ~~It should get signed off by Cath.~~ It has been signed off by Cath.


app/templates/auth/login.html
```diff
@@ -12,8 +12,10 @@
     {% endif %}
         <p class="banner-message">
             {% if message == 'no_account' %}
-            Sorry, we couldn't find a supplier account with that username
-            and password.
+            Sorry, we couldn't log you in with that username and password.
+            Accounts are locked after five failed attempts. If you think you
+            may have locked your account, please contact
```
allait: "you may have locked your account" sounds like we're blaming the user, I'm sure there must be a more neutral way to say "email us if you can't log in"
And I think `support@` is confusingly the email all of our monitoring is going to, the contact email in the footer is "enquiries@digitalmarketplace.service.gov.uk"

TheDoubleK: How about "If you think your account has been locked please contact..." or even more wishy-washy "If you think your account may be locked please contact..."

allait:
I've updated the message with "If you think your account has been locked please contact..." and changed the contact email address to "enquiries@"

app/templates/auth/login.html
```diff
@@ -12,8 +12,10 @@
     {% endif %}
         <p class="banner-message">
             {% if message == 'no_account' %}
-            Sorry, we couldn't find a supplier account with that username
-            and password.
+            Sorry, we couldn't log you in with that username and password.
```
TheDoubleK: Is it worth breaking up this text a bit do you think?  Maybe a line break or two after the first sentence?

screen shot 2015-07-06 at 09 34 28

TheDoubleK:
The `/suppliers/services` page is not properly locked out for users without the supplier role.  It has 'None' displayed for supplier name and then lists all services (I think from any supplier):

screen shot 2015-07-06 at 09 40 25

TheDoubleK:
I'm not sure that a 404 the correct response for a user who is logged in but has the wrong role for the page they are trying to access.  It might be better for non-supplier users to be redircted to the login page with a suitable message (e.g. 'you need to be logged in with a supplier account to access this page').  It seems weird to redirect to login for not-logged-in-at-all users but to 404 for logged-in buyers/admins.

app/templates/emails/reset_password_email.html
```diff
@@ -9,10 +9,16 @@
 <br /><br />
 You recently asked to reset your Digital Marketplace password. 
 <br /><br />
+{% if locked %}
```
TheDoubleK: Should we have some testing around the email content?

PR #95, opened 2015-07-08
[#97631082] Use specific serviceTypes for SaaS, IaaS, SCS

TheDoubleK:
Now we have more than one possible lot we need to use the specific `serviceTypes` for each lot.

Once this and the corresponding API pull-request is merged we'll be able to create drafts for any lot and work through the flow up as far as questions with assurance, which will not validate yet (as the assurance questions don't yet exist on the pages).

### DIFF
--- a/app/new_service_manifest.yml
+++ b/app/new_service_manifest.yml
@@ -18,7 +18,17 @@
   name: Service type
   editable: true
   questions:
-    - serviceTypes
+    - serviceTypesIaaS
+-
+  name: Service type
+  editable: true
+  questions:
+    - serviceTypesSaaS
+-
+  name: Service type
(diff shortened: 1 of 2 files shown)


app/new_service_manifest.yml
```diff
@@ -18,7 +18,17 @@
   name: Service type
   editable: true
   questions:
-    - serviceTypes
+    - serviceTypesIaaS
+-
+  name: Service type
```
quis: Any reason these sections need to be repeated? Could we have one section with the three questions in?

TheDoubleK: Oh probably - I'll check it works now and update if it does.

PR #105, opened 2015-07-16
[#97631142] Validate required questions per page

TheDoubleK:
See: https://www.pivotaltracker.com/story/show/97631142

This adds a hidden field to `edit section` pages that lists all the questions on the page.  This way the API can check if there were required fields on the page that have not been submitted with the form, and if so return a validation error for the page.

It also adds a `percentage` macro to display the percentage box for the "service uptime" question.

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -113,11 +113,21 @@ def edit_section(service_id, section):
     return render_template(
         "services/edit_section.html",
         section=content.get_section(section),
+        page_questions=get_section_questions(content.get_section(section)),
         service_data=service,
         service_id=service_id,
         return_to=".edit_service",
+        dashboard=".list_services",
+        dash_label="Services",
         **main.config['BASE_TEMPLATE_DATA']
     )
+
+
+def get_section_questions(section):
+    questions = []
+    for question in section.get('questions', None):
+        questions.append(question.get('id', None))
+    return questions
 
 
 @main.route(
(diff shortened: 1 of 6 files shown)


robyoung:
I have added quite a lot of test and live code so it would be good if someone else could take a look at this as well.

app/main/views/services.py
```diff
@@ -102,24 +101,36 @@ def update_service_status(service_id):
...
     content = existing_service_content.get_builder().filter(service)
+    section = content.get_section(section)
+    if section is None:
+        abort(404)
 
     return render_template(
         "services/edit_section.html",
-        section=content.get_section(section),
+        section=section,
+        page_questions='|'.join(get_section_questions(section)),
```
allait: I'm having trouble figuring out what `page_questions` is for. It's set as a hidden input field whenever the template is rendered (even though we seem to have access to `section` which we can get the questions from), but then https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/105/files#diff-32bf67afda65ad26faef3bf3efd522b1R155 just removes it right away without using it. Am I missing something?

robyoung: That is a very good point. I have updated the PR to not pass the `page_questions` to the template and only use the ones from the section when needed.

app/main/views/services.py
```diff
@@ -314,117 +301,140 @@ def view_service_submission(service_id):
...
+
+
+def _is_boolean_type(key):
+    """Return True if a given key is a boolean type"""
+    return new_service_content.get_question(key)['type'] == 'boolean'
+
+
+def _get_formatted_section_data(section):
+    section_data = dict(
+        list(request.form.items()) + list(request.files.items())
```
allait: This explicitly adds files to the `section_data`, but since files are neither list nor boolean they don't end up in the return value anyway

tests/app/main/test_services.py
```diff
@@ -282,11 +284,121 @@ def test_should_redirect_to_login_if_not_logged_in(self, data_api_client):
...
+        }
+    }
+
+    def setup(self):
+        super(TestEditService, self).setup()
+        with self.app.test_client():
+            self.login()
+
+    def test_questions_for_this_section_can_be_changed(self):
+        with mock.patch('app.main.views.services.data_api_client') as data_api_client:
```
allait: Any reason why this is not a class-level decorator instead? As far as I can see every test is using this `with` for the full method body

robyoung: No, absolutely not. I meant to do that.

allait:
:+1:

PR #107, opened 2015-07-20
Ensure cookie page uses local cookie page

minglis:
- seen cookie message pointed to GOV.UK
- changed to use /cookies
- note on local builds this will not work as cookie pages served from buyer app


quis:
:+1:

PR #111, opened 2015-07-23
Make link to service relative to environment

quis:
Currently the "view service" link in the admin app always points at live. It should be relative to the environment that the app is being used it (so that it can be tested).

Actual: https://www.digitalmarketplace.service.gov.uk/service/5126395806089216

Expected: https://preview.development.digitalmarketplace.service.gov.uk/service/5126395806089216

Similar to: https://www.pivotaltracker.com/story/show/94162812


robyoung:
The expected path doesn't match the code but it does redirect to that one so :+1:

PR #112, opened 2015-07-23
[#98628954] Build user list

pcraig3:
**Everything is different.**

~~Listing users with their names, email addresses, last logged in dates, and whether they're locked or not.
Using markup similar to the services listing, which means I'm not using the toolkit.~~

This pull request brings in a listing of <del>users</del> contributors on the supplier dashboard as well as a <del>user</del> contributor management page.
A logged-in supplier can see all of the other user accounts associated with its supplier, and (optionally) deactivate all of them.  A logged-in supplier can't deactivate themselves.

The 'Add or remove' link on the dashboard as well as the GET and POST routes in `user.py` are behind a feature flag, but the initial list of users on the dashboard isn't.

#### Notes:
1. I've changed 'users' to 'contributors' wherever it was printed onscreen, but I've not changed the url paths.  Should I?  Or do we want to call them something other than contributors, even?
2. I've just written the status message [in the route](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/pc_unconceal_users/app/main/views/users.py#L86).  Do we put them in other places?
3. [Error messages](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/pc_unconceal_users/app/main/views/users.py#L49).  Is there a format we like?

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/98628954)
[>> @ralph-hawkins's prototype](http://dm-prototype.herokuapp.com/dashboard-framework-open-new-supplier)

***

#### Supplier dashboard with list of contributors
*<sub>[deprecated screenshot](https://cloud.githubusercontent.com/assets/2454380/8855019/d75f3166-315a-11e5-9eeb-62aebad2feed.png)</del>*
screen shot 2015-07-28 at 15 26 53

#### Contributor page (before deactivating an account)
*<sub>deprecated screenshots: [v1](https://cloud.githubusercontent.com/assets/2454380/8855023/db649120-315a-11e5-8825-53ab1f9a56db.png), [v2](https://cloud.githubusercontent.com/assets/2454380/8934398/592674ce-353f-11e5-98b5-e5a08b3b06eb.png)</sub>*
screen shot 2015-08-03 at 12 27 56

#### After deactivating an account
*<sub>[deprecated screenshot](https://cloud.githubusercontent.com/assets/2454380/8934398/592674ce-353f-11e5-98b5-e5a08b3b06eb.png)</sub>*
screen shot 2015-08-03 at 12 28 09


robyoung:
The couple of things:
1. I agree, I think we should use the same format strings between the different apps. We have some  [format strings set in utils](https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/formats.py#L2).
2. If it's identical could it live in the frontend toolkit? cc/ @tombye @quis 

In the third column locked is aligned right and the timestamp is aligned left. Is that intentional?

app/templates/services/list_services.html
```diff
@@ -35,8 +35,8 @@
   <div class="grid-row">
     <div class="column-two-thirds">
       {% with
-        context = current_user.email_address,
-        heading = current_user.supplier_name
+        context = current_user.supplier_name,
+        heading = "Current services"
```
robyoung: Yes!

tests/app/main/test_users.py
```diff
@@ -0,0 +1,50 @@
...
+                    'supplier': {
+                        'name': "Supplier Name",
+                        'supplierId': 1234
+                    }
+                }]
+            }
+
+            res = self.client.get('/suppliers/users')
+            assert_equal(res.status_code, 200)
+            data_api_client.find_users.assert_called_once_with(
```
robyoung: This doesn't need to change I'm just pointing it out because I remembered it yesterday. We now have a line length limit of 120 characters so this could be one line.

tests/app/main/test_users.py
```diff
@@ -0,0 +1,50 @@
...
+                "Don Draper",
+                "don@scdp.com",
+                "Wednesday, 06 May 2015 at 11:57"
+            ]:
+                assert_in(
+                    self.strip_all_whitespace(
+                        # class names are slightly different for different cells >.<
+                        '{}</td>'.format(string)
+                    ),
+                    self.strip_all_whitespace(res.get_data(as_text=True))
```
robyoung: I think these would be be a bit clearer with lxml.

quis:
> 2. If it's identical could it live in the frontend toolkit? cc/ @tombye @quis
I'm OK with this not going into the toolkit. Very small amount of code and specific to just this page.

> In the third column locked is aligned right and the timestamp is aligned left. Is that intentional?
It's two columns.

quis:
@pcraig3 I took the liberty of adding the latest toolkit tables in f07bf7c.

app/main/helpers/helpers.py
```diff
@@ -0,0 +1,20 @@
+from flask import abort
+from dmutils.apiclient import APIError
+
+from ... import data_api_client
+
+
+def call_data_api_client(method_name, **kwargs):
```
robyoung: I don't think this is a good approach, it means we have to change how we call the API client everywhere. I'd prefer something like @allait suggested where we do this with an error handler. see: http://flask.pocoo.org/docs/0.10/patterns/apierrors/

pcraig3: Okay! Added an error handler that calls our 500-level error page.

robyoung:
The link says "Add or remove" but on the page we can only view and remove. Otherwise it looks good to me.

pcraig3:
The link is behind a feature flag, and Martyn's pull request is coming in soon afterwards.  I'll ask Martyn what he thinks.

robyoung:
:+1:

PR #113, opened 2015-07-24
Make URLs for assets cache-proof

quis:
This commit adds a query string to assets URLs which is generated from a hash of the file contents. When asset files are changed they will now be served from a different URL, which means they wont be loaded from browser cache.

This is similar to how GOV.UK template adds its version number as a querystring parameter for its assets, and matches how the buyer app does it too.

Addresses [98936730](https://www.pivotaltracker.com/story/show/98936730)

### DIFF
--- a/config.py
+++ b/config.py
@@ -1,6 +1,15 @@
 import os
+import hashlib
 import jinja2
 from dmutils.status import enabled_since, get_version_label
+
+
+def get_asset_fingerprint(asset_file_path):
+    hasher = hashlib.md5()
+    with open(asset_file_path, 'rb') as asset_file:
+        buf = asset_file.read()
+        hasher.update(buf)
+    return hasher.hexdigest()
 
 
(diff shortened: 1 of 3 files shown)


config.py
```diff
@@ -29,7 +38,11 @@ class Config(object):
     ASSET_PATH = STATIC_URL_PATH + '/'
     BASE_TEMPLATE_DATA = {
         'asset_path': ASSET_PATH,
-        'header_class': 'with-proposition'
+        'header_class': 'with-proposition',
+        "asset_fingerprints": {
+            "css": get_asset_fingerprint("app/static/stylesheets/application.css"),
```
robyoung: What if `application-ie6.css` changes but `application.css`? Looks like the fingerprint wouldn't change.

quis: You're right. Going to move this out to utils and then reopen this PR.

PR #115, opened 2015-07-24
Ensure user is logged out when their account is deactivated.

minglis:
The flask_login method that loads the user on every request now checks the active / locked status of the user and will log the user out if that status is invalid.

This now ensures that if some one gets disabled by an admin, they are immediately logged out.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -59,8 +59,11 @@ def create_app(config_name):
 @login_manager.user_loader
 def load_user(user_id):
     user_json = data_api_client.get_user(user_id=int(user_id))
+
     if user_json:
-        return User.from_json(user_json)
+        user = User.from_json(user_json)
+        if user.is_active() and not user.is_locked():
+            return user
 
 
 def config_attrs(config):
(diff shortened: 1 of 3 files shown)


minglis:
Note requires that we have a dependency on merging utils

allait:
If the user is deactivated and then reactivated while the session cookie is still valid will they appear to be logged in?

minglis:
@allait nope - logged out and stay out.

app/__init__.py
```diff
@@ -60,8 +60,11 @@ def refresh_session():
 @login_manager.user_loader
 def load_user(user_id):
     user_json = data_api_client.get_user(user_id=int(user_id))
+
     if user_json:
-        return User.from_json(user_json)
+        user = User.from_json(user_json)
+        if user.is_active() and not user.is_locked():
+            return user
```
robyoung: I've done a bit of investigating and we can just do `return User.from_json(user_json)` but it would require a couple of changes in digitalmarketplace-utils. Could we catch up this tomorrow?

robyoung: User could be just returned if https://github.com/alphagov/digitalmarketplace-utils/pull/110 gets merged.

minglis:
Is this redundant now the utils is merged? Should I close it?

robyoung:
Talked to @minglis about this in meat space and it was a :+1:

PR #117, opened 2015-07-27
Use new tables for listing existing services

quis:
image


allait:
Tests are failing, but the change looks good :+1:

PR #121, opened 2015-07-27
Add utils-based cache busting of assets

quis:
Brings in https://github.com/alphagov/digitalmarketplace-utils/pull/102


allait:
:+1:

PR #122, opened 2015-07-27
Don't lowercase service IDs

quis:
- Service IDs are stored as uppercase
- Old URLs may be in lower case but they are in the `/services/x-xxx-xxx` format,
  and the redirect uppercases them to `/g-cloud/services/X-XXX-XXXX`


allait:
:+1: :shipit:

PR #124, opened 2015-07-28
Rewrite supplier edit forms with toolkit form patterns

allait:
Fixes `"None"` strings in address/contact input fields by setting
`value=value or ''`. Jinja renders missing variables as empty
strings, but "None" gets converted to string. We've only seen this
happen with contact details since other fields shouldn't have `None`
values.

There's an interesting interaction between macros and the context of
variable-based pattern templates: macro arguments are visible to the
included template as normal variables, so as long as macro argument
name matches the variable name used by the pattern template there's
no need to set it using `{% with name=name %}`.

### DIFF
--- a/app/templates/macros/forms.html
+++ b/app/templates/macros/forms.html
@@ -1,54 +1,36 @@
-{% macro question_input(name, label, value='', hint='', errors=None) -%}
-  {{ _question_header(name, label, hint, errors) }}
-    <input type="text" class="text-box" name="{{ name }}" id="{{ name }}-input" value="{{ value }}" />
-  {{ _question_footer(name, label, hint, errors) }}
+{% macro question_input(name, question, value='', hint='', errors=None) -%}
+  {% with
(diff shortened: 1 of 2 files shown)


robyoung:
As our use of macros starts getting more involved it would be good to get some test coverage on these.

allait:
@robyoung these ones don't have any logic (unlike the ones they're replacing), so it'd be mostly testing the toolkit. And since this is frontend it's better served by the functional tests: there's more code related to using the macro than the macro itself, so it seems better to test that everything is in place on a rendered page and that errors show up in the right place.

robyoung:
I'm not sure I agree with your second point. The macros are small units of code that are used from multiple locations. I'd be more the side of testing them in isolation and then testing the user flows in the functional tests in a way that is agnostic to how the pages have been generated. Good point about the frontend toolkit, that would be the better place.

robyoung:
:+1:

PR #125, opened 2015-07-28
Service use template inheritance

robyoung:
Refactor the service overview and edit section templates for services and service submissions so that they inherit from common base templates. This means that we have less duplication and can avoid passing configuration variables into the templates. It does add a bit of indirection that may make the templates less obvious.

### DIFF
--- a/app/templates/services/_base_edit_section_page.html
+++ b/app/templates/services/_base_edit_section_page.html
@@ -1,0 +1,42 @@
+{% extends "_base_page.html" %}
+{% import "macros/toolkit_forms.html" as forms %}
+
+{% block page_title %}{{ section.name }} – Digital Marketplace{% endblock %}
+
+{% block main_content %}
+
+  {% if errors %}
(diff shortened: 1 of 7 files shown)


app/templates/services/_base_edit_section_page.html
```diff
@@ -0,0 +1,42 @@
...
+      {% with
+        heading = section.name,
+        smaller = true
+      %}
+        {% include 'toolkit/page-heading.html' %}
+      {% endwith %}
+
+    </div>
+  </div>
+  <form method="post" action="">
```
quis: > The action and formaction content attributes, if specified, must have a value that is a valid non-empty URL

https://html.spec.whatwg.org/multipage/forms.html#attr-fs-action

> The best thing you can do is leave out the action attribute altogether. If you leave it out, the form will be submitted to the document's address, i.e. the same page.

http://stackoverflow.com/a/1132015

allait:
:+1:

PR #127, opened 2015-07-29
Updated 503 tests

tombye:
The 503 test needed to stub out the `get_supplier` method because it is the only method of the API client the view code for the base ('/') suppliers route calls.

In addition to that, failed requests handled by the API Client return instances of the `HTTPError` class, not its derived class, `APIError` so the existing code behaved unpredictably.


robyoung:
`HTTPError` extends from `APIError` so using `APIError` would also catch `InvalidResponse`.

robyoung:
:+1:

PR #128, opened 2015-07-29
Mark all external links in footer as such

tombye:
Brings in work done in the buyers' app in https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/110.


quis:
image

:cry:

tombye:
Dry those tears, I added the fix in 23bca76

PR #130, opened 2015-07-30
Pass through 'hint' content for all question types

quis:
image

### DIFF
--- a/app/templates/macros/toolkit_forms.html
+++ b/app/templates/macros/toolkit_forms.html
@@ -2,7 +2,7 @@
   {%
     with
     question=question_content.question,
-    hint=question_content.hint or None,
+    hint=question_content.hint,
     name=question_content.id,
     value=service_data[question_content.id],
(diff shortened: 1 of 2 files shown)


app/templates/macros/toolkit_forms.html
```diff
@@ -2,6 +2,7 @@
   {%
     with
     question=question_content.question,
+    hint=question_content.hint or None,
```
allait: `or ''`, but toolkit really should check for `{% if hint %}` instead of `if hint is defined`,  otherwise we'll get an empty hint span for all questions

quis: Let's fix the toolkit first then. I'll look into it.

quis: @allait https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/126

allait:
:+1:

PR #131, opened 2015-07-30
Invite supplier users

minglis:
Adds ability to invite users from the supplier app.

Note this depends on: 
https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/112
As the invite user form should be on the pages in that pull request

And
https://github.com/alphagov/digitalmarketplace-utils/pull/111
As email integration is now in utils

It delivers:
https://www.pivotaltracker.com/n/projects/997836/stories/98628914

Premise:
Existing supplier is able to invite a user to contribute to the company:
- enters email address
- email is sent via mandrill
- content signed of by @superroz 

Email has a link generated in the same way as the password reset. Link contains
- supplier_id
- supplier_name
- email_address of invited user

When link clicked takes user to either
- Update page - if user exits we allow a one button click to associated that account with the supplier in the token
    - Locked, inactive or existing supplier accounts are all in error
- Create page - if no user with that email exists we allow the user to enter a name and password and create their account

- Both forms will log the user in and redirect to the dashboard

Create User:
create_contributor_account_ _digital_marketplace

Update user:
create_contributor_account_ _digital_marketplace


minglis:
needs utils merge to pass. still up for a review

minglis:
What do people think about having the data api send emails? That would allow us to have email code in one place again, and also make it easier to audit the sending of emails without creating a "create-audit" endpoint in the API.

app/main/views/login.py
```diff
@@ -146,9 +182,178 @@ def update_password(token):
...
+    if form.validate_on_submit():
+
+        token = generate_token(
+            {
+                "supplier_id": current_user.supplier_id,
+                "supplier_name": current_user.supplier_name,
+                "email_address": form.email_address.data
+            },
+            current_app.config['SECRET_KEY'],
+            current_app.config['RESET_PASSWORD_SALT']
```
allait: Salts are used for token namespacing, so this should be different from the reset password one

app/main/views/login.py
```diff
@@ -146,9 +180,178 @@ def update_password(token):
...
+                "auth/create-user.html",
+                form=form,
+                email_address=token['email_address'],
+                supplier_name=token['supplier_name'],
+                token=encoded_token,
+                **template_data), 200
+
+
+@main.route('/update-user/<string:encoded_token>', methods=["POST"])
+def submit_update_user(encoded_token):
```
allait: This endpoint lets you log in without ever entering your password.

I think it should also check that the existing user has a buyer role, otherwise it makes it possible to reattach supplier (and admin) accounts by emailing everyone a link using our own form.

config.py
```diff
@@ -75,6 +80,7 @@ class Test(Config):
 class Development(Config):
     DEBUG = True
     SESSION_COOKIE_SECURE = False
+    WTF_CSRF_ENABLED = False
```
allait: I'm not sure this is a good idea. It hides errors related to the `csrf_token`, which I believe the new forms are missing.

app/templates/auth/update-user.html
```diff
@@ -0,0 +1,44 @@
...
+  with
+  smaller = true,
+  heading = "Create contributor account for  " ~ supplier_name,
+  context = email_address
+%}
+  {% include "toolkit/page-heading.html" %}
+{% endwith %}
+
+    <form autocomplete="off" action="{{ url_for('.submit_update_user', encoded_token=token) }}" method="POST" id="updateUserForm">
+        <button class="button-save">Create contributor account</button>
```
allait: I think this form is missing a `csrf_token`

app/main/views/login.py
minglis: The link prior to this blocks - so you don't see the create/update pages unless you're in a valid state.

Will add a check in here though in case some one just posts to this directly

minglis: With regards to login - you need to have a valid token to be able to see this page. You've been invited and the token is emailed to you. Once thats happened a click through here will log you in. 

A check in here to ensure valid user status (active/not a supplier/etch) should be OK.

minglis: More checks added to blow it up if bad user tries to submit to either update or create user

app/main/views/login.py
```diff
@@ -171,3 +386,32 @@ def decode_password_reset_token(token):
         return None
 
     return decoded
+
+
+def decode_invitation_token(encoded_token):
+    try:
+        token, timestamp = decode_token(
+            encoded_token,
+            current_app.config['SHARED_EMAIL_KEY'],
```
robyoung: Why is this called `SHARED_EMAIL_KEY` when it's not shared with anything?

app/main/helpers/email.py
```diff
@@ -1,63 +0,0 @@
...
-                       salt=main.config["RESET_PASSWORD_SALT"],
-                       max_age=86400, return_timestamp=True)
-    return decoded
-
-
-def token_created_before_password_last_changed(token_timestamp, user):
-
-        password_last_changed = datetime.strptime(
-            user['users']['passwordChangedAt'], DATETIME_FORMAT)
-        return token_timestamp < password_last_changed
```
robyoung: This function has been changed to now just check two timestamps. It feels like a nice abstraction if it takes a `User` hash but a bit redundant if it's just checking two timestamps.

app/templates/auth/update-user.html
```diff
@@ -0,0 +1,45 @@
...
+
+{% block main_content %}
+
+{% if not token %}
+
+<header class="page-heading-smaller">
+  <h1>This link is invalid</h1>
+</header>
+<p>
+    Check you&#146;ve entered the correct link or ask the person who invited you to send a new invitation.
```
robyoung: Shouldn't this be an apostrophe?

app/templates/auth/submit-email-address.html
```diff
@@ -0,0 +1,57 @@
...
+            {{ form.email_address(class="text-box", autocomplete="off") }}
+        </div>
+    {% if form.email_address.errors %}
+        </div>
+    {% endif %}
+
+    <button class="button-save">Invite</button>
+</form>
+
+{% endblock %}
```
robyoung: brrp, this is not a text file

app/templates/auth/create-user.html
```diff
@@ -0,0 +1,76 @@
...
+{% block page_title %}Create contributor account – Digital Marketplace{% endblock %}
+
+{% block main_content %}
+
+{% if not token %}
+<header class="page-heading-smaller">
+  <h1>This link is invalid</h1>
+</header>
+<p>
+    Check you&#146;ve entered the correct link or ask the person who invited you to send a new invitation.
```
robyoung: I think this should be an apostrophe as well.

app/main/views/login.py
minglis: Shared with admin app

app/templates/auth/update-user.html
minglis: Needs to be a UTF-8 curly apostrphe - this is the encoded version

app/main/views/login.py
robyoung: You beat me to it! :grin: I was just checking your other pull request and realised.

app/templates/auth/update-user.html
robyoung: You can use UTF-8 characters in Jinja2 so I think either this should be a UTF-8 character or the apostrophe on https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/131/files#diff-fd6ab505bb4711c18d7b63bd49b6e2e7R14 should be changed.

tests/app/main/test_login.py
```diff
@@ -305,14 +409,724 @@ def test_reset_password_form_and_inputs_not_autofillable(
...
+            )
+            assert_equal(res.status_code, 400)
+            assert_equal(data_api_client.get_user.called, False)
+            assert_equal(data_api_client.get_supplier.called, False)
+
+    @mock.patch('app.main.views.login.generate_token')
+    def test_should_be_a_bad_request_if_token_expired(self, generate_token):
+        with self.app.app_context():
+
+            generate_token.side_effect = BadTimeSignature("Too old")
```
robyoung: This should be mocking `decode_invitation_token` here. As it stands the 400 is because the token is invalid.

tests/app/main/test_login.py
```diff
@@ -305,14 +409,724 @@ def test_reset_password_form_and_inputs_not_autofillable(
...
+                self.app.config['SHARED_EMAIL_KEY'],
+                self.app.config['INVITE_EMAIL_SALT']
+            )
+
+            res = self.client.post(
+                '/suppliers/create-user/{}'.format(token),
+                data={
+                    'password': None,
+                    'name': None
+                }
```
robyoung: I don't think this would ever happen. Either the password and name are not submitted, in which case the data dictionary would be empty, or they are submitted empty, in which case they would have empty strings as values.

tests/app/main/test_login.py
```diff
@@ -305,14 +409,724 @@ def test_reset_password_form_and_inputs_not_autofillable(
...
+                    'password': "123456789",
+                    'name': ""
+                }
+            )
+
+            assert_equal(res.status_code, 400)
+            assert_equal(
+                "Please enter a name" in res.get_data(as_text=True),
+                True
+            )
```
robyoung: This is a strange assertion. `assert_true` seems more appropriate, or even `assert_in`.

app/main/helpers/email.py
minglis: Yeah, it got really changed, Only value is in naming that check really. Makes the calling method slightly easier to read.

tests/app/main/test_login.py
minglis: fixed. *guilty face* been doing that a bit, will remember and fix 'em when I see 'em

tests/app/main/test_login.py
minglis: Yep. Did this with empty dictionaries elsewhere. My bad. Fixed

app/templates/auth/update-user.html
minglis: Done. UTF-8 ones broke tests. Just read up on file encodings to make it work.

tests/app/main/test_login.py
minglis: Token is valid, but missing correct keys. `decode_invitation_token` throws a ValueError in the block

` if not token.get('email_address', None):
            raise ValueError('Missing email address from token')
`

Test is to check response to a valid token, which has missing data.

minglis: Though made it nicer

app/main/views/login.py
```diff
@@ -146,9 +180,184 @@ def update_password(token):
...
+            "auth/update-user.html",
+            token=None,
+            email_address=None,
+            supplier=None,
+            **template_data), 400
+    else:
+        user = data_api_client.get_user(email_address=token.get("email_address"))
+
+        user = User.from_json(user)
+        if user.is_locked() or not user.is_active() or user.supplier_id is not None:
```
allait: I think checking for "buyer" role would fit better, since it would also exclude admin accounts

minglis: Users don't have a "role" in utils yet. Will need to add it.

minglis: Didn;t see the "user_has_role" method. Used this.

app/main/helpers/email.py
robyoung: The only place this is used is https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/131/files#diff-fbfacb66b19509ef6d34a19cc59dd93bR378 and the `user` object isn't used anywhere else in that function. So could the user / datetime parsing code be moved back into this function? That seems to be a nicer abstraction.

minglis: Done

PR #132, opened 2015-07-31
Error page titles on supplier app updated

minglis:
- now match what the buyer app does

### DIFF
--- a/app/templates/errors/404.html
+++ b/app/templates/errors/404.html
@@ -1,5 +1,8 @@
 {% extends "_base_page.html" %}
 
+{% block page_title %}Page could not be found - Digital Marketplace{% endblock %}
+
+
 {% block main_content %}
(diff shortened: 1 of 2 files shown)


tombye:
LGMT though it would be good to make these pages match those on the admin and buyers' apps:

- https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/master/app/templates/errors/500.html
- https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/master/app/templates/errors/404.html
- https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/master/app/templates/errors/404.html
- https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/master/app/templates/errors/500.html

tombye:
:+1:

PR #133, opened 2015-07-31
Include fix for bug with toolkit summary tables

quis:
Dashboard was throwing a 500 for suppliers with clients.

This was because of a mis-named variable in one of the toolkit templates, fixed in:
https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/127/files

This commit updates the toolkit version to include the fix.


allait:
:+1:

PR #135, opened 2015-08-04
Fix variable name in template

pcraig3:
Accidentally a word in the `list_users` template.

It doesn't cause any practical problems at the moment, but it could eventually.


quis:
:+1:

PR #137, opened 2015-08-04
Documents upload

allait:
## Enable document upload from edit draft section page
Validates files using dmutils to check file size and document type
and uploads to S3 if there were no errors.

Uploaded file URLs are sent to the API with the rest of form data
for validation.

If other fields fail validation the form is displayed again with
errors.

### Don't display uploaded file URL if form failed validation
If other form fields on the same page fail validation the file
has to be attached again, so file URLs shouldn't be added to the
form data.

### Filter file uploads from request data
File URLs shouldn't be submitted directly since there's no validation
that the URL matches the file uploaded to S3.

### Save submission documents to S3_DRAFT_DOCUMENTS_BUCKET
Draft bucket is intended for drafts that don't yet have a service ID.

Since the documents will have to be copied when the draft gets a
service ID storing them separately makes it easier to clean up
unused files after the framework goes live and makes sure the files
are not accessible through the assets domain.

### Upload draft documents as private
Documents should not be publicly visible before the framework goes
live.

### Save info page as the uploaded draft document URL
Uses a buyer frontend page as a base URL for the document uploads
during the service submission. The URL links to an info page that
displays some text about the state of the document and the reason
it can't be displayed before the framework goes live.

Since the documents will have to be moved once the draft gets a
service ID the URLs will change anyway and temporary base URL
doesn't matter.

## Move document questions to separate sections
After looking at different ways to do data + files validation
flow it looks like moving each document upload to a separate
page is the most reasonable option right now.

Mixing documents with other fields on a single form behaves
differently from a simple form:
* If the file is uploaded first but sent to the API together
  with the form fields then, errors in other fields fail the
  form and the file URL is not saved
* If the uploaded file URL is saved to form data and sent
  back after the user resubmits the form then the URL needs
  to be validated (which we don't need to do otherwise).
  If the user closes the page without submitting the form the
  file URL is not saved to the API (which is consistent with
  how other forms behave, but might be different from what
  you'd expect from a file upload field)
* If the file is uploaded and the URL is saved to the API before
  other form fields are sent to the API then the form behaves
  differently from other sections (field value is saved even
  though the form submission showed errors)

The main downside to current approach is that some documents are
optional (SFIA rate card), which means that the whole section can
be skipped, but there's no indication of that on the page.

## Restore DM_HTTP_PROTO config variable
Fixes an issue with password reset links starting with 'http://'.
Adds common base config for live environments.

### Add a flash message when creating a draft copy

### DIFF
--- a/app/new_service_manifest.yml
+++ b/app/new_service_manifest.yml
@@ -36,7 +36,15 @@
     - educationPricing
     - trialOption
     - freeOption
+-
+  name: Pricing document
+  editable: true
+  questions:
     - pricingDocumentURL
+-
+  name: SFIA rate card
+  editable: true
+  questions:
     - sfiaRateDocumentURL
 -
   name: Terms and conditions
@@ -44,6 +52,10 @@
   questions:
     - terminationCost
     - minimumContractPeriod
+-
+  name: Terms and conditions document
+  editable: true
+  questions:
     - termsAndConditionsDocumentURL
 -
   name: Support
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -3,7 +3,6 @@ import re
 from datetime import timedelta
 from flask import Flask, request, redirect, session
 from flask_login import LoginManager
-from flask._compat import string_types
 from flask_wtf.csrf import CsrfProtect
 
 from dmutils import apiclient, init_app, flask_featureflags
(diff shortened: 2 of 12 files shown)


robyoung:
:+1:

app/new_service_manifest.yml
```diff
@@ -36,7 +36,15 @@
     - educationPricing
     - trialOption
     - freeOption
+-
+  name: Pricing document
```
TheDoubleK: The file upload pages should be the last four pages (to avoid people getting blocked half way through submitting a service). I think the best ordering for them would be: Service definition, T&Cs, Pricing, SFIA rate card.

pcraig3: Every time I see that acronym, I think of the calling card my mom uses to phone overseas: the [SIFA card](http://www.sifacallingcard.com/).

PR #138, opened 2015-08-04
Move file uploads to the end of the question sequence

TheDoubleK:
We don't want users to be blocked part-way through completing details of a service because they don't have the right document to hand.  By putting documents at the end of the question flow they can answer all questions without having the documents to upload.


robyoung:
:+1:

robyoung:
How do I run the jasmine tests?

PR #139, opened 2015-08-04
Hide submission document urls

allait:
### Hide the document URL for uploaded G7 submission documents
Ralph suggested that document upload and draft section view pages
should show "A document was uploaded" instead of the info URL for
uploaded G7 documents.

Since we use the same macros and toolkit templates for both live
and submission services (and there's no easy way to split them
that I can think of) this adds an additional check to the `upload`
macros that checks if the document is attached to a submission
draft (by checking that the URL starts with an "info" URL prefix).

Also removes quotes from the "service copied" message


quis:
:+1:

PR #140, opened 2015-08-05
Fix checkboxes questions

quis:
The 'type' of multiple choice answers is 'checkboxes' plural.

[See the schema here](https://github.com/alphagov/digitalmarketplace-ssp-content/blob/master/tests/schemas/question.json#L11)


allait:
:+1:

PR #141, opened 2015-08-05
Fix document form rendering failing when missing previous value

allait:
Fixes a 500 error in release-139


quis:
:+1: When the test pass

PR #142, opened 2015-08-05
Order of supplier details on supplier account

ralph-hawkins:
Works better with account creation to show information we’ve asked for first.


quis:
LGTM :+1:

PR #143, opened 2015-08-05
Fix bug when assurance question is not answered

quis:
If a user tries to save a page but doesn't answer all the assurance questions the following error occurs:

```
argument of type 'NoneType' is not iterable
```

This commit makes sure that assurance is never set to `None`, so that the template always has a value (albeit an empty one) to look `in` [here](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/toolkit/templates/forms/selection-buttons.html#L33).

### DIFF
--- a/app/main/helpers/services.py
+++ b/app/main/helpers/services.py
@@ -78,7 +78,7 @@ def get_formatted_section_data(section):
         if _has_assurance(key):
             section_data[key] = {
                 "value": section_data[key],
-                "assurance": request.form.get(key + '--assurance')
+                "assurance": request.form.get(key + '--assurance', '')
             }
 
     return section_data


allait:
Why is the toolkit template using `in`? There could be only one selected value for the radio buttons input, so `==` should work too, unless I'm missing something.

quis:
@allait The `selection-buttons` template is generic and shared between checkboxes and radio buttons

allait:
@quis oh yeah, I've seen the "radio" types for Yes/No and didn't notice that we're using a different part of the template. 

I still think this is a template issue: it shouldn't be using `in` checks for radio buttons, since radio buttons are strings and if one of the options is a substring of another the check will pass twice. Is there a way to make the template smarter, perhaps by performing different check based on `type`?

quis:
@allait Sounds reasonable. Going to close this in favour of https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/132

PR #145, opened 2015-08-05
Clean up supplier app

quis:
## Remove old code

A bunch of code in the supplier app is no longer actually used, or can be trivially replaced with a newer way of doing things.

This commit removes the things that are easy to remove.

## Fix bug when assurance question is not answered …

If a user tries to save a page but doesn't answer all the assurance questions the following error occurs:

```
argument of type 'NoneType' is not iterable
```

This commit updates the version of the toolkit to include alphagov/digitalmarketplace-frontend-toolkit#132, which fixes this issue.

## Restrict width of hints, labels and error messages …

To maintain a reasonable line length this commit restricts the above elements to 66% wide.

image


minglis:
👍

PR #147, opened 2015-08-05
Make Heroku-friendly

robyoung:
This adds the changes @minglis put together for running frontend apps in heroku. The intention is to separate it out from the pull request on supplier declarations.

Other than adding heroku specific things like buildpacks, Procfile and README changes this also includes;
- Allow the port the app runs on to be controlled by an environment
  variable
- Bind to 0.0.0.0 rather than the default 127.0.0.1
- Change how the npm post install is called

### DIFF
--- a/application.py
+++ b/application.py
@@ -4,6 +4,8 @@ import os
 from app import create_app
 from flask.ext.script import Manager, Server
 
+port = int(os.getenv('PORT', 5003))
+
 application = create_app(
     os.getenv('DM_ENVIRONMENT') or 'development'
 )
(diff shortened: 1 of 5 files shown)


quis:
:+1:

application.py
```diff
@@ -14,7 +16,7 @@
 }
 
 manager = Manager(application)
-manager.add_command("runserver", Server(port=5003))
+manager.add_command("runserver", Server(host='0.0.0.0', port=port))
```
allait: `python application.py runserver` accepts port and host as options, so we can set them in Procfile instead and there's no need for a custom ENV variable

allait:
:+1:

PR #148, opened 2015-08-06
Generate a signed S3 document URL for submission documents

allait:
Removes the "A document was uploaded" summary table text in favour
of a temporary URL to the uploaded document.

Only the logged in users attached to the supplier that uploaded the
document can generate the link. Generated links lead directly to S3
and expire in 30s.


quis:
30s seems kind of short. This is average time on page (`h:m:s`) for the 10 pages in the SSP that had the most time spent on them:

image

allait:
@quis the URL is generated by the `/submission/documents/` when the user clicks on the document link, so it's time for the redirect response and connection to S3 only. I think it doesn't even include the download time, so 30s should be enough for slow connections too.

quis:
@allait Ah, that's fine then.

robyoung:
:+1:

pcraig3:
Got it to work :+1:

PR #151, opened 2015-08-06
Standardise markup for supplier dashboard.

pcraig3:
Markup for the 'contributors' section of the dashboard didn't have the same `column-one-whole` tags as the rest of the sections.
Upon further discussion, all of those tags were unnecessary.

### DIFF
--- a/app/templates/suppliers/dashboard.html
+++ b/app/templates/suppliers/dashboard.html
@@ -27,103 +27,94 @@
       {% endwith %}
     </div>
   </div>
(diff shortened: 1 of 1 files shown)


quis:
```HTML
<div class="grid-row">
    <div class="column-one-whole">
```

None of these full width column bits are needed.

pcraig3:
Right, back-pedalled everything.

quis:
Git has a real hard time diffing that. But :+1:

PR #152, opened 2015-08-06
Fix caption on supplier dashboard

pcraig3:
One of the tables had the wrong caption.
Probably the first time CPDD has proved to be problematic.


quis:
:+1:

PR #154, opened 2015-08-07
BUGFIX: serviceTypes isn't a real question

robyoung:
This was causing a KeyError on the Service type page because the field name `serviceTypes` isn't really a question. There are specific question types for each of the different lots.

### DIFF
--- a/app/main/helpers/services.py
+++ b/app/main/helpers/services.py
@@ -241,7 +241,7 @@ def _is_pricing_type(key):
 
 def _is_type(key, *types):
     """Return True if a given key is one of the provided types"""
-    return new_service_content.get_question(key)['type'] in types
+    return new_service_content.get_question(key).get('type') in types
 
 
 def _has_assurance(key):


app/main/helpers/services.py
```diff
@@ -241,7 +241,7 @@ def _is_pricing_type(key):
 
 def _is_type(key, *types):
     """Return True if a given key is one of the provided types"""
-    return new_service_content.get_question(key)['type'] in types
+    return new_service_content.get_question(key).get('type') in types
```
allait: I thought `serviceTypes` was handled by `_is_list_type` explicitly, and it's the first check in `get_formatted_section_data`, where is the error triggered from?

robyoung: `_is_pricing_type` is called from `pricing_question_filter` we could add a more specific check there but I think that this is actually correct. `_is_pricing_type("non_existent_key")` is `False`.

allait:
:+1:

PR #156, opened 2015-08-07
Go to services page after 'delete' and 'copy'

quis:
Previously all services were on the dashboard.

Now that they are on a separate page it makes sense for said page to be where the user sees the messages for 'deleted' and 'copied'.

It also feels like the next thing they do after deleting or copying a service would be accessible from the services page.

image


allait:
:+1:

PR #157, opened 2015-08-07
Add ITT supplier pack link endpoint

allait:
[#100213582](https://www.pivotaltracker.com/story/show/100213582)

Similar to submission service documents, /download-supplier-pack
generates a temporary signed S3 link for the supplier pack.

Supplier pack S3 path is always set to `g-cloud-7-supplier-pack.zip`.
`g-cloud-7` prefix allows to use a single location prefix in assets
nginx to redirect all relevant G7 URLs to the draft bucket.

Versioning is not implemented, but could achieved by relying on
`dmutils.s3.S3.move_existing` during service pack uploads, although
copied file name should use suffix instead of prefix to keep the
`g-cloud-7-...` name/location pattern if old versions need to be
accessible through the assets domain.

Supplier pack "last updated" date is static, but could be wired to
the S3 object "last modified" timestamp in the future.

Requires:
* [ ] https://github.com/alphagov/digitalmarketplace-aws/pull/129

### DIFF
--- a/app/main/helpers/services.py
+++ b/app/main/helpers/services.py
@@ -9,6 +9,11 @@ from dmutils.content_loader import PRICE_FIELDS
 
 from ...main import new_service_content
 
+try:
+    import urlparse
+except ImportError:
+    import urllib.parse as urlparse
+
 
 def get_service_attributes(service_data, service_questions):
     return map(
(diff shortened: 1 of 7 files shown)


quis:
:+1: When AWS PR is merged.

robyoung:
How can we test this locally?

allait:
There should be a file in the dev bucket at that path now.

(Which reminds me that since the supplier pack is a ZIP file we need to change the s3 upload to set the correct content-type for ZIPs)

robyoung:
Cat falling in to bath confirmed! :+1:

PR #158, opened 2015-08-07
Supplier declaration

robyoung:
https://www.pivotaltracker.com/story/show/100215022

This wires up the dummy supplier declaration page with real questions and makes use of the content loader.

This change also makes use of form parsing code that has now moved into `dmutils` (https://github.com/alphagov/digitalmarketplace-utils/pull/126).

Validation and reporting on whether the declaration is complete is part of another story (https://www.pivotaltracker.com/story/show/100215328).


app/__init__.py
```diff
@@ -53,6 +57,16 @@ def refresh_session():
         session.permanent = True
         session.modified = True
 
+    @application.template_filter()
+    @evalcontextfilter
+    def nl2br(eval_ctx, value):
+        result = value.replace('\n', '<br>\n')
```
allait: @quis says this isn't used anywhere

app/templates/macros/toolkit_forms.html
```diff
@@ -112,8 +112,8 @@
     priceUnit=service_data.priceUnit,
     priceInterval=service_data.priceInterval,
     value=service_data[question_content.id],
-    question=question_content.question,
-    hint=question_content.hint,
+    hint=(question_content.hint or '')|markdown,
+    hint=question_content.hint or ''|markdown,
```
quis: This line overwrites the previous one, with undesired consequences.

app/templates/macros/toolkit_forms.html
```diff
@@ -78,8 +78,8 @@
   {%
     with
     name=question_content.id,
-    question=question_content.question,
-    hint=question_content.hint,
+    question=question_content.question|markdown,
+    hint=(question_content.hint or '')|markdown,
     value=service_data[question_content.id],
     id=question_content.id,
     hint=question_content.hint,
```
quis: This line overwrites the previous one, with undesired consequences.

quis:
This should probably redirect to the dashboard page when you save.

allait:
:+1:

PR #160, opened 2015-08-10
Align current service editing with SSP

quis:
Add a 'view service' button from the admin app to the page for editing a current service

image

--

Add a 'view service' link to the table of current services and changes the name-of-service link to point to the edit page and simplifies the table of current services to show either the 'bad' status (private/removed) _or_ the link to view the service

image

### DIFF
--- a/app/templates/services/service.html
+++ b/app/templates/services/service.html
@@ -22,6 +22,13 @@
 {% endblock %}
 
 {% block before_sections %}
+  {% if service_data.status == "published" %}
+    <div class="column-one-third">
+      {% with label="View service", url="/g-cloud/services/" + service_id %}
+        {% include "toolkit/link-button.html" %}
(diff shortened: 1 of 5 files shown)


app/templates/services/service.html
```diff
@@ -22,6 +22,13 @@
 {% endblock %}
 
 {% block before_sections %}
+  {% if service_data.status == "published" %}
+    <div class="column-one-third">
+      {% with label="View service", url="/g-cloud/services/" + service_id %}
```
minglis: Can this use a url_for?

quis: Nope, different app

robyoung:
:+1:

pcraig3:
Seems more sensible than the "Edit" link at the right-hand side, so #win.

PR #161, opened 2015-08-10
Selection questions validation

robyoung:
See: https://www.pivotaltracker.com/story/show/100215328

Add validation for the selection questions. Validations are hard coded in the frontend. If any validations fail the form will not be submitted to the API. Once the questions have been successfully submitted the message on the dashboard will change to "You have made the declaration".

## Validations that apply
- All required fields must be filled in (some are optional, some are optional under certain conditions).
- Character limits must not be exceeded
- Formatting rules for DUNS and VAT numbers must be followed

## Related
https://github.com/alphagov/digitalmarketplace-ssp-content/pull/53

### DIFF
--- a/app/main/helpers/frameworks.py
+++ b/app/main/helpers/frameworks.py
@@ -8,32 +8,33 @@ from werkzeug.datastructures import ImmutableOrderedMultiDict
 def get_required_fields(all_fields, answers):
     required_fields = set(all_fields)
     #  Remove optional fields
-    optional_fields = [
+    optional_fields = set([
         "SQ1-1e", "SQ1-1f", "SQ1-1p-i", "SQ1-1p-ii", "SQ1-1p-iii", "SQ1-1p-iv",
-        "SQ1-1q-i", "SQ1-1q-ii", "SQ1-1q-iii", "SQ1-1q-iv"
-    ]
-    required_fields -= set(optional_fields)
+        "SQ1-1q-i", "SQ1-1q-ii", "SQ1-1q-iii", "SQ1-1q-iv", "SQ1-1cii", "SQ1-1i-ii",
+        "SQ1-1j-ii", "SQ4-1c", "SQ3-1k"
+    ])
+    required_fields -= optional_fields
     #  If you answered other to question 12
-    if answers.get('SQ1-1ci') != 'other (please specify)':
-        required_fields.remove('SQ1-1cii')
+    if answers.get('SQ1-1ci') == 'other (please specify)':
+        required_fields.add('SQ1-1cii')
     #  If you answered yes to question 20
-    if not answers.get('SQ1-1i-i', False):
-        required_fields.remove('SQ1-1i-ii')
+    if answers.get('SQ1-1i-i', False):
+        required_fields.add('SQ1-1i-ii')
     #  If you answered 'licensed' or 'a member of a relevant organisation' in question 22
-    if len(answers.get('SQ1-1j-i', [])) == 0:
-        required_fields.remove('SQ1-1j-ii')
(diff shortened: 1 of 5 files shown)


tests/app/main/helpers/test_frameworks.py
```diff
@@ -0,0 +1,182 @@
...
+    submission['SQ1-1g'] = 'invalid'
+    assert_equal(get_all_errors(content, submission), {'SQ1-1g': 'invalid_format'})
+
+
+def test_character_limit_errors():
+    cases = [
+        ("SQ1-1a", 255),
+        ("SQ1-1cii", 255),
+        ("SQ1-1d", 255),
+        ("SQ1-1i-ii", 4096),
```
quis: We should review these character limits. They were copied from Bravo Solution™ and feel quite arbitrary—they've been removed from the content already.

quis: As discussed in meatspace, character limits can be removed.

quis:
This is kind of grim:

image

I think it would be better if it just said
> - [Question 20](#)
> - [Question 47](#)
> - [Question 53](#)

instead.

robyoung:
Yes, good call, now looks like...
screen shot 2015-08-11 at 16 57 44

app/main/helpers/frameworks.py
```diff
@@ -0,0 +1,129 @@
...
+    return required_fields
+
+
+def get_all_fields(content):
+    return reduce(add, (section.get_question_ids() for section in content))
+
+
+def get_question_numbers(content):
+    numbers = {}
+    for section in content:
```
allait: You could use `get_all_fields` with `enumerate` here

app/main/views/frameworks.py
```diff
@@ -66,20 +76,26 @@ def framework_services():
 @flask_featureflags.is_active_feature('GCLOUD7_OPEN')
```
pcraig3: Is it easier to do `methods=['GET', 'POST']` here?  
I've generally been splitting them into different routes.

robyoung: Marginally as both render the same template. I'm not very opinionated either way but we should be consistent. Maybe something for the tech debt wall for us to decide and change wholesale?

app/main/helpers/frameworks.py
```diff
@@ -0,0 +1,121 @@
...
+        "SQ1-1a": 5000,
+        "SQ1-1cii": 5000,
+        "SQ1-1d": 5000,
+        "SQ1-1i-ii": 5000,
+        "SQ3-1k": 5000,
+    }
+    errors_map = {}
+    for field, limit in length_limits.items():
+        if len(answers.get(field, "")) > limit:
+            errors_map[field] = "under_character_limit"
```
pcraig3: So `under_character_limit` means they've exceeded the character limit, right?

app/main/helpers/frameworks.py
```diff
@@ -0,0 +1,130 @@
...
+def get_question_numbers(content):
+    numbers = {}
+    for section in content:
+        for question_id in section.get_question_ids():
+            numbers[question_id] = len(numbers) + 1
+    return numbers
+
+
+def get_fields_with_values(answers):
+    return set(key for key, value in answers.items() if not isinstance(value, six.string_types) or len(value) > 0)
```
allait: What about empty lists for `checkboxes` questions?

app/main/helpers/frameworks.py
robyoung: Yes, this matches the error key used for the SSP.

allait:
:+1:

PR #163, opened 2015-08-11
Make Gulp errors more detailed & like NPM ones

tombye:
image


quis:
I feel like it would read better/follow convention as
```
ERROR: invalid property name in /Users/tombyers/…/_single-question-page.scss:11
``

tombye:
Well the convention I was trying to follow was that of NPM (swapping out NPM for Gulp) so once you've got that pattern you recognise it:

image

Thinking about this, I've lowercased the 'gulp' to cause a bit less friction.

quis:
The NPM ones are kind of icky, but convention wins, :+1:

PR #165, opened 2015-08-12
Fix summary table caption on supplier dashboard

pcraig3:
All of our other summary tables have the table header as the caption.
The "Contributors" table should too.

##### Note
This change affects the [functional tests](https://github.com/alphagov/digitalmarketplace-functional-tests/pull/53) I've written, so I've labeled that PR as `[WIP]` until this goes through.

### DIFF
--- a/app/templates/suppliers/dashboard.html
+++ b/app/templates/suppliers/dashboard.html
@@ -122,7 +122,7 @@
   {% endif %}
   {% call(item) summary.table(
     users,
-    caption="Users for " + current_user.supplier_name,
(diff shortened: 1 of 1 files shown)


quis:
I think @tombye had some reckons about this

quis:
Rewind that last comment. Possibly removing the headings above tables in favour of captions is a wider change, shouldn't block the merging of this PR. :+1:

tombye:
For the HTML patterns we have at the moment, this is a sensible change and much better than what exists at the moment :+1:

PR #166, opened 2015-08-12
Remove stub pages which are not used anywhere

quis:
Both the things that would have gone on these pages are now part of the updates page.


minglis:
👍

PR #167, opened 2015-08-12
Create new company

minglis:
Create new company:

depends on:

https://github.com/alphagov/digitalmarketplace-api/pull/206

https://github.com/alphagov/digitalmarketplace-utils/pull/133

Create company flow:
Part 1 - Company

- DUNS number - has to be unique in DB - makes an API call to get suppliers by DUNS to check there isn;t a presisting one. Duns are numbers, 9 chars in length

- Companies house number - optional. Validation is 8 chars. Should be either 8 numbers or 2 letters and 6 numbers but couldn't find a specific definition of this so validation is slightly looser. (8 chars)

- Company Name

- Company contact details (email, name, phone)

- Summary - confirm these details

- Submit the company and an API call is made to POST company to API and wipe session

Part 2 - User

- the company name and the new ID are persited in the session to this part of the process.

- Capture user email, make a token with email, supplier id and supplier name and send email to user to claim the account.

- The email link will take user into the "invite user flow" and the user can create an account by adding name and password.


app/main/views/suppliers.py
```diff
@@ -105,3 +107,302 @@ def update_supplier():
...
+    required_fields = [
+        "email_supplier_id",
+        "email_company_name",
+    ]
+
+    missing_fields = [field for field in required_fields if field not in session]
+
+    if missing_fields:
+        current_app.logger.info("Failed to create user as broken session", session)
+        abort(503)
```
robyoung: This feels more like a 400.

app/templates/suppliers/company_contact_details.html
```diff
@@ -0,0 +1,93 @@
...
+
+{% block breadcrumb %}
+{%
+with items = [
+{
+"link": "/",
+"label": "Digital Marketplace"
+}
+]
+%}
```
robyoung: I think the indenting's been lost here.

config.py
```diff
@@ -32,6 +32,7 @@ class Config(object):
     INVITE_EMAIL_FROM = 'enquiries@digitalmarketplace.service.gov.uk'
     INVITE_EMAIL_SUBJECT = 'Your Digital Marketplace invitation'
 
+    CREATE_USER_SUBJECT = 'Create your Digital Marketplace account'
```
robyoung: This seems like an odd thing to configurable. Seems to make more sense as a constant in the module.

app/templates/suppliers/company_contact_details.html
minglis: Fixed

config.py
minglis: All content for these is here. If we want to move these, and maybe we do, I think it;s a later refactor.

robyoung:
:+1:

PR #169, opened 2015-08-12
Complete draft

allait:
### Add /complete draft endpoint
Sends the service id to the API to mark the service as "completed"
and redirects to the draft services page.

Shows "service was completed" message on the services page.

### Split service name and description into separate sections
Since services without names are not shown in the service list
saving the service name should be as simple as possible and
shouldn't require a value for the service summary.

### Hide "return to service" links for drafts without a name
screen shot 2015-08-12 at 17 06 05

Drafts without a name won't appear in the drafts list, so it shouldn't
be possible to input any other data until the draft name is set.

Hiding the links for unnamed drafts means there's no way to get to
the draft sections page before entering the service name.

### Don't allow updates to non-editable sections
Section editable flag affects whether section "Edit" link is visible.
But it was still possible to open the section edit page for non-editable
sections and the changed data was validated and saved as service update.

This aborts any edit attempts to readonly sections with a 404.

### Show complete and draft submission lists and counts
Splits draft services based on status into complete and incomplete
drafts.

### Hide unnamed services from the draft services list
The need for non-stored "New service" name for unnamed services
pops up in different contexts (message notifications, button labels,
copied service names etc.), so in order to avoid these edge cases
we've decided to pretend that unnamed services weren't saved in the
first place.

Service name is the section right after service lot, so we've removed
any links to the draft overview page for unnamed services and split
the description into a separate section to make this step easier.

This way if the user decides not to enter a service name the service
is never displayed in the drafts list and there's no way to get to
any other edit page to enter other service data (apart from manually
entering the URL).

Unnamed drafts can be cleaned up when the framework goes live.

### Reverse draft lists so that newer drafts are always first
Since API returns drafts order by asc(id), reversing the list shows
newly created or copied drafts at the top of the list.

### Add "Move to complete" button and questions count to the draft page
Uses content questions data to count required questions that don't
have a value (this matches the check for "Answer required" in the
section fields).

If all required questions are answered a "Move to complete" button
is displayed.

### Add last edit information to the draft service page …
Shows time and user name of the last update using the auditEvents
data sent by the API endpoint.

screen shot 2015-08-12 at 17 02 47
screen shot 2015-08-12 at 17 03 11
screen shot 2015-08-12 at 17 03 25

### Show unanswered question counts on the drafts list page
screen shot 2015-08-12 at 17 01 17

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -8,7 +8,7 @@ from dmutils.email import send_email, MandrillException
 from ...main import main, declaration_content
 from ..helpers.frameworks import get_error_messages
 from ... import data_api_client
-from ..helpers.services import get_draft_document_url
+from ..helpers.services import get_draft_document_url, get_drafts
 from ..helpers.frameworks import register_interest_in_framework
 
 
@@ -23,12 +23,10 @@ def framework_dashboard():
 
     try:
         register_interest_in_framework(data_api_client, 'g-cloud-7')
-        drafts = data_api_client.find_draft_services(
-            current_user.supplier_id,
-            framework='g-cloud-7'
-        )['services']
     except APIError as e:
         abort(e.status_code)
+
+    drafts, complete_drafts = get_drafts(data_api_client, current_user.supplier_id, 'g-cloud-7')
 
     try:
         declaration_made = bool(data_api_client.get_selection_answers(
@@ -43,7 +41,7 @@ def framework_dashboard():
         "frameworks/dashboard.html",
         counts={
             "draft": len(drafts),
-            "complete": 1,
+            "complete": len(complete_drafts),
         },
         declaration_made=declaration_made,
         **template_data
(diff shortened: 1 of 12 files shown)


quis:
Need to speak to @ralph-hawkins about this, but I think we need a line under the complete button that says something like "You can still edit services once they're complete".

Otherwise we'll probably have a situation where everyone waits until the last minute to complete their services, just in case they want to make changes.

app/main/views/frameworks.py
```diff
@@ -54,18 +50,24 @@ def framework_dashboard():
...
+        draft['priceString'] = format_service_price(draft)
+        content = new_service_content.get_builder().filter(draft)
+        sections = get_service_attributes(draft, content)
+
+        draft['unanswered_questions'] = len([
+            q.label
+            for s in sections
+            for q in s['rows']
+            if q.answer_required
+        ])
```
robyoung: This feels like it should be a function in service helpers. It's used again [here](https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/169/files#diff-32bf67afda65ad26faef3bf3efd522b1R364)

robyoung:
:+1:

PR #170, opened 2015-08-13
Register interest in framework

robyoung:
When a user clicks the 'Apply to become a G-Cloud 7 supplier' link on the supplier dashboard this registers their interest in the framework. This in turn will result in them going on an email list for framework information and changes the dashboard link.

### DIFF
--- a/app/main/helpers/frameworks.py
+++ b/app/main/helpers/frameworks.py
@@ -1,8 +1,30 @@
+from flask_login import current_user
+from dmutils.audit import AuditTypes
 import re
 from operator import add
 from functools import reduce
 import six
 from werkzeug.datastructures import ImmutableOrderedMultiDict
+
+
+def has_registered_interest_in_framework(client, framework_slug):
+    audits = client.find_audit_events(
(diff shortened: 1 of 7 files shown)


app/main/helpers/frameworks.py
```diff
@@ -5,6 +7,26 @@
...
+        object_type='suppliers',
+        object_id=current_user.supplier_id)
+    for audit in audits['auditEvents']:
+        if audit['data'].get('frameworkSlug') == framework_slug:
+            return True
+    return False
+
+
+def register_interest_in_framework(client, framework_slug):
+    client.create_audit_event(
```
allait: Does this need a `try .. except`?

robyoung: It gets handled by the try catch over here https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/170/files#diff-2830492bae5ad218e386be64bc065150R25

tombye:
I've added some copy changes from @superroz to the dashboard page (in 5c2bbc7)

tombye:
Checked the HTML and :+1: by me.

PR #173, opened 2015-08-13
Remove more questions from the declaration

quis:
> ok. i think we've decided that the parent company questions are likely to be
> ditched so can we delete SQ1-1p-i, SQ1-1p-ii, SQ1-1p-iii, SQ1-1p-iv,
> SQ1-1q-i, SQ1-1q-ii, SQ1-1q-iii, SQ1-1q-iv

—Roz Strachan 12:56 PM


minglis:
👍

PR #174, opened 2015-08-13
Correct the deadline on the framework dashboard

tombye:
It should match that on https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/132.


quis:
:+1:

PR #175, opened 2015-08-13
Content changes for dashboard

minglis:
Changes for content as specified by Ralph.


quis:
We should merge this now for the sake of the demo, so :+1: 

--

…but it still feels a bit weird. There is now the concept of "automatically submitted" as well as "eligible", which are both words for the same thing.

image

PR #176, opened 2015-08-13
Add a "you can still edit" hint after "Move to complete" button

allait:
@quis:

    Need to speak to @ralph-hawkins about this, but I think we need a line under the complete button
    that says something like "You can still edit services once they're complete".

    Otherwise we'll probably have a situation where everyone waits until the last minute to complete
    their services, just in case they want to make changes.

screen shot 2015-08-13 at 14 49 24


quis:
:+1:

PR #178, opened 2015-08-14
Ensure edit page can handle the reduced data from create supplier

minglis:
- Supplier creation only provides name, duns, email and phone.
- Current edit page expects a description
- Also expects postcode to be present

This removes those requirements.


allait:
:+1:

PR #179, opened 2015-08-14
Fix empty drafts list layout bug

allait:
Using an empty iterator in the summary-table renders an empty table
instead of empty message text since `{% if items %}` returns True
even for empty iterators.


quis:
:+1:

PR #180, opened 2015-08-14
Fix margin-top on the "Move to complete" button

allait:
Overwrite the new .button-save "margin-top: 45px" for the "Move to
complete" button on the service submission page.

screen shot 2015-08-14 at 12 21 33


quis:
:+1:

PR #182, opened 2015-08-17
Form accessibility improvements

quis:
## Use Javascript to jump to field with error
- Implements (and depends on) alphagov/digitalmarketplace-frontend-toolkit#141

## Use toolkit-based question numbers
- Implements (and depends on) alphagov/digitalmarketplace-frontend-toolkit#140

### DIFF
--- a/app/templates/macros/toolkit_forms.html
+++ b/app/templates/macros/toolkit_forms.html
@@ -1,4 +1,4 @@
-{% macro text(question_content, service_data, errors) -%}
+{% macro text(question_content, service_data, errors, question_number=None) -%}
   {%
     with
     question=question_content.question|markdown,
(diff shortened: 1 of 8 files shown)


app/templates/macros/toolkit_forms.html
```diff
@@ -129,7 +134,8 @@
     hint=(question_content.hint or '')|markdown,
     name=question_content.id,
     value=service_data[question_content.id],
-    error=errors.get(question_content.id)['message']
+    error=errors.get(question_content.id)['message'],
+    question_number=question_number
```
allait: Small tip for the future: macro arguments exist in the same context as other variables, so if the variable name and value that included template expects match the macro argument, there's no need to pass it in `with`, since it's available anyway.

allait:
:+1:

PR #183, opened 2015-08-17
Refactor tests login method

robyoung:
Refactor the login method method that it only mocks the login `data_api_client` for the duration of the login so that other tests on the same views module are not affected. The mocking of get_user method in a more aggressive way stays the same because it is needed for subsequent
requests. However, it is now cleaned up in the teardown.

This change also updates all the tests to use the same login method. Some tests had replaced the core one in an attempt to allow testing of the things it was mocking. This should no longer be necessary.

### DIFF
--- a/tests/app/helpers.py
+++ b/tests/app/helpers.py
@@ -11,6 +11,7 @@ class BaseApplicationTest(object):
     def setup(self):
         self.app = create_app('test')
         self.client = self.app.test_client()
+        self.get_user_patch = None
 
     def teardown(self):
         self.teardown_login()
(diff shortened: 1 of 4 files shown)


tests/app/helpers.py
```diff
@@ -89,19 +92,24 @@ def services():
...
-            return_value=(self.user(
-                123, "email@email.com", 1234, 'Supplier Name', 'Name')))
+        with patch('app.main.views.login.data_api_client') as login_api_client:
+            login_api_client.authenticate_user.return_value = self.user(
+                123, "email@email.com", 1234, 'Supplier Name', 'Name')
 
-        data_api_client.get_user = Mock(
-            return_value=(self.user(
-                123, "email@email.com", 1234, 'Supplier Name', 'Name')))
+            self.original_get_user = data_api_client.get_user
```
allait: Instead of saving the old method we could use `patch` here manually:
```
self.patch = mock.patch(data_api_client, 'get_user', ...)
self.patch.start()
...
self.patch.stop()
```

allait:
:+1:

PR #186, opened 2015-08-18
Additional messaging for supplier sign up

TheDoubleK:
The copy here may need tweaking still ( @ralph-hawkins , @superroz ), but these are essentially the extra messages we need for supplier sign-up.

### DIFF
--- a/app/templates/suppliers/duns_number.html
+++ b/app/templates/suppliers/duns_number.html
@@ -18,6 +18,17 @@
 
 {% block main_content %}
 
+{% if form.duns_number.errors[0] == 'DUNS number already used' %}
+<div class="validation-masthead" role="group" aria-labelledby="validation-masthead-heading">
+    <h1 class="validation-masthead-heading" id="validation-masthead-heading">
+        A supplier account already exists with that DUNS number
+    </h1>
+    <p class="validation-masthead-description">
+        If you no longer have your account details, or if you think this may be an error, please contact <a href="mailto:enquiries@digitalmarketplace.service.gov.uk?subject=DUNS%20number%20question" title="Please contact enquiries@digitalmarketplace.service.gov.uk">enquiries@digitalmarketplace.service.gov.uk</a>
+    </p>
+</div>
+{% endif %}
+
 <div class="single-question-page">
   {%
   with
--- a/tests/app/main/test_suppliers.py
+++ b/tests/app/main/test_suppliers.py
@@ -109,8 +109,8 @@ class TestSuppliersDashboard(BaseApplicationTest):
             res = self.client.get("/suppliers")
             assert_equal(res.status_code, 200)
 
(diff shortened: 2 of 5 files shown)


tests/app/main/test_suppliers.py
```diff
@@ -409,11 +409,13 @@ def test_should_be_an_error_if_duns_number_in_use(self, data_api_client):
...
             "/suppliers/duns-number",
             data={
-                'duns_number': "9999999999"
+                'duns_number': "123456789"
             }
         )
         assert_equal(res.status_code, 400)
-        assert_true("DUNS Number must be 9 digits" in res.get_data(as_text=True))
+        page = res.get_data(as_text=True)
+        assert_true("A supplier account already exists with that DUNS number" in page)
```
allait: It's better to use `assert_in` for these checks, it gives a much better error message showing the actual page text instead of `False is not True` that `assert_true` returns.

app/templates/suppliers/duns_number.html
```diff
@@ -18,6 +18,17 @@
 
 {% block main_content %}
 
+{% if form.duns_number.errors[0] == 'DUNS number already used' %}
+<div class="validation-masthead" role="group" aria-labelledby="validation-masthead-heading">
+    <h1 class="validation-masthead-heading" id="validation-masthead-heading">
+        A supplier account already exists with that DUNS number
```
allait: Can we set this text in the actual form error instead of comparing the message only to rewrite it?

tests/app/main/test_suppliers.py
TheDoubleK: You're right - I just used what was there already but will fix it up now.

robyoung: Ordinarily I'd agree but these kinds of tests are testing whether some text appears in the entire HTML payload. The standard `assert_in` error message spits out the whole page making it really hard to read.

app/templates/suppliers/duns_number.html
TheDoubleK: The short version that gets passed in is shown unedited in-line next to the input box.  One of the places where the message is displayed will need to rewrite it, and I think it's nicer to keep the thing we test as the short one.

tests/app/main/test_suppliers.py
TheDoubleK: If the tests are passing then you don't see any of the mess, and if they fail then maybe you do need to sift through the output to find the error?  Otherwise with failing tests you have to stick a `print` in there to work out what's gone wrong.

So I vote to go with `assert_in`, even though I see your point.

app/templates/suppliers/duns_number.html
TheDoubleK: Also, the copy of the long message at the top is likely to change when Roz gets back. The short message is more likely to be OK as it is I think.

robyoung:
Discussed in real life :+1:

PR #187, opened 2015-08-18
[#101485630] Fix bug in invite audit events

robyoung:
See: https://www.pivotaltracker.com/story/show/101485630

The API client expects a string rather than an Enum object.


allait:
:+1:

PR #188, opened 2015-08-18
Fix bug with features, benefits, client list

quis:
Implements: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/142

--

Ensure list entries keep their 'name' attribute

Any changes made to features and/or benefits are not being persisted on a save.

Replication:
- Navigate to the edit features and benefits page of a service.
- Edit features and/or benefits and click "Save and return to service"

Actual:
- Changes made are not persisted

Expected:
- Changes to either features and/or benefits should be persisted.

Reason:
- `name` parameter of each feature is getting changed by the JS from
`serviceFeatures` to `input-serviceFeatures`

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -5,7 +5,7 @@
     "jquery": "1.11.2",
     "hogan": "3.0.2",
     "jquery-details": "https://github.com/mathiasbynens/jquery-details/archive/v0.1.0.tar.gz",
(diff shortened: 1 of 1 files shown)


allait:
:+1: once the tests pass

PR #189, opened 2015-08-18
[#100214054] Add Communications and Clarifications files to updates page.

pcraig3:
### Still left to do

- [x] Merge [utils PR 137](https://github.com/alphagov/digitalmarketplace-utils/pull/137) and update `requirements.txt` accordingly.
- [x] Depends on https://github.com/alphagov/digitalmarketplace-utils/pull/141
- [x] @tombye had some opinions about the date formats

***

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/100214054)

This update displays (on the framework updates page) a listing of files in an s3 bucket based on an assumed directory structure.  Links for each file are signed URLs which are valid for 30 seconds -- exactly like the ITT .zip file.

Taking [Ralph's prototype](http://dm-prototype.herokuapp.com/updates) as a basis, I created a root-level directory with two subdirectories: one for communications (emails) and one for clarification questions.

```
# Adding an email file 
g-cloud-7-updates/communications/Title of Email.odt

# Adding a clarifications questions file
g-cloud-7-updates/communications/Clarification questions and answers.odt
```
If the directory structure doesn't exist, then both tables display their empty messages.

##### Things to know about files 

- Listing the [Keys](http://boto.readthedocs.org/en/latest/ref/s3.html#module-boto.s3.key) in an S3 bucket gives us each one's `name` (path), `size`, and `last_modified` timestamp, which we process in utils before returning to the app.
- There's a problem [getting back metadata](https://github.com/boto/boto/issues/570) when listing files, so we're not setting/getting any.  
 - It looks like [we can use spaces](https://github.com/boto/boto/issues/570) in filenames without any problems, so it didn't seem like a concern.
- Uploading a file with the same name as an existing file overwrites the existing file.
- Slashes (`/`) in filenames will be converted to colons (`:`) during the upload.

***

### Screenshot

g7-updates

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -142,11 +142,11 @@ def framework_supplier_declaration(section_id):
     ), status_code
 
 
-@main.route('/frameworks/g-cloud-7/download-supplier-pack', methods=['GET', 'POST'])
+@main.route('/frameworks/g-cloud-7/<path:filepath>', methods=['GET'])
 @login_required
 @flask_featureflags.is_active_feature('GCLOUD7_OPEN')
-def download_supplier_pack():
-    url = get_draft_document_url('g-cloud-7-supplier-pack.zip')
+def download_supplier_file(filepath):
+    url = get_draft_document_url(filepath)
     if not url:
         abort(404)
 
@@ -196,7 +196,7 @@ def framework_updates(error_message=None):
 def framework_updates_email_clarification_question():
 
     # Stripped input should not empty
-    clarification_question = escape(request.form.get(CLARIFICATION_QUESTION_NAME, '')).strip()
+    clarification_question = request.form.get(CLARIFICATION_QUESTION_NAME, '').strip()
 
     if not clarification_question:
         return framework_updates("Question cannot be empty")
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -165,21 +165,23 @@ def framework_updates(error_message=None):
     uploader = s3.S3(current_app.config['DM_G7_DRAFT_DOCUMENTS_BUCKET'])
     file_list = uploader.list('g-cloud-7-updates/')
 
-    sections = {
-        'clarifications': {
+    sections = [
+        {
+            'section': 'communications',
             'heading': "G-Cloud 7 communications",
             'empty_message': "No communications have been sent out",
             'files': []
         },
-        'communications': {
+        {
+            'section': 'clarifications',
             'heading': "G-Cloud 7 clarification questions and answers",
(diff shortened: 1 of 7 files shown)


ralph-hawkins:
Can we move the 'Ask question button' a lot closer to the text input box

ralph-hawkins:
This page needs a 'Return to G-Cloud 7 application' link at the bottom. See the Supplier declaration for the pattern.

TheDoubleK:
Clicking the "Clarification questions" gives me a file error - but think thats an issue with the file in S3 (it's not actually a PDF?) rather than the code here?

app/main/views/frameworks.py
```diff
@@ -155,12 +187,12 @@ def framework_updates():
...
     # Stripped input should not empty
-    clarification_question = escape(request.form.get(CLARIFICATION_QUESTION_NAME, '')).strip()
+    clarification_question = request.form.get(CLARIFICATION_QUESTION_NAME, '').strip()
 
     if not clarification_question:
-        return _framework_updates_page("Question cannot be empty")
+        return framework_updates("Question cannot be empty")
     elif len(clarification_question) > 5000:
-        return _framework_updates_page("Question cannot be longer than 5000 characters")
+        return framework_updates("Question cannot be longer than 5000 characters")
```
TheDoubleK: When this error happens the submitted text is not re-populated into the text box.  That would be quite frustrating if I'd just written a genuine 5001+ character question.

TheDoubleK:
Also, I'm not entirely sure how this works for me locally - has it picked up the right boto credentials by chance from my settings, or is the dev bucket open?

app/main/views/frameworks.py
```diff
@@ -143,10 +145,40 @@ def download_supplier_pack():
...
+        {
+            'section': 'communications',
+            'heading': "G-Cloud 7 communications",
+            'empty_message': "No communications have been sent out",
+            'files': []
+        },
+        {
+            'section': 'clarifications',
+            'heading': "G-Cloud 7 clarification questions and answers",
+            'empty_message': "No clarification questions exist",
```
TheDoubleK: Should this be "No clarification *answers* exist"?  Because some questions might have been asked but not yet answered - e.g. if I submitted a clarification question I might be worried if the message still said "no questions exist", and wonder if my message had been sent in.

pcraig3: You're probably right.
In the [prototype](http://dm-prototype.herokuapp.com/updates), the section name is 'Clarification questions' while the files are all named 'Clarification questions and answers'.  
The point of this section is obviously that it contains answers though.

TheDoubleK:
:+1: - nice.

PR #190, opened 2015-08-19
[#101511456] Should not send a "create account" email if an account already exists

TheDoubleK:
Fixes this bug: https://www.pivotaltracker.com/story/show/101511456

"Create your account" shouldn't send an email if an account already exists for the email address supplied.  This adds a check that a user account with the supplied email doesn't already exist.

I also learned a lot about how mocks need to be ordered to make the tests pass OK... ouch.


TheDoubleK:
Going to fix as discussed in real life.

quis:
:+1:

PR #191, opened 2015-08-19
Remove service name links to edit page if editing services flag is False

allait:
Previous feature flag check removed the "View service" link if the
service editing was turned off, but there's still a service name link
to the "edit" page.

Instead, we keep the "View service" links and remove the service name
link if the editing is disabled.

### DIFF
--- a/app/templates/services/list_services.html
+++ b/app/templates/services/list_services.html
@@ -57,43 +57,31 @@
   ) %}
     {% call summary.row() %}
 
-      {{ summary.service_link(
-          item.serviceName,
-          url_for('.edit_service', service_id=item.id)
-      ) }}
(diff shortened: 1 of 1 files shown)


robyoung:
:+1: I was going to suggest keeping the catch all condition for statuses we create in the future. However, I think that we should explicitly handle them as and when we need to.

TheDoubleK:
Something goes wrong with the layout when feature flag is off - the columns get messed up:

With feature flag on - 
screen shot 2015-08-19 at 11 58 40

With feature flag off - too much wrapping - 
screen shot 2015-08-19 at 11 59 03

allait:
@TheDoubleK should be fixed now

PR #192, opened 2015-08-19
Correct URL of return link on declaration form

tombye:
As described by this story: https://www.pivotaltracker.com/story/show/101502250

### DIFF
--- a/app/templates/frameworks/edit_declaration_section.html
+++ b/app/templates/frameworks/edit_declaration_section.html
@@ -69,7 +69,7 @@
       {% include "toolkit/button.html" %}
     {% endwith %}
 
-    <a href="/suppliers">Return to G-Cloud 7 application</a>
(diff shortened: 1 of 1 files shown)


app/templates/frameworks/edit_declaration_section.html
```diff
@@ -69,7 +69,7 @@
       {% include "toolkit/button.html" %}
     {% endwith %}
 
-    <a href="/suppliers">Return to G-Cloud 7 application</a>
+    <a href="/suppliers/frameworks/g-cloud-7">Return to G-Cloud 7 application</a>
```
allait: Could we use `url_for` here?

tombye:
@allait switched the HTML out for a call to `url_for` in 722b17e

TheDoubleK:
:+1:

PR #193, opened 2015-08-19
Move equality question to end of first section

tombye:
As requested by @superroz 

Before:

declaration_question_move_master

After:

declaration_question_move


allait:
:+1:

PR #196, opened 2015-08-19
[#101414610] Add more audit logging

robyoung:
## Fix some tests on update-user

They were pointing at create-user and were not giving any data so the failure was actually coming from the form validation. I couldn't find anything useful to test in the response to distinguish them so I've added an assertion on get_user being called which will at least confirm that the right view is being requested.

## Only register interest in a framework once

See: https://www.pivotaltracker.com/story/show/101415342

## Add various application and audit logs

See: https://www.pivotaltracker.com/story/show/101414610

## Depends on
- [x] https://github.com/alphagov/digitalmarketplace-utils/pull/140

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -4,6 +4,7 @@ from flask import render_template, request, abort, flash, redirect, url_for, esc
 from flask_login import login_required, current_user
 
 from dmutils.apiclient import APIError
+from dmutils.audit import AuditTypes
 from dmutils import flask_featureflags
 from dmutils.email import send_email, MandrillException
 from dmutils.formats import format_service_price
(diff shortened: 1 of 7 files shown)


app/main/views/frameworks.py
```diff
@@ -143,6 +144,8 @@ def download_supplier_pack():
 @login_required
 @flask_featureflags.is_active_feature('GCLOUD7_OPEN')
 def framework_updates():
+    current_app.logger.info("g7updates.viewed: user_id:%s supplier_id:%s",
```
allait: Should we just log user_email/supplier_id as part of the logged request data in the supplier app? That should cover the G7 updates and help debugging any issues with the non-200 status responses, which might be hard to do with just a supplier app URL

allait:
:+1:

PR #197, opened 2015-08-19
Create supplier content changes

tombye:
As requested by @superroz and @cathrooney.

### DIFF
--- a/app/main/forms/suppliers.py
+++ b/app/main/forms/suppliers.py
@@ -84,6 +84,6 @@ class CompanyContactDetailsForm(Form):
 
 class EmailAddressForm(Form):
     email_address = StringField('Email address', validators=[
-        DataRequired(message="Email can not be empty"),
-        Email(message="Please enter a valid email address")
+        DataRequired(message="You must provide a email address."),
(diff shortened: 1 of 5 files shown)


app/main/forms/suppliers.py
```diff
@@ -48,42 +48,42 @@ class EditContactInformationForm(Form):
...
+        DataRequired(message="You must enter a DUNS number with 9 digits."),
+        NumberRange(min=100000000, max=999999999, message="You must enter a DUNS number with 9 digits.")
     ])
 
 
 class CompaniesHouseNumberForm(Form):
     companies_house_number = StringField('Companies house number', validators=[
         Optional(),
-        Length(min=8, max=8, message="Companies house number must be 8 characters")
+        Length(min=8, max=8, message="You must enter a Companies House number with 8 characters.")
```
robyoung: This sounds strange given that the field is optional.

tombye: That's a good point. This change wasn't in the doc with the others, I just added it to make that sentence conform to the new style.

@superroz can you advise on how we should explain this?

PR #199, opened 2015-08-20
Extract script to set test environment

robyoung:
This makes it easier to set the environment into your current shell for running specific tests.


allait:
:+1:

PR #204, opened 2015-08-21
Do not connect S3 if no files are being uploaded

robyoung:
I only noticed this because I was trying to manually test the supplier app when not connected to the internet. Instantiating `S3` talks to S3, this isn't necessary if we're not uploading files.


quis:
LGTM :+1:

PR #206, opened 2015-08-21
[#101505222] Split supplier declaration

TheDoubleK:
See: https://www.pivotaltracker.com/story/show/101505222

This splits the supplier declaration over four pages, one for each logical section - the first three have a `Save and continue` button and the final page has a `Submit` button, which returns the user to the dashboard.

The `Edit supplier declaration` link from the dashboard always starts the declaration from page 1, even if it has been started already.  I think this is OK, as previous answers will be pre-filled and we don't really want to get into having a summary page for the declaration.

There are now three possible states for the declaration - unstated, started and complete. Dashboard messaging reflects the current status e.g. "You have started making the supplier declaration, but it is not yet finished"

The story says we should have a summary page, but I think that can be done as a later add-on, as this works as it is.


app/main/views/frameworks.py
```diff
@@ -207,3 +216,22 @@ def _framework_updates_page(error_message=None):
...
+        )['selectionAnswers']['questionAnswers']
+    except APIError as e:
+        if e.status_code == 404:
+            return 'unstarted'
+        else:
+            abort(e.status_code)
+
+    if not answers:
+        return 'unstarted'
+    elif all(keys in answers for keys in ONE_REQUIRED_QUESTION_FROM_EACH_PAGE):
```
robyoung: /nitpick I think `keys` should be singular here. It confused me for a minute.

ralph-hawkins:
Can the final button say ‘Make declaration’. And have the hint text ‘You can edit the declaration until the application deadline’.

And make it look like the pattern we have on the ‘Move to complete’ button on the service page.

screen shot 2015-08-21 at 16 43 09

app/main/helpers/frameworks.py
```diff
@@ -28,8 +28,8 @@ def register_interest_in_framework(client, framework_slug):
             data={'frameworkSlug': framework_slug})
 
 
-def get_required_fields(all_fields, answers):
-    required_fields = set(all_fields)
+def get_required_fields(page_fields, answers):
```
allait: I think it'd be much simpler to keep this as is and apply the `& set(page_fields)` to the resulting list of required questions with a separate function. That way we can test them separately, there's no need for the `in page_fields` checks and there's no risk of missing any of the checks.

app/main/helpers/frameworks.py
```diff
@@ -39,19 +39,17 @@ def get_required_fields(all_fields, answers):
...
         required_fields.add('SQ1-1cii')
-
     #  If you answered yes to question 27 (non-UK business registered in EU)
-    if answers.get('SQ1-1i-i', False):
+    if 'SQ1-1i-i' in page_fields and answers.get('SQ1-1i-i', False):
         required_fields.add('SQ1-1i-ii')
-
     #  If you answered 'licensed' or 'a member of a relevant organisation' in question 29
-    if len(answers.get('SQ1-1j-i', [])) > 0:
+    if 'SQ1-1j-i' in page_fields and len(answers.get('SQ1-1j-i', [])) > 0:
```
allait: I'm not sure why do we need the `in page_fields` check for the special questions. We check that 'SQ1-1j-i' is on the page, but then just assume that 'SQ1-1j-ii' is on the same page too, which doesn't have to be true.

allait:
> we don't really want to get into having a summary page for the declaration.

My impression from the last conversation about this was that we were going to do the summary page like we have for services, since it's content-driven anyway. Was there an issue with this?

app/main/helpers/frameworks.py
TheDoubleK: Without this check an error is generated for that field on every page of the split-up declaration, even though the question wasn't on the page.

allait: @TheDoubleK I think that's caused by the `not answers.get('SQ5-2a')` further in the function, which should check that the key exists first.
Anyway, I think splitting the `page_questions` from this function is the way to go.

TheDoubleK:
No issue with the summary page, but it doesn't feel necessary to me.

app/main/views/frameworks.py
allait: I think we should validate the full declaration to determine the status, not check a subset of questions that we assume cover all the pages. And we should store the status once the declaration is complete since "declaration is complete" is used by the API MI stats.

app/main/views/frameworks.py
```diff
@@ -92,38 +85,54 @@ def framework_supplier_declaration():
     content = declaration_content.get_builder()
     status_code = 200
 
+    try:
+        page = int(request.args.get('page', 1))
```
allait: We used section ids for URLs in service edit, any reason why we can't do that here? We have the find section/next section logic for it and the URLs would make more sense than the pagination, since we're not really dealing with a list of items here.

app/templates/frameworks/edit_declaration_section.html
```diff
@@ -26,6 +26,20 @@
 
 {% block main_content %}
 
+{% with messages = get_flashed_messages() %}
+  {% if messages %}
+    {% for message in messages %}
+      {% if message == 'supplier_declaration_saved' %}
+      <div class="banner-success-without-action">
+        <p class="banner-message">
+          Updates to your supplier declaration so far have been saved
```
quis: I don't think we need to give feedback here. It gives a false sense of completeness and isn't something we do for services.

TheDoubleK:
Think everything is fixed up now.

* New final button looks like this:
screen shot 2015-08-24 at 09 44 58

* Relevant page field errors are filtered after validating against a complete declaration. 

* Flash messages are gone.

* New URLs for declaration (instead of `page=`) are:
/suppliers/frameworks/g-cloud-7/declaration/g_cloud_7_essentials
/suppliers/frameworks/g-cloud-7/declaration/about_you
/suppliers/frameworks/g-cloud-7/declaration/grounds_for_mandatory_exclusion
/suppliers/frameworks/g-cloud-7/declaration/grounds_for_discretionary_exclusion

* `status` field (started/completed) is saved in the JSON.

app/main/views/frameworks.py
```diff
@@ -207,3 +219,20 @@ def _framework_updates_page(error_message=None):
         error_message=error_message,
         **template_data
     ), status_code
+
+
+def _get_declaration_status():
```
allait: Could we remove the `_` and move this to helpers?

TheDoubleK: I put the method there to start with but I didn't like the feel of making api calls from helpers so moved it back here... I know it's a bit awkward here, but at least it has the api client, etc. to hand. 
What do you think?  It should move anyway?

allait:
:+1:

app/main/views/frameworks.py
allait: I think it's a good idea to limit `views.py` to view functions, especially once the file grows large. It makes it a bit easier to audit routes/decorators etc.

API calls from helpers shouldn't be a problem as long as you pass in the `apiclient` as an argument: that way we don't need to mock any additional references and it's a bit more explicit. I think we've done this in a few other places.

PR #209, opened 2015-08-21
[#101689570] Script to generate CSV files of questions for each lot

TheDoubleK:
See: https://www.pivotaltracker.com/story/show/101689570

Here are the G6 CSVs for comparison (we need somewhere new to put the G7s I think):
https://github.com/alphagov/supplier-submission-portal/tree/master/conf


allait:
Could we rename it to something like `generate_questions_csv.py`?

robyoung:
:+1:

PR #210, opened 2015-08-24
Configure JSON apache logs

robyoung:
Configure Apache to output JSON formatted logs and CloudWatch logs to read them. With the JSON formatted logs more sophistocated filter patterns can be used. This change also introduces the more focused 4xx filter and alarms currently used in the data API (400 - 403).


allait:
:+1:

PR #211, opened 2015-08-25
Edit copy service name

allait:
#### Redirect to service name edit page after making a copy
After discussing this a bit, we've decided to change the way "Make
a copy" works. It still creates a new draft as before, but instead
of returning to the services dashboard it will redirect to the
editing of service name section of the newly created draft.

The assumption is that this will encourage suppliers to give
unique and meaningful names to the copied drafts so that we
don't end up with multiple copies of the same service with the
same name.

#### Fix "Return to summary" link disappearing after validation errors
[#101571000](https://www.pivotaltracker.com/story/show/101571000)

"Return to summary" was hidden for services without a service name
in 09acc4f. The page rendered after validation only uses the update
data, not the full draft service data, so serviceName should be
explicitly added to the context. This was achieved by adding
serviceName to the `posted_data`, but `posted_data` is no longer
passed to the template as it was replaced by `update_data`.

This adds `serviceName` to update_data for both API and document
validation failures.

#### Redirect to summary after editing single section/make a copy
After submitting a section update instead of redirecting to the
next section the user is returned to the summary page if:

* Section edit was started by clicking the "Edit" link on the
  service summary page
* Section edit was started after clicking the "Make a copy" button.
  After editing the service name the user is returned to the new
  service summary page

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -274,7 +274,10 @@ def copy_draft_service(service_id):
     except APIError as e:
         abort(e.status_code)
 
-    return redirect(url_for(".edit_service_submission", service_id=draft_copy['id'], section_id='service_name'))
+    return redirect(url_for(".edit_service_submission",
+                            service_id=draft_copy['id'],
+                            section_id='service_name',
+                            return_to_summary=1))
 
 
 @main.route('/submission/services/<string:service_id>/complete', methods=['POST'])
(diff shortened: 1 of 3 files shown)


TheDoubleK:
:+1: 
But I still think it might be nice to have a "continue" link that take you back into the flow.  The "Answer required" link is not at all obvious.

PR #213, opened 2015-08-25
Remove extra template variable

tombye:
The list macro had two 'hint' variables declared
so was using the second, which wasn't being
filtered for markdown.


quis:
:+1:

PR #215, opened 2015-08-25
Patch data_api_client more better

pcraig3:
Importing the `data_api_client` and patching an individual method (in `test_application.py`) wasn't a great idea, so I've removed that.
I've also removed `Mock`s in a few places so as to be more consistent across the tests.

The supplier app has changed since [story on Pivotal](https://www.pivotaltracker.com/story/show/97111136) related to this was written.  The tests weren't failing when I started this, but the solution wasn't a very sustainable one.

### DIFF
--- a/tests/app/helpers.py
+++ b/tests/app/helpers.py
@@ -1,5 +1,5 @@
 import re
-from mock import patch, Mock
+from mock import patch
 from app import create_app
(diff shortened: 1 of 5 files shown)


allait:
looks more better to me:+1:

PR #216, opened 2015-08-26
Change upload document display

robyoung:
https://www.pivotaltracker.com/story/show/101692502

Change how the previously uploaded file is displayed on the upload form so that rather than showing a long URL we show a message saying when the file was uploaded.

This is implemented in the Jinja2 macro to avoid changing the draft data in place (and thus lying about the value is) and to avoid doing the formatting in the frontend toolkit (which doesn't have access to the
required template filters).

## Before
screen shot 2015-08-26 at 08 40 09

## After
screen shot 2015-08-26 at 08 41 29

## Refactor service helpers
Move `unformat_section_data` to `dmutils.content_loader`:
This continues the work of moving content section related functions into `dmutils`. This removes the need for any question type checking functions in the frontend app any more.

Pass `service_content` into `get_section_error_messages`:
Along with the previous change this removes the dependency on `new_service_content`.

Move `parse_document_upload_time` into helpers:
It is a service helper of a sort.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -1,6 +1,6 @@
 import re
 
-from datetime import timedelta
+from datetime import timedelta, datetime
 from flask import Flask, request, redirect, session, Markup
 from flask_login import LoginManager
 from flask_wtf.csrf import CsrfProtect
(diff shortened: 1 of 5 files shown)


allait:
:+1:

ralph-hawkins:
:+1:

PR #218, opened 2015-08-26
BUGFIX: Need to have the JSONLogFile set for logs

robyoung:
CloudWatch logs configuration needs to have the JSONLogFile variable set
as it's being used later in the file.


allait:
:+1:

PR #220, opened 2015-08-26
[#101903480] Enable G7 on Staging

TheDoubleK:
See: https://www.pivotaltracker.com/story/show/101903480

The Frameworks table is already updated to have G7 `open`, so this should be all that's required to turn on G7 features in staging.

### DIFF
--- a/config.py
+++ b/config.py
@@ -122,7 +122,7 @@ class Production(Live):
 
 class Staging(Production):
     FEATURE_FLAGS_USER_DASHBOARD = enabled_since('2015-08-19')
-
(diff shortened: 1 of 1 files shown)


pcraig3:
Not part of this story, but we should probably remove some of the flags we're done with.

:+1:

TheDoubleK:
Yep I thought the same thing.  Feels like a job for after G7 is launched - at that point I think we can remove everything except the supplier edits to live services. Although actually we'll want to turn G7 off again at the end of the submissions window.  There are definitely some things we can take the flags off altogether though.

PR #221, opened 2015-08-26
Fix file upload form display

robyoung:
It fails if there is no existing file. I did not test this well enough, I am sorry.

### DIFF
--- a/requirements_for_test.txt
+++ b/requirements_for_test.txt
@@ -4,4 +4,5 @@ pep8==1.5.7
 requests-mock==0.6.0
 mock==1.0.1
 lxml==3.4.4
+cssselect==0.9.1
 freezegun==0.3.4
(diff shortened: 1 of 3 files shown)


requirements_for_test.txt
```diff
@@ -4,4 +4,5 @@ pep8==1.5.7
 requests-mock==0.6.0
 mock==1.0.1
 lxml==3.4.4
+cssselect==0.9.1
```
quis: This is so much nicer than XPath.

TheDoubleK:
https://www.youtube.com/watch?v=ttPwA8hitoY

:+1:

requirements_for_test.txt
robyoung: ... for certain types of problem

PR #222, opened 2015-08-26
[#101512016] Clarification question response emails

TheDoubleK:
See: https://www.pivotaltracker.com/story/show/101512016

This sends an email to the user who submitted a clarification question, to confirm that we have received it.  The email looks like this:
screen shot 2015-08-26 at 13 40 33

I also tested the do-not-reply email address, which seems to do the job:
screen shot 2015-08-26 at 15 51 32

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -210,13 +210,14 @@ def framework_updates_email_clarification_question():
             default_textbox_value=clarification_question
         )
 
+    # Submit email to Zendesk so the question can be answered
+    # Fail if this email does not send
     email_body = render_template(
         "emails/clarification_question.html",
         supplier_name=current_user.supplier_name,
         user_name=current_user.name,
         message=clarification_question
     )
-
     try:
         send_email(
             current_app.config['DM_CLARIFICATION_QUESTION_EMAIL'],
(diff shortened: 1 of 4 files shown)


app/main/views/frameworks.py
```diff
@@ -233,6 +234,29 @@ def framework_updates_email_clarification_question():
...
+    )
+    try:
+        send_email(
+            current_user.email_address,
+            email_body,
+            current_app.config['DM_MANDRILL_API_KEY'],
+            current_app.config['CLARIFICATION_EMAIL_SUBJECT'],
+            current_app.config['CLARIFICATION_EMAIL_FROM'],
+            current_app.config['CLARIFICATION_EMAIL_NAME'],
+            ["clarification-question-confirm"]
```
pcraig3: Are these worth storing in config variables?  The last one I put in a config variable because the email was to be a secret.  We're not planning on overriding this one, are we?

tests/app/main/test_frameworks.py
```diff
@@ -465,9 +471,16 @@ def _assert_email(self, send_email, is_called=True):
...
+        if succeeds:
+            send_email.assert_any_call(
+                "email@email.com",
+                mock.ANY,
+                "MANDRILL",
+                "Thanks for your clarification question",
+                "do-not-reply@digitalmarketplace.service.gov.uk",
+                "Digital Marketplace Admin",
+                ["clarification-question-confirm"]
+            )
```
pcraig3: Gotta love those three-state binaries :smile:

TheDoubleK: I know... it seemed better than unpicking the whole lot into several methods though.

app/main/views/frameworks.py
TheDoubleK: I did start with them hard-coded in the view but then changed my mind.

I think it's better than having them buried in the code, and the reset password ones are all in config already so it means everything like this in the same place.  It *might* also be good if we ever want to change things per environment (...no, me neither).

tests/app/main/test_frameworks.py
pcraig3: Yeah, it's nice the logic still (mostly) holds together without your having to do anything too crazy.

config.py
```diff
@@ -32,6 +32,10 @@ class Config(object):
     INVITE_EMAIL_FROM = 'enquiries@digitalmarketplace.service.gov.uk'
     INVITE_EMAIL_SUBJECT = 'Your Digital Marketplace invitation'
 
+    CLARIFICATION_EMAIL_NAME = 'Digital Marketplace Admin'
+    CLARIFICATION_EMAIL_FROM = 'do-not-reply@digitalmarketplace.service.gov.uk'
```
allait: Why is this different from invite and password reset emails?

TheDoubleK: Cath wanted a `do-not-reply` email for the clarification questions. We aren't allowed to communicate with suppliers about their G7 questions directly, so makes sense they can't reply to our confirmation email.

But if they have problems with their account access (invites/passwords) then the enquiries email address is the one they should use to get Nicole to fix things up for them.

pcraig3:
:+1:

config.py
minglis: Does the do-not-reply email account exist?

PR #223, opened 2015-08-26
[#102066014] More required declaration answers

TheDoubleK:
See: https://www.pivotaltracker.com/story/show/102066014

Following discussions with CCS these fields are now required in the declaration: SQ1-1b, SQ1-1d-i, SQ1-1d-ii, SQ1-1e

SQ1-1b wasn't optional before anyway, but the other three were and this makes them required.

Also fixes a small validation bug - previously using `if not ...` on the `SQ5-2a` field returned true for either `None` or `False`, but the required fields should only be added to the errors for the `False` case, and not for the `None` case.  So using `is False` is the right test.

### DIFF
--- a/app/main/helpers/frameworks.py
+++ b/app/main/helpers/frameworks.py
@@ -73,7 +73,7 @@ def get_required_fields(all_fields, answers):
         required_fields.add('SQ3-1k')
 
     # If you answered No to question 26 (established in the UK)
-    if answers.get('SQ5-2a') is False:
+    if 'SQ5-2a' in answers and not answers['SQ5-2a']:
         required_fields.add('SQ1-1i-i')
         required_fields.add('SQ1-1j-i')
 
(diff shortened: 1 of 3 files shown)


app/main/helpers/frameworks.py
```diff
@@ -65,7 +64,7 @@ def get_required_fields(all_fields, answers):
         required_fields.add('SQ3-1k')
 
     # If you answered No to question 26 (established in the UK)
-    if not answers.get('SQ5-2a'):
+    if answers.get('SQ5-2a') is False:
```
allait: What about `if 'SQ5-2a' in answers and not answers['SQ5-2a']`. It's more verbose, but I think it describes what we're checking better and it avoids the identity check

TheDoubleK: Yep can do - what is the problem with the `is False` check? The only values that come in are True or False.

allait: Identity checks tend to behave differently in different interpreters (e.g. PyPy) based on how frames/object interning works. `is None` is ok since `None` is strictly defined as a singleton, but everything else is a bit fuzzy. I think `is False` could be fine, but I'd still rather avoid this when possible.

minglis:
👍 ship it

TheDoubleK:
Rebased and just waiting for Travis...

PR #203, opened 2015-08-21
Change messages, add warnings

quis:
This pull request encompases a bunch of changes:
- from @superroz and @cathrooney's Google Doc
- from when @ralph-hawkins, @katieminatie and I took some creative license yesterday and researched the dashboard page
- from when @ralph-hawkins and I did some guerilla research at the DS£MED

It mostly involves:
- wording of how things are labelled
- more, contextual warning messages if your application to G-Cloud 7 wont  succeed
- ~~the renaming of 'copy' to 'clone'~~
- duplicating the 'mark complete' button at the bottom of the page (because you'll probably click it after reviewing the whole service)

Stepping through a submission:
--
screen shot 2015-08-21 at 07 45 17

screen shot 2015-08-21 at 07 45 34

screen shot 2015-08-21 at 07 45 55

screen shot 2015-08-21 at 07 47 56

screen shot 2015-08-21 at 07 52 29

screen shot 2015-08-21 at 07 46 14

screen shot 2015-08-21 at 07 46 34

PR #224, opened 2015-08-26
[#101483360] Display last modified dates on the G-Cloud 7 dashboard page

pcraig3:
This pull request adds a bit of contextual information to the G-Cloud 7 dashboard about when the supplier pack and the files on the updates page were last modified.

#### Date formats?

I prefer the dates the way they're shown in [Ralph's prototype's dates](http://dm-prototype.herokuapp.com/new-framework) (ie, `26 August` <sub>(I've included an image at the bottom)</sub>), but we don't have this format in our list of formats.  Is it worth adding in, or are we otherwise happy using our `Wednesday 26 August 2015` format? 

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/101483360)

***
### Before

screen shot 2015-08-26 at 15 48 54

### Now

screen shot 2015-08-26 at 15 47 33

### Ralph's Prototype

screen shot 2015-08-26 at 15 57 25

### DIFF
--- a/app/main/helpers/frameworks.py
+++ b/app/main/helpers/frameworks.py
@@ -30,6 +30,19 @@ def register_interest_in_framework(client, framework_slug):
             data={'frameworkSlug': framework_slug})
 
 
+def get_last_modified_from_first_matching_file(key_list, path_starts_with):
+    """
+    Takes a list of file keys and a string.
+    Returns the 'last_modified' timestamp for first file whose path starts with the passed-in string,
+    or None if no matching file is found.
+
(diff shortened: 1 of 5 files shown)


quis:
Making some fairly significant changes to this page in #203. Would be good to get that merged first (waiting on @cathrooney) and then take another look at how these dates can be displayed.

pcraig3:
@ralph-hawkins wants the `26 August` dates instead of the `Wednesday 26 August 2015` dates.

pcraig3:
@quis Right, I'll wait on you then

PR #225, opened 2015-08-26
[#101660010] Extract closing date to a config variable

TheDoubleK:
See: https://www.pivotaltracker.com/story/show/101660010

This will ensure that the date displayed is accurate and consistent in all places it is displayed.

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -46,6 +46,7 @@ def framework_dashboard():
             "complete": len(complete_drafts),
         },
         declaration_status=declaration_status,
+        deadline=current_app.config['G7_CLOSING_DATE'],
         **template_data
     ), 200
 
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -1,5 +1,6 @@
 from flask_login import login_required, current_user
-from flask import render_template, request, redirect, url_for, abort, flash
+from flask import render_template, request, redirect, url_for, abort, flash, \
+    current_app
 
 from ...main import main, existing_service_content, new_service_content
 from ..helpers.services import (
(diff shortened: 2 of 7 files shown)


pcraig3:
We don't care that we're losing the `<abbr>` element, do we? 
I don't really.

TheDoubleK:
I consider it collateral damage.

pcraig3:
Just gave Chris' pull request a thumbs-up, so you might need to refactor this soon.
But it *should* be relatively straightforward.

pcraig3:
Actually, what do you think of doing something like incorporating the `<time>` element? 

I did it [here](https://github.com/alphagov/digitalmarketplace-supplier-frontend/commit/d47a72125c519ea241a51cbbc5c79e5b54e1be7b), but I don't think anyone else has ever done it.  I think, basically, it would call for a timestamp string and a format, both of which could be config variables.

allait:
Can we keep the `&nbsp;`? 3pm and BST get a line break between them on the service overview page at common resolutions

TheDoubleK:
That's why I added the <br> in front of the date - it forces the 3pm BST on to a new line, just as the &nbsp; did before.

TheDoubleK:
Although we could put the `&nbsp;` into the time string itself - wouldn't hurt to have it everywhere the date appears rather than just on the overview page.

TheDoubleK:
@pcraig3 - the time element is a nice idea, but feels like overkill maybe - this date isn't going to change once we go live, so why not just have the string we want to show... no need to recompute and format it every time.

TheDoubleK:
Don't know what happened there... anyway, `&nbsp;` re-introduced and the whole lot rebased so it merges now.

quis:
:+1:

PR #226, opened 2015-08-27
Error page content updated as per Roz suggestions.

minglis:
Lots of people complaining that their link is "invalid" when probably it's expired.

Updates to the wording to help users understand what is going on.


superroz:
👍 :+1:

PR #227, opened 2015-08-27
update content on create page invalid error to match update

minglis:
Missed the create page on last pull request on this.


quis:
:+1:

PR #228, opened 2015-08-27
Add a second submit button when editing drafts

quis:
image

--

This PR give users the explicit choice of whether to:
- save and continue to the next page of questions
- or save and return to the 'service overview' page

It does this by naming each button, and redirecting based on which button is clicked. To do this required bringing in https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/146, which is why some of the tests/CSS have needed to change.


ralph-hawkins:
:+1: :+1: :+1:

PR #229, opened 2015-08-27
Remove ‘Account’ from above page title

ralph-hawkins:
Unnecessary, so shouldn't be there.

Removing highlighted:

screen shot 2015-08-26 at 22 30 27 copy

### DIFF
--- a/app/templates/auth/login.html
+++ b/app/templates/auth/login.html
@@ -52,7 +52,6 @@
 
 {% with
   heading = "Log in to the Digital Marketplace",
-  context = "Account"
 %}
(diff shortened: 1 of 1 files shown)


app/templates/auth/login.html
```diff
@@ -52,7 +52,6 @@ <h3 class="validation-masthead-heading" id="validation-masthead-heading">
 
 {% with
   heading = "Log in to the Digital Marketplace",
```
quis: You need to remove the trailing comma here.

quis:
:+1:

PR #231, opened 2015-08-27
Add trailing commas and Yes-No answers

TheDoubleK:
* The generated CSV files were missing answers for `boolean` type questions - now they are included.
* Lines with fewer options than the maximum are now right-padded with commas so that each line has the same number of columns.  This makes it look pretty in Github, and may help with importing into spreadsheets.


robyoung:
:+1:

PR #232, opened 2015-08-27
Save if changed

allait:
#### Revert to "Save and continue" after clicking "Edit" in service summary
[#102058034](https://www.pivotaltracker.com/story/show/102058034)

We're keeping "Save and return to summary" only for service name edits
after "Make a copy".

#### Only send draft update request to API if any fields were changed
There's an expectation that some suppliers will use the section
editing flows to check filled in questions. When every update
sends a request to the API there's a risk that this would lead
to lost writes if other users are editing the same service at
the same time.

To avoid this we check if any of the submitted fields are different
from existing service data and only send the API update request if
fields were changed.

This works for input fields, but not for empty checkboxes, radios and
file uploads. Since they are not part of the posted data when no value
is selected, they don't trigger the "field change" event, which means
that the request is not sent to the API and there's no "answer required"
error shown.

To keep the "answer required" checks performed by the API we look for
the section questions in the service data and send the posted data to
the API if any of the keys are missing, even if there were no changes
made.

#### Update dmutils to 6.5.1
Allows removing previous priceMax value

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -7,7 +7,8 @@ from ..helpers.services import (
     get_section_error_messages,
     is_service_modifiable, is_service_associated_with_supplier,
     upload_draft_documents, get_service_attributes,
-    get_draft_document_url, count_unanswered_questions
+    get_draft_document_url, count_unanswered_questions,
+    has_changes_to_save
 )
 from ..helpers.frameworks import get_declaration_status
 from ... import data_api_client, flask_featureflags
@@ -443,6 +444,7 @@ def update_section_submission(service_id, section_id):
     else:
         update_data.update(uploaded_documents)
 
+    if not errors and has_changes_to_save(section, draft, update_data):
         try:
             data_api_client.update_draft_service(
                 service_id,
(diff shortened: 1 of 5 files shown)


app/main/views/services.py
```diff
@@ -432,6 +433,7 @@ def update_section_submission(service_id, section_id):
     else:
         update_data.update(uploaded_documents)
 
+    if not errors and has_changes_to_save(section, draft, update_data):
```
robyoung: Is this a sneaky way of reducing the diff or is there another reason I'm missing?

allait: I didn't want to nest the API request even more and I like the way documents_errors check has two clear outcomes now.

TheDoubleK: This is fine as long as we have document uploads on their own page (as we do), but if there were pages with a mix of questions and document uploads (as we had last time) then any validation errors on the questions wouldn't make it to the page, only the document errors.

This probably isn't a problem, because we'll probably stick with document uploads on their own page.

TheDoubleK:
Looks good to go :+1: - but we should bear the thing in mind if we go back to mixed pages in the future.

app/main/views/services.py
allait: @TheDoubleK it was always like that. Getting validation errors from non-upload fields requires sending the request to the API, which might be valid and get saved while the documents aren't, leaving the page partially completed. Not showing all of the form errors seems to be a better option in comparison.

PR #235, opened 2015-08-28
Fix bug whereby token fields not passed to template

minglis:
When the validation fails I was passing None to the template instead of the values from the token we needed to render the page correctly


allait:
:+1:

PR #238, opened 2015-08-28
[#102242820] Validation fixes again

TheDoubleK:
We lost some in a bad merge yesterday, and Cath had a couple more to add for the checkboxes.

See: https://www.pivotaltracker.com/story/show/102242820


quis:
:+1:

PR #239, opened 2015-08-28
[#102242764] Update email copy - do not reply

TheDoubleK:
See: https://www.pivotaltracker.com/story/show/102242764


pcraig3:
We don't want to give any contact email at all?

PR #241, opened 2015-08-28
Redirect to overview after new service name

allait:
#### Fix "save and return to overview" button missing after form validation
next_section_name should be set in both GET and POST views, otherwise
only "Save and continue" button is shown on the page with validation
errors.

#### Use common edit section template for new service lot page
Reuses the existing template by manually creating section and errors
structures. Since page title is different from the section name on the
overview page we use a dictionary instead of the actual section object.

This updates the breadcrumbs to include "Services" item, replaces Save
button with submit input and changes the page heading size to match
other section pages.

#### Show section errors below the page heading
After talking with Chris we've decided that errors block fits better
below the page heading.

#### Redirect to overview page after entering new service name
Ralph suggested we redirect suppliers to service overview page after
they've entered a new service name.


pcraig3:
Seems cleaner and works the same.

:+1:

PR #242, opened 2015-08-28
Use DM-Request-ID header from the response

robyoung:
Use the DM-Request-ID header from the response rather than the request.
As this app is what generates the request ID (Nginx strips any that may
have been nefariously set from outside) there will be no DM-Request-ID
request header here.

http://httpd.apache.org/docs/2.2/mod/mod_log_config.html#formats


allait:
:+1:

PR #243, opened 2015-08-28
Add boto configuration documentation to README file

pcraig3:
Explains how to locally configure boto and links to official documentation.

I asked @allait about it, and both @tombye and @TheDoubleK asked me about it, so there's probably a reason to stick it in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,29 @@ The supplier frontend runs on port 5003. Use the app at [http://127.0.0.1:5003/]
 To use feature flags, check out the documentation in (the README of)
 [digitalmarketplace-utils](https://github.com/alphagov/digitalmarketplace-utils#using-featureflags).
 
+### Configuring boto
+
+[boto](https://github.com/boto/boto) provides a Python interface to Amazon Web Services; it's what we're using to download from and upload to our s3 buckets.
+
+If you don't [configure your AWS credentials correctly](http://boto.readthedocs.org/en/latest/boto_config_tut.html?highlight=~/.aws/credentials#credentials)
+on your local machine, you'll probably run into a nasty-looking `boto.exception.NoAuthHandlerFound` page at some point.
+
+The short version is that you should create an `~/.aws/credentials` file formatted like so:
+```bash
+[default]
+aws_access_key_id = ...
+aws_secret_access_key = ...
+```
+
+AWS access keys can be found/configured in the Identity and Access Management (IAM) section of the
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -112,6 +112,20 @@ The supplier frontend runs on port 5003. Use the app at [http://127.0.0.1:5003/]
 To use feature flags, check out the documentation in (the README of)
 [digitalmarketplace-utils](https://github.com/alphagov/digitalmarketplace-utils#using-featureflags).
 
+### Configuring boto
+
+[boto](https://github.com/boto/boto) provides a Python interface to Amazon Web Services; it's what we're using to download from and upload to our s3 buckets.
+
+If don't [configure your AWS credentials correctly](http://boto.readthedocs.org/en/latest/boto_config_tut.html?highlight=~/.aws/credentials#credentials)
```
TheDoubleK: If *you* don't?

README.md
```diff
@@ -112,6 +112,20 @@ The supplier frontend runs on port 5003. Use the app at [http://127.0.0.1:5003/]
...
+on your local machine, you'll probably run into a nasty-looking `boto.exception.NoAuthHandlerFound` page at some point.
+
+The short version is that you should create an `~/.aws/credentials` file formatted like so:
+```bash
+[default]
+aws_access_key_id = ...
```
TheDoubleK: Which keys should we use? Maybe specify to have the dev account keys rather than prod?  Or is that obvious?

pcraig3: Yeah, I kind of thought it was obvious, but I suppose all's fair in a README.

allait:
We've run into an issue with @tombye today were the credentials file wasn't used because some AWS-related environment variables were set with a key from another account. Might be worth adding a note about checking `env` for any boto variables (e.g. AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_SECURITY_TOKEN and AWS_PROFILE) and removing them with `unset`

minglis:
Might be worth talking about this in the manual also. Do we have a "set up you're DM environemnt" page in the manual? How to do each app, it's dependants (files/data/S3/creds/etc) and so on?

Each app should have something, manual could link to the READMEs. But there should be a place where we talk about the environment broadly and start talking about the AWS depenencies.

pcraig3:
@minglis I'll put something into the design manual soon, but is it worth merging this sooner rather than later?

allait:
I think this is useful as is.
The manual right now refers to each individual app README for setup instructions, I agree it's worth talking about how we organise this in the future, but it shouldn't block this PR, so :+1:

PR #246, opened 2015-09-01
Content changes to have better:

minglis:
1) Messaging when no clarification questions exist
2) Actually say the closing date rather than "x"


app/templates/frameworks/updates.html
```diff
@@ -111,7 +111,7 @@
         {% include "toolkit/forms/textbox.html" %}
       {% endwith %}
       <p>
-        The deadline for clarification questions is x
+        The deadline for clarification questions is 5pm BST, 22 September 2015. All responses will be published by 5pm BST, 29 September 2015.
```
tombye: Can the 'BST' text be wrapped in an `abbr` tag, like we do on the buyers' app?

https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/f19b5a784b136a2a5036e7357de833da3e2afe1a/app/templates/index.html#L49

pcraig3:
We're not using `<abbr>` tags for our [G7 closing deadline](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/config.py#L21) variable, just FYI.

pcraig3:
Also, we changed the wording from "clarification questions" to "clarification answers" [in this pull request](https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/189#discussion_r37846134).  Are we sure we want to go back to "questions"?

tombye:
@pcraig3 I've put up a fix for the missing `<abbr>` you mention: https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/247

tombye:
@pcraig3 I'm not sure the distinction (between clarification questions and answers) exists. The validation messaging there is saying no questions have been answered whereas the section is called 'validation questions and answers', I assume, because the answers will be presented alongside the question they relate to.

The text this relates to in this pull request is:

1. "No clarification questions have been posted yet"
2. "The deadline for clarification questions..."

If 1 is right in saying the supplier hasn't submitted any questions then that still makes sense. Likewise, the deadline, again I assume, is for questions to be submitted, not answered so isn't all of this correct?

robyoung:
Regarding the abbr for BST. We're not adding the time zone in any of our generated date strings. Is that a problem?

minglis:
UPdated to be questions and answers. The doc will contain both, including unanswered questions. Note also this text will exist only until there is a document at which point it goes aways.

pcraig3:
@tombye But the first message doesn't mean that the Supplier hasn't submitted any questions, it means that no documents containing answers to submitted questions have been published (by us).  

A supplier who's sent in one or more questions might misinterpret the message `No clarification questions have been posted yet` as saying that their questions haven't been received.

tombye:
@pcraig3 In that case, yeah, the message should say what has happened, not an interpretation of it.

PR #247, opened 2015-09-01
Explain 'BST' where used in the G7 deadline

tombye:
As pointed out by @pcraig3 in https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/246#issuecomment-136648789

### DIFF
--- a/config.py
+++ b/config.py
@@ -18,7 +18,7 @@ class Config(object):
     DM_DATA_API_AUTH_TOKEN = None
     DM_MANDRILL_API_KEY = None
     DM_CLARIFICATION_QUESTION_EMAIL = 'digitalmarketplace@mailinator.com'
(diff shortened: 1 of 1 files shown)


minglis:
👍

TheDoubleK:
The `<abbr>` that got removed previously was "British Standard Time" rather than "British Summer Time".  I have no opinion on which is correct - just saying this is different from how it was.

TheDoubleK:
Either way I do think all three words should be capitalised though - it is a proper name for the time zone.

tombye:
@TheDoubleK I think this deals with the other instance of that capitalisation problem: https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/247#issuecomment-136662723

TheDoubleK:
:+1:

PR #250, opened 2015-09-01
Bump utils to 6.7.1

tombye:
Brings in the fix for a unicode error from the
process that generates a fingerprint for the
static asset files:

https://github.com/alphagov/digitalmarketplace-utils/pull/155


quis:
:+1:

PR #251, opened 2015-09-01
Fix CSRF session expiration error

allait:
#### Add CSRF error handler with log in redirect
If a session expires and the first user action is to POST a form
CSRF check fails to find the session token and shows an error.

Since CSRF is triggered before the login_required decorator the user
gets a CSRF error, even though the problem was caused by the expired
session.

To fix this we use a custom CSRF error handler that checks if session
was expired and shows a log in page instead of CSRF error response.

Since CSRF middleware stores a new token in session before error handler
is called, we need to check for `user_id` key instead of checking if the
session is empty.

#### Remove CSRF token time limit
CSRF tokens are stored in the session cookie, so there's little
security benefit from having an additional expiration time on the
token itself: if the session cookie is leaked the attacker can
generate new CSRF tokens anyway.

Instead, we remove the expiration time on the CSRF token and rely on
session expiration. Once user session is renewed after a fresh log in
a new CSRF token will be generated.

#### Set session lifetime to 4 hours
Moves session expiration time to config.py and sets it to 4 hours.
Expiration time is extended after each request, so the session
only expires if there was a 4 hour gap between requests.

#### Update flask-WTF and remove requests dependency
A newer version of requests is installed by dmutils and the package
is no longer used directly by the app, so there's no need to list it
in requirements.


app/__init__.py
```diff
@@ -48,6 +48,21 @@ def create_app(config_name):
...
+                u'csrf.session_expired: Redirecting user to log in page'
+            )
+
+            return application.login_manager.unauthorized()
+
+        application.logger.info(
+            u'csrf.invalid_token: Aborting request, user_id: %s', session['user_id']
+        )
+
+        abort(400, reason)
```
minglis: Is 400 correct? Just wondering if this will reflect in the right alerting/logging pot. Do we consider this some form of login issue or a bad request? 

Thinking 400 is probably OK, just wanted to mention it.

allait: That's what the original error handler did. 400-403 are covered by the same metric right now, so I think it's ok. We can always use the log message for additional alerts.

robyoung:
:+1:

PR #253, opened 2015-09-01
Change password reset message.

pcraig3:
Previous password reset message lied to users.
Emails have to be associated with supplier accounts -- validity is necessary but not sufficient.


TheDoubleK:
:+1: from me, but maybe run the wording past Cath too, and wait for Travis to be happy.

PR #254, opened 2015-09-02
[#101379472] Highlight assurance questions for missing assurance answers

TheDoubleK:
See: https://www.pivotaltracker.com/story/show/101379472

Previously the main question was highlighted rather than the assurance part of the question.
This will make it clearer which part of the question is missing an answer.

### DIFF
--- a/app/main/helpers/services.py
+++ b/app/main/helpers/services.py
@@ -121,22 +121,27 @@ def upload_draft_documents(service, request_files, section):
 
 def get_section_error_messages(service_content, errors, lot):
     errors_map = {}
-    for error, message_key in errors.items():
-        if error == '_form':
+    for error_field, message_key in errors.items():
+        question_key = error_field
+        if error_field == '_form':
             abort(400, "Submitted data was not in a valid format")
-        else:
-            if error == 'serviceTypes':
-                error = 'serviceTypes{}'.format(lot)
-            elif error in PRICE_FIELDS:
-                message_key = _rewrite_pricing_error_key(error, message_key)
-                error = 'priceString'
-            validation_message = get_error_message(error, message_key, service_content)
-            question_id = service_content.get_question(error)['id']
-            errors_map[question_id] = {
-                'input_name': error,
-                'question': service_content.get_question(error)['question'],
-                'message': validation_message
-            }
+        elif error_field == 'serviceTypes':
+            error_field = 'serviceTypes{}'.format(lot)
+            question_key = error_field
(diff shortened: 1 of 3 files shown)


app/main/helpers/services.py
```diff
@@ -121,20 +121,27 @@ def upload_draft_documents(service, request_files, section):
 
 def get_section_error_messages(service_content, errors, lot):
     errors_map = {}
-    for error, message_key in errors.items():
-        if error == '_form':
+    for error_field, message_key in errors.items():
+        question_key = error_field
+        if error_field == '_form':
             abort(400, "Submitted data was not in a valid format")
         else:
```
allait: There's no need to have nested `if` here, we can just go on with `elif`

app/main/helpers/services.py
```diff
@@ -121,20 +121,27 @@ def upload_draft_documents(service, request_files, section):
...
-                'input_name': error,
-                'question': service_content.get_question(error)['question'],
+            if error_field == 'serviceTypes':
+                error_field = 'serviceTypes{}'.format(lot)
+                question_key = error_field
+            elif error_field in PRICE_FIELDS:
+                message_key = _rewrite_pricing_error_key(error_field, message_key)
+                error_field = 'priceString'
+                question_key = error_field
+            elif error_field.endswith('--assurance'):
```
allait: Does this come out of the API? My impression was that we pack assurance before sending it to the API, so where's the `--assurance` key coming from?

TheDoubleK: Good point - that will simplify things a bit. This fix took a fair bit of trial and error, and I haven't properly tidied it up.

TheDoubleK:
Tidied up a bit now.

robyoung:
This doesn't seem to work for me. Editing a PaaS service trying to submit the Data-in-transit protection page with no assurance answers selected I get the following.

screenshot-localhost 5003 2015-09-04 10-58-47

TheDoubleK:
That's weird/annoying.  I'll take a look again when I get a minute.

quis:
@robyoung I think you're missing some combination of `npm install`/`pip install`/`npm run…`. Works for me locally.

quis:
Reviewed and ran, :+1:

robyoung:
@quis I'm pretty sure I'm up to date, unless there's an additional step I'm not aware of. I have the latest API, pip installed, npm installed, npm run frontend-build:production (although I'm not sure this is required) and I still get the error.

PR #255, opened 2015-09-02
Add missing pageview tracking to analytics

tombye:
This was missed out when the new functionality was added.

To run the tests, open `spec/javascripts/support/LocalTestRunner.html` in a browser.


robyoung:
:+1:

PR #257, opened 2015-09-02
Add link to supplier A–Z

quis:
Since this ([soon won't be](https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/150)) behind a feature flag we can reliably assume, in this app, that it exists and therefore link to it.


TheDoubleK:
:+1:

PR #259, opened 2015-09-02
Add more supplier create logging

robyoung:
Nicole has been receiving reports of users creating their supplier account but not receiving the invite email. It's currently difficult to monitor this because we can't see whether the users ever visited the create-your-account page.

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -343,6 +343,9 @@ def submit_company_summary():
 
 @main.route('/create-your-account', methods=['GET'])
 def create_your_account():
+    current_app.logger.info(
+        "suppliercreate: get create-your-account supplier_id:{}".format(
+            session.get('email_supplier_id', 'unknown')))
     form = EmailAddressForm()
 
     template_data = main.config['BASE_TEMPLATE_DATA']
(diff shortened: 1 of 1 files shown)


app/main/views/suppliers.py
```diff
@@ -343,6 +343,9 @@ def submit_company_summary():
 
 @main.route('/create-your-account', methods=['GET'])
 def create_your_account():
+    current_app.logger.info(
+        "suppliercreate: get create-your-account supplier_id:{}".format(
+            session['email_supplier_id']))
```
TheDoubleK: Might be worth doing a `.get` with a default here, just in case (I know people shouldn't be going direct to the URL without a session in place, but you never know...)

allait: should this be `session.get('email_supplier_id', 'unknown')` too?

TheDoubleK: Although, do we want it to error in that case and not log the get?  In which case we can leave it as is.  It will 500 though I think, which isn't nice.

robyoung: yes it is, good catch

allait:
:+1:

PR #260, opened 2015-09-03
Monitoring fixes

robyoung:
## Rename supplier app to be inline with other alarms

Other alarms use 'supplier frontend' rather than 'supplier app'. This change makes it easier to determining the application that has raised the alarm.

## Raise the failed logins threshold
This alarm is firing continuously on preview and regularly on staging and production. The intention of the alarm is to warn us of a possible attack, 10 requests in a minute is not indicative of that.

### DIFF
--- a/.ebextensions/cwl-logs-application.config
+++ b/.ebextensions/cwl-logs-application.config
@@ -82,7 +82,7 @@ Resources:
         "Fn::Join":
           - ""
           -
-            - "There have been too many failed logins on the supplier app.\n"
+            - "There have been too many failed logins on the supplier frontend.\n"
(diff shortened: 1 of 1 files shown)


TheDoubleK:
:+1: 
Why are there so many failed logins though? 10 a minute seems like a lot. Are people that bad at entering passwords?

allait:
I don't think we had any production alerts on the failed logins. Functional tests trigger it on preview quite regularly, but we ignore the preview alerts anyway. I'm reluctant to change this until we get any context on why this might alert.

robyoung:
Fair point, I've removed raising the failed login threshold.

PR #261, opened 2015-09-04
Add JSON application logs and `login.success`

robyoung:
## Add JSON application logs to CloudWatch
See: https://github.com/alphagov/digitalmarketplace-utils/pull/151

## Add login.success log message
This allows us to calculate a rate metric for login failures as apposed to the absolute number we currently have.


allait:
:+1:

PR #262, opened 2015-09-04
Add fields to apache access log

robyoung:
Add `requestTimeMicro` (the time a request took in microseconds) and `requestPath` (the request path without the query string) to the apache access logs.

`requestPath` is useful for metric filters against a specific page.

`requestTimeMicro` is justified here https://github.com/alphagov/digitalmarketplace-api/commit/9ec62be7f51997e7876e056bbb0236fc6fee5f7e

I have only added these to the JSON logs as they are the format we are moving towards and making changes to the plain text logs requires a bit more thought.

https://httpd.apache.org/docs/2.2/mod/mod_log_config.html#formats


allait:
:+1:

PR #264, opened 2015-09-04
Allow the JavaScript tests to run in CI

tombye:
You can currently only run the JavaScript tests by loading the test runner in a browser.

This adds a Gulp task which uses PhantomJS to run the tests so it can be done from a command as well.

Travis apparently has PhantomJS as part of its environment: http://docs.travis-ci.com/user/ci-environment/#Headless-Browser-Testing-Tools

### Notes

There were a few problems getting the plugin to work with Travis, the details of which (and the reason we're pointing to a Github repo in the `package.json`) is explained here:

http://stackoverflow.com/questions/32461456/travis-ci-can-the-version-of-node-installed-on-all-vm-images-match-the-latest

and here:

https://github.com/alphagov/gulp-jasmine-phantom/pull/1

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,10 +77,17 @@ Note: `npm run frontend-install` is run automatically as a post-install task whe
 
 ### Run the tests
 
+To run the Python tests:
+
 ```
 ./scripts/run_tests.sh
 ```
 
+To run the JavaScript tests:
+
+```
+npm test
+```
 
 ### Run the server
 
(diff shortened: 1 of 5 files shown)


tombye:
Closing this until I can work out how to set Node to 0.12.x in travis (if at all).

quis:
I think you need to add a line to `./scripts/run_tests.sh` to actually make them them run on Travis. I can't see any Jasmine output [here](https://travis-ci.org/alphagov/digitalmarketplace-supplier-frontend/jobs/79316891).

tombye:
@quis good point! I plugged it in in 5ae5a8e so should be done now.

tombye:
OK, I've altered `scripts/run_tests.sh` to better follow the convention and to make it explicit which tests have passed/failed.

quis:
:+1:

PR #265, opened 2015-09-04
Remove CloudWatch filters and alarms

robyoung:
These are being managed by dmaws now.

The reset email sent and reset email invalid metrics have been dropped
as it wasn't clear what they were providing. If they are needed they can
be added to dmaws.


allait:
:+1:

PR #269, opened 2015-09-08
[#102167826] Fix invitation link behaviour

pcraig3:
This pull request addresses incorrect messaging we were potentially sending to invite-link-clicking enthusiasts.

For example, if a new user

1. clicked an invite link
2. created an account
3. clicked the invite link again

they would end up on an error page telling them their account was locked or inactive, which it clearly wasn't. 

The following table should help: first column outlines a scenario, second column describes the result.  (Screenshots below.)

State of the user when they click a valid invite link | Result
------------ | -------------
Totally new user | sees the Create User Form
User with a Buyer account | sees the Update User Form (which converts a `buyer` account to a `supplier` account).
Locked user | sees an error message telling them they're locked
Deactivated user | sees an error message telling them they're deactivated
User registered to one supplier but invited by a different supplier | sees an error message telling them they're currently registered to one supplier but were invited by another
Valid supplier user (not logged in) | sees the Login page
Valid supplier user (logged in) | sees the supplier dashboard

All of the onscreen messages either went through @ralph-hawkins or @superroz.  There's a heavily-edited document hanging around if anyone wants to see it.

[>> Story in Pivotal](https://www.pivotaltracker.com/story/show/102167826)

***

# Screenshots 

### i. Totally new user 
screen shot 2015-09-08 at 19 16 00

### ii. User with a Buyer account
screen shot 2015-09-08 at 19 14 55

### iii. Locked user
screen shot 2015-09-08 at 19 20 04

### iv. Deactivated user 
screen shot 2015-09-08 at 19 18 27

### v. User registered to one supplier but invited by a different supplier 
screen shot 2015-09-08 at 19 21 34

### vi. Valid supplier user (not logged in)
screen shot 2015-09-08 at 19 17 44

### vii. Valid supplier user (logged in)
screen shot 2015-09-08 at 19 16 58

***

### \*bonus\* Slightly modified login error message \*/bonus\*
screen shot 2015-09-08 at 19 29 10

### DIFF
--- a/app/main/views/login.py
+++ b/app/main/views/login.py
@@ -195,7 +195,14 @@ def create_user(encoded_token):
 
     if token is None:
         current_app.logger.warning("createuser.token_invalid: %s", encoded_token)
-        _token_error("auth/create-user.html", form=form, template_data=template_data)
+        flash('token_invalid', 'error')
+        return render_template(
+            "auth/create-user.html",
+            form=form,
+            token=None,
+            email_address=None,
+            supplier_name=None,
+            **template_data), 400
 
     else:
         user_json = data_api_client.get_user(email_address=token.get("email_address"))
@@ -255,7 +262,13 @@ def submit_update_user(encoded_token):
     token = decode_invitation_token(encoded_token)
     if token is None:
         current_app.logger.warning("createuser.token_invalid: %s", encoded_token)
-        _token_error("auth/update-user.html", template_data=template_data)
+        flash('token_invalid', 'error')
+        return render_template(
+            "auth/update-user.html",
+            token=None,
+            email_address=None,
+            supplier_name=None,
+            **template_data), 400
 
     else:
         user_json = data_api_client.get_user(email_address=token.get("email_address"))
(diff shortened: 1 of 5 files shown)


superroz:
All good apart from v. which can be simplified. 

Could be: Log in to your [Digital Marketplace Team] account or email enquiries@digitalmarketplace.service.gov.uk to register your account with a different organisation.

Will update the file directly if I can.

TheDoubleK:
Just a thought... but would it be better to prevent invitations from being sent to people with existing supplier accounts in the first place? We could do a lookup before any email is sent.

TheDoubleK:
The messaging isn't quite right if someone is trying to sign up with an admin account email address:
screen shot 2015-09-09 at 13 42 16

Admittedly this is an extreme edge case - i.e. it would only ever happen if Nicole or CCS tried to invite themselves to an account.

So I'm happy for this not to get fixed...

TheDoubleK:
I'm not sure about "There was a problem with the account details you provided." - to me this sounds like a problem with the account itself rather than something that could be fixed by re-typing the log-in details again.

Could we add "Please try re-entering your email address and password" before the account-could-be-locked blurb?  Otherwise I think people might email Nicole without even trying to log in again.

superroz:
Cool. Let's use: 

```html
<p>Make sure you've entered the right email address and password.</p> 
<p>Accounts are locked after 5 failed attempts. If you think your account has been locked, email enquiries@digitalmarketplace.service.gov.uk.</p>
```

ralph-hawkins:
:+1

TheDoubleK:
:+1:

PR #271, opened 2015-09-09
Change signup order

quis:
_These changes are based on:_
- _feedback that has come into enquiries@digitalmarketplace.service.gov.uk_
- _looking at the funnels in analytics_
- _some guerilla research that @ralph-hawkins and I did yesterday_

## Reorder account creation methods

This just changes the order in the code, so that the functional changes in subsequent commits will be easier to review. This pull request will be easier to review on a commit-by-commit basis.

## Move personal account page before summary page

img_6793

We have had some users dropping out of the supplier creation process after they have clicked the green button on the summary page.

We posit that this is because:
- they think they're done at this point
- we don't do a good job of explaining why you need to enter an email address for the second time

This commit changes the order of the pages to put the summary last, and not create the supplier until we have both:
- the company details
- and the login details

It also adds a new section to the summary page showing the user the email address that they have entered for their login. This:
- reinforces that we have asked for two email addresses and that they are for different reasons
- allow them to go back and change their login address if they have made a typo

## Make content match new order

Some of the messages no longer made sense because they referred to parts of the process, the order of which has changed.

We also removed some messages which weren't immediately relevant to the user’s next steps. While the information they conveyed was important, it could be pushed down to the point at which it was most relevant.

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -274,7 +274,7 @@ def submit_company_contact_details():
         session[form.email_address.name] = form.email_address.data
         session[form.phone_number.name] = form.phone_number.data
         session[form.contact_name.name] = form.contact_name.data
-        return redirect(url_for(".company_summary"))
+        return redirect(url_for(".create_your_account"))
     else:
         current_app.logger.warning(
             "suppliercreate.fail: duns:%s company_name:%s %s",
@@ -299,6 +299,7 @@ def create_your_account():
     return render_template(
         "suppliers/create_your_account.html",
         form=form,
+        email_address=session.get('account_email_address', ''),
         **template_data
     ), 200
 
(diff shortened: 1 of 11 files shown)


ralph-hawkins:
:+1:

app/main/views/suppliers.py
```diff
@@ -325,108 +363,66 @@ def submit_company_summary():
...
-            send_email(
-                form.email_address.data,
-                email_body,
-                current_app.config['DM_MANDRILL_API_KEY'],
-                current_app.config['CREATE_USER_SUBJECT'],
-                current_app.config['RESET_PASSWORD_EMAIL_FROM'],
-                current_app.config['RESET_PASSWORD_EMAIL_NAME'],
-                ["user-creation"]
+            email_body = render_template(
+                "emails/create_user_email.html",
```
TheDoubleK: I've just noticed that this template doesn't wrap the URL in an anchor tag at all.  I guess *most* mail clients will turn it into a link, but we might want to add it while we're fixing this up?

TheDoubleK: The same goes for invite_user and change_password email templates.

TheDoubleK:
Tested locally and the flow is much better - nice work.

quis:
@cathrooney Noticed that some of the `<title>`s on these pages were wrong. Fixed in f42e181.

TheDoubleK:
:+1:

PR #272, opened 2015-09-09
[#102983572] Remove validation of VAT number

TheDoubleK:
See: https://www.pivotaltracker.com/story/show/102983572

Our rules were too strict for non-UK companies, so we're not going to validate this field any more.


quis:
:+1:

PR #273, opened 2015-09-09
Validation tweaks for none of the above

TheDoubleK:
Adding 'none of the above' option means that the rules for the dependent question on that page need to be a bit smarter.

Also pulls in latest content with the 'none of the above' option.


quis:
Content changes: https://github.com/alphagov/digitalmarketplace-ssp-content/compare/aa1b1c41b8c6872c852ed6d6f38c4e26fefc6ed1...3351a71d67e482a51bda3933ae9e6e1c120dac00

quis:
:+1:

PR #275, opened 2015-09-09
Update Node modules

tombye:
Replaced the deprecated gulp-uglifyJS with gulp-uglify, as per the following warning message it was giving out:

`npm WARN deprecated gulp-uglifyjs@0.6.0: Since gulp-sourcemaps now works, use gulp-uglify instead`

While I was there I also updated Bower as it was also complaining of being a few releases behind and added sourcemaps to the JavaScript compression process.

If this is accepted I'll push it out to the other apps.


quis:
:+1:

quis:
Sourcemaps: :cool:

PR #276, opened 2015-09-09
Improve and standardise logs

robyoung:
This is a broad commit that does a few things.

Only put hashes of email addresses in logs. This means that we are not storing personal information in logs but we can search for log messages by email address if we hash the address locally first.

Extract variables out into extra parameters. The logging in dmutils has been changed to attempt to format log messages with the log record (containing all log variables). This means that if a log message is provided as a template and the variables provided in the extra dict then the message will be rendered correctly and the variables will be pulled out in the JSON logs.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ Flask-Script==2.0.5
 Flask-WTF==0.12
 
 git+https://github.com/madzak/python-json-logger.git@v0.1.3#egg=python-json-logger==v0.1.3
-git+https://github.com/alphagov/digitalmarketplace-utils.git@6.8.1#egg=digitalmarketplace-utils==6.8.1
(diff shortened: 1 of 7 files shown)


requirements.txt
```diff
@@ -4,7 +4,7 @@ Flask-Script==2.0.5
 Flask-WTF==0.12
 
 git+https://github.com/madzak/python-json-logger.git@v0.1.3#egg=python-json-logger==v0.1.3
-git+https://github.com/alphagov/digitalmarketplace-utils.git@6.8.1#egg=digitalmarketplace-utils==6.8.1
+git+https://github.com/alphagov/digitalmarketplace-utils.git@6.9.0#egg=digitalmarketplace-utils==6.9.0
```
TheDoubleK: Utils is at 6.9.1 now, although not desperate to bring in the patch change right now particularly.

TheDoubleK:
Looks good - with or without the Utils version bump - :+1:

PR #277, opened 2015-09-10
Give the JS tests their own exit status

tombye:
It's useful to have something that points to the
testsuite that's being run, as mentioned in:

https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/159#discussion_r39135917

This is now in the buyers app so we should also have it here.


quis:
:+1:

PR #278, opened 2015-09-10
Update password-reset error messages.

pcraig3:
"[Some of the wording we're using in our user-facing messaged is not ideal](https://www.pivotaltracker.com/story/show/97612300)", and though some of it had been looked at, there were a few places left over that weren't.

We're not referring explicitly to forms any longer, but rather asking users to do something with the form.
Also, removed the bit about getting in touch from the password reset email unless getting in contact would be necessary.

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/97612300)

### DIFF
--- a/app/templates/auth/request-password-reset.html
+++ b/app/templates/auth/request-password-reset.html
@@ -18,11 +18,12 @@
                         we'll send a link to reset the password.
 
                         {% elif message == 'token_expired' %}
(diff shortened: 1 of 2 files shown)


superroz:
Changed Team to team in the email. But otherwise: 👍

pcraig3:
Not really being an Admin Team player, here @superroz.

PR #279, opened 2015-09-10
[#101673638] Indent list items with validation errors

TheDoubleK:
See this bug: https://www.pivotaltracker.com/story/show/101673638

Previously the validation error red bars overlapped the outdented bullet points on list items.
Now bullets are indented if there is a validation error for the question.

Before: 
screen shot 2015-09-10 at 16 59 25

After:
screen shot 2015-09-10 at 16 58 42


quis:
:+1:

PR #280, opened 2015-09-11
Give all submit buttons a submit attribute

tombye:
IE7 needs submit buttons to have an attribute of `type="submit"` for them to work and a lot of ours don't.

This fixes that by adding the necessary attribute where needed.

### DIFF
--- a/app/templates/auth/create-user.html
+++ b/app/templates/auth/create-user.html
@@ -67,7 +67,13 @@
         </div>
     {% endif %}
 
-        <button type="submit" class="button-save">Create contributor account</button>
+        {%
+          with
+          type = "save",
+          label = "Create contributor account"
+        %}
+          {% include "toolkit/button.html" %}
+        {% endwith %}
     </form>
 
         </p>
--- a/app/templates/auth/create-user.html
+++ b/app/templates/auth/create-user.html
@@ -67,7 +67,7 @@
         </div>
     {% endif %}
 
-        <button class="button-save">Create contributor account</button>
+        <button type="submit" class="button-save">Create contributor account</button>
     </form>
 
         </p>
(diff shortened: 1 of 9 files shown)


quis:
These should be converted to use [the button pattern from the toolkit](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/button.html) which [fixes this bug already](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/146).

tombye:
@quis switched to use those partials in 310835a

quis:
:+1: 

You should give @andrewchong a heads up before you merge/deploy this because it's very likely to have an impact on the tests.

tombye:
@quis I've had a look through the functional tests and, as far as I can see, the targeting of buttons is limited to the following methods:

- [has_button?](http://www.rubydoc.info/github/jnicklas/capybara/Capybara/Node/Matchers#has_button%3F-instance_method)
- [click_link_or_button](http://www.rubydoc.info/github/jnicklas/capybara/Capybara/Node/Actions:click_link_or_button)
- [have_button](http://www.rubydoc.info/github/jnicklas/capybara/Capybara/RSpecMatchers:have_button)

...all of which use the `:button` or `:link_or_button` selectors, the button bit of which matches either input or button tags.

(See https://github.com/jnicklas/xpath/blob/59badfa50d645ac64c70fc6a0c2f7fe826999a1f/lib/xpath/html.rb#L22).

I'll still talk to @andrewchong but thought it was worth checking first.

andrewchong:
@tombye I think we are fine with this in terms of impact on the functional test. I believe I have already addressed the button issue from the last time @quis brought my attention to the changes he made. If there is any impact from your changes the resolution will not be a big problem.

PR #281, opened 2015-09-14
[#103193094] Prevent text being obscured on assurance validation error

TheDoubleK:
Fixes this bug: https://www.pivotaltracker.com/story/show/103193094

Before (Firefox):
screen shot 2015-09-14 at 09 40 52

After (Firefox):
screen shot 2015-09-14 at 09 41 20


quis:
Checked out locally and tested in a few versions of IE, :+1:

PR #282, opened 2015-09-15
Allow refresh on supplier account create completion

robyoung:
We have been seeing exceptions in production for the supplier account creation complete view. It seems this is caused by a refresh because the entire session is cleared. Ensuring the `email_sent_to` field is re-added to the session means that at least this page can be refreshed.


TheDoubleK:
:+1:

PR #285, opened 2015-09-17
Track when users click buttons

quis:
It would be very useful to know, for example, the split between how many users are clicking the 'save and continue' button and how many are clicking the 'save and return to service overview'.

This commit adds code similar to the link tracking code to fire an event when any button is clicked.

By recording the text of the button and the URL of the page that the user is on we can filter the data to work out what's going on.

### DIFF
--- a/app/assets/javascripts/analytics/_events.js
+++ b/app/assets/javascripts/analytics/_events.js
@@ -98,6 +98,10 @@
     return ( positionTop > $root.scrollTop() && positionTop < ($root.scrollTop() + $root.height()) );
   };
 
+  GOVUK.GDM.analytics.isQuestionPage = function(url) {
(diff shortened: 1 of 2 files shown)


tombye:
:+1: for the code changes. I'm surprised we can't track this through the normal page views sent, seeing as we're just trying to see the next page requested after each edit page.

TheDoubleK:
This kind of dark art is why I have an ad-blocker.

quis:
## Only track buttons on question pages

…slightly less dark.

This also removes the tracking of the URL in the label parameter (because the URLs will be broadly the same) and replaces it with the page title, as requested by Ash.

tombye:
:+1:

PR #286, opened 2015-09-18
Tweak copy to match updated version in story

TheDoubleK:
This is what happens when the story description gets updated after you've already copy-and-pasted the required content out of the story.


quis:
:+1:

PR #288, opened 2015-09-21
[#103088520] Close G7 submissions

TheDoubleK:
See: https://www.pivotaltracker.com/story/show/103088520

There is a further story for @ralph-hawkins and/or @superroz to review the copy for G7-switchoff here: https://www.pivotaltracker.com/story/show/103088582

The copy changes in this pull-request are **TheDoubleK** specials.

The actual switch-off will be done by changing the `status` of the `g-cloud-7` framework from `open` to `pending`.  We should be able to merge and test this code on preview/staging by manually changing the database.  Once tested we should change the framework status back again, as the functional tests will fail once G7 submissions are closed.

With no change to the database then everything should be the same as it is now.
Once the framework is pending the following changes should be seen:

On `/suppliers`:
* The link to  "Continue your G-Cloud 7 application" / "Register your interest in becoming a G-Cloud 7 supplier" changes to "View your G-Cloud 7 application", without and sub-text or hint text.

screen shot 2015-09-21 at 15 25 20

On `/suppliers/frameworks/g-cloud-7`:
* The heading "Apply to G-Cloud 7” changed to “Your G-Cloud 7 application” 
* The "Make/Edit supplier declaration" section is gone
* "Deadline 3pm BST, 6 October 2015" changed to "Deadline for submissions passed at 3pm BST, 6 October 2015"
* "Add, edit and delete services" link changed to "View services"

screen shot 2015-09-21 at 10 20 01

On `/suppliers/frameworks/g-cloud-7/services`:
* "You need to make the supplier declaration…" warning banner removed
* "Make a copy" buttons gone
* "Service can be marked as complete" text gone

screen shot 2015-09-21 at 14 10 20

On `/suppliers/submission/services/<id>`:
Removed:
* “Mark as complete” button at the top of the page
* “Mark as complete” button at the bottom of the page
* “Delete this service” button from the bottom of the page
* “Edit” links from the right-hand side of each section heading

In addition to this all endpoints relating to editing the supplier declaration or making updates to services will return a 404 response once the framework is `pending`.

The required API and Utils changes required for this to work are already merged and live:
- [x] https://github.com/alphagov/digitalmarketplace-api/pull/248
- [x] https://github.com/alphagov/digitalmarketplace-utils/pull/163

### DIFF
--- a/app/templates/frameworks/dashboard.html
+++ b/app/templates/frameworks/dashboard.html
@@ -20,17 +20,25 @@
 {% endblock %}
 
 {% block main_content %}
-
-  {% with
-     heading = "Apply to G-Cloud 7",
-     smaller = True,
-     with_breadcrumb = True
-  %}
-    {% include "toolkit/page-heading.html" %}
-  {% endwith %}
-
+  {% if g7_status == 'open' %}
+    {% with
+       heading = "Apply to G-Cloud 7",
+       smaller = True,
+       with_breadcrumb = True
+    %}
+      {% include "toolkit/page-heading.html" %}
+    {% endwith %}
+  {% else %}
+    {% with
+      heading = "Your G-Cloud 7 application",
+      smaller = True,
+      with_breadcrumb = True
+    %}
+      {% include "toolkit/page-heading.html" %}
+    {% endwith %}
+  {% endif %}
   <aside role="complementary" class="framework-application-status">
-    Deadline <strong>{{ deadline|safe }}</strong>
+    Deadline {% if g7_status == 'pending' %} for submissions passed at {% endif %} <strong>{{ deadline|safe }}</strong>
   </aside>
 
   <nav role="navigation">
@@ -41,7 +49,15 @@
             {% if not counts.draft %}
               <div class="framework-section-validation-bar-good">
             {% endif %}
-              <a class="browse-list-item-link" href="{{ url_for('.framework_services') }}"><span>Add, edit and delete services</span></a>
+              <a class="browse-list-item-link" href="{{ url_for('.framework_services') }}">
+                <span>
+                {% if g7_status == 'open' %}
+                  Add, edit and delete services
+                {% else %}
+                  View services
+                {% endif %}
+                </span>
+              </a>
             {% if not counts.draft %}
               </div>
             {% endif %}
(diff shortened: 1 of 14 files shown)


robyoung:
What status will the framework be changed to when it's no longer open?At the moment we have `pending`, `open`, `live` and `expired`. `pending` was used for before the framework was opened, is it ok to use the same state for the period between `open` and `live`?

I think that if we have more states a framework can be in multiple if statements will become harder to work with. Under what conditions would the framework dashboard be available? Is the plan to be able to view the G-Cloud 7 dashboard even after the framework has gone live or would the services just be visible through 'Current services' on the supplier dashboard? If so, where would they view their supplier declaration?

app/templates/frameworks/dashboard.html
```diff
@@ -20,17 +20,25 @@
...
+      heading = "Your G-Cloud 7 application",
+      smaller = True,
+      with_breadcrumb = True
+    %}
+      {% include "toolkit/page-heading.html" %}
+    {% endwith %}
+  {% endif %}
   <aside role="complementary" class="framework-application-status">
-    Deadline <strong>{{ deadline|safe }}</strong>
+    Deadline {% if g7_status == 'pending' %} for submissions passed at {% endif %} <strong>{{ deadline|safe }}</strong>
```
robyoung: Is this correct? I thought pending was the state the framework was in before it went live. I remember discussing framework statuses with @cathrooney but I can't remember the details.

TheDoubleK:
@minglis said that `pending` will be the status for the standstill period - I didn't realise it was also used before submissions opened, but probably should have.

I think it's OK to stick with `pending` for now, although if there will be some overlap of DOS with G7 standstill maybe we do need a new status for the standstill period?

As I understand it some form of G7 dashboard will still exist once the framework is open - this is where suppliers will upload and download the PDF framework agreements.  It probably won't have the services and drafts on any more, as live services will be on the main dashboard and we'll probably throw away draft services that didn't make the cut (although none of this is finalised AFAIK).

TheDoubleK:
Just spoke to @cathrooney and @ralph-hawkins about this - we're going to stick with `pending` for here as it feels like the best fit, and then create a new status for before DOS opens (something like `in-preparation` but will be finalised later).

pcraig3:
It seems like a lot of the stuff the Feature Flags plugin does is being replicated.
- Routes decorated with a falsey FF flag value will throw a 404.
- A FF can be used as a conditional operator in a template.

I understand the distinction between having to deploy an app with a changed FF value versus flicking a switch in the database, but would it be worth, for example, trying to base the value of the `GCLOUD7_OPEN` flag on the whether g-cloud 7 was `open` in the frameworks table?  That way we could avoid, for example, needing the `g_cloud_7_is_open_or_404`.

(Of course, it's also possible that trying to get the `GCLOUD7_OPEN` flag to be something else would require just as much code.)

Any thoughts?

quis:
I'd strongly suggest not having feature flags which are tied to a specific framework. Imagine if, in the future, we had an admin page where you could manage the status of all frameworks—you'd want those changes reflected on the site without having to deploy any code.

robyoung:
@TheDoubleK OK, so we take another look at the templates as part of the DOS work. Sounds sensible. 

:+1:

PR #289, opened 2015-09-22
Add grids to forms content

tombye:
Wraps all form content in a two-thirds grid so their width is dictated by the grid instead of part of their styling.

Part of https://www.pivotaltracker.com/story/show/103093496

Brings in this work from the DM Toolkit: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/158

This pull request also includes a shim for the `console` object so browsers that don't support it won't throw an error when it is used.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -22,6 +22,14 @@
 
   var module;
 
+  if(typeof console === 'undefined') {
+    console = {
+      log: function () {},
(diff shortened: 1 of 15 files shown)


quis:
Checked out locally and browsed through the forms. Generally looks good.

Forms like this are less than ideal because the text box is much wider than the content. But :+1: to this PR in the name of progress.
image

PR #290, opened 2015-09-22
Switch on PROD feature flag to turn off clarification questions

TheDoubleK:
This needs to be deployed at 5pm today.


robyoung:
:+1:

PR #292, opened 2015-09-23
Bring in corrected grid classes

tombye:
Brings in https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/161 which moves the grid class names used in this repo to the DM Toolkit.

Fixes this problem:

image

...which came as a result of this https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/158 using these classes without them being added to the toolkit


quis:
:+1:

PR #293, opened 2015-09-24
[#103779820] Improve messaging for removed users.

TheDoubleK:
Part of this bug fix: https://www.pivotaltracker.com/story/show/103779820

When a supplier removed a user from their supplier account the message said that they could re-invite the user at any time.  This isn't entirely true, as Nicole will have to re-activate the user first.

Rather than spell this out in a lengthy flash message (as re-adding a removed user is an unlikely edge case), it's better to just confirm that the user has been removed.

Depends on: 
- [x] https://github.com/alphagov/digitalmarketplace-utils/pull/166/files


quis:
:+1:

PR #294, opened 2015-09-24
Display error when entering existing email address

quis:
screen shot 2015-09-24 at 16 08 20

--

Just wrote all this sweet code and then had a thought: **is this a security hole because we're exposing which accounts exist?**

--

When a user enters an email address which is already associated account we should stop them.

This is relevant:
- when creating a new supplier account
- when inviting users to an existing supplier

--

@superroz How's the words?

--

Fulfils https://www.pivotaltracker.com/story/show/103066110

### DIFF
--- a/app/main/helpers/__init__.py
+++ b/app/main/helpers/__init__.py
@@ -8,3 +8,11 @@ def hash_email(email):
     m.update(email.encode('utf-8'))
 
     return base64.urlsafe_b64encode(m.digest())
+
+
+def is_existing_supplier_user(user):
+    if not user:
+        return False
+    if user['users'].get('role') is 'supplier':
+        return True
+    return False
--- a/app/main/views/login.py
+++ b/app/main/views/login.py
@@ -369,7 +369,11 @@ def invite_user():
 def send_invite_user():
     form = EmailAddressForm()
 
-    if form.validate_on_submit():
+    existing_user = data_api_client.get_user(
+        email_address=form.email_address.data
+    )
+
+    if form.validate_on_submit() and not existing_user:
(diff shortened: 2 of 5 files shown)


TheDoubleK:
Yes this came up in the sprint planning meeting that we might not want to do this after all - think the decision was maybe we should send an email saying "You've been invited to supplier X but already have an account"

minglis:
Yeah I don't think we can do this. We shouldn't allow email discovery through our tools.

Putting it in the email is one option, alternativly changing the wording on the landing page might also be a thing? At that point we know they've the email and we know the error, so we can help them make the right request of support?

quis:
What happens if I'm already with a supplier and get invited to another one? Does my account get attached to the new one?

allait:
@quis I don't think it is reattached right now and I don't think it should be. We could send a message to the email address saying that it's used with another supplier, but reattaching could potentially be used for phishing.

TheDoubleK:
There is an error page when the link is the email is clicked that says it might be attached to another supplier.  And to contact Nicole to fix it up.  Which I think is the right thing.

pcraig3:
Just in regards to @TheDoubleK's comment:

https://cloud.githubusercontent.com/assets/2454380/9743679/7babd84a-565f-11e5-9175-e72aafeb9b0d.png

TheDoubleK:
Thanks @pcraig3 - nice to see you doing the hard work to make my life simple :)

quis:
Seems like the stuff that Martyn is talking about is already in place. Going to close this for now and add a note of uncertainty to the story.

PR #295, opened 2015-09-29
Fix updating of password

quis:
Brings in:
- [x] https://github.com/alphagov/digitalmarketplace-utils/pull/170

Discussion point: is it right to say it's the user who is updating their password here—they're not logged in—is it something we can really know?


TheDoubleK:
As they have clicked the link from an email received at that email address I think it is OK to say that that email address is updating the password.

TheDoubleK:
:+1: once Travis is happy.

PR #296, opened 2015-09-29
[#104451064] BUGFIX (urgent): There may not be a logged-in user for user updates

TheDoubleK:
This method was assuming that `current_user` exists, but this method doesn't require login and breaks if called without a `current_user`.

500s in the logs caused by:
```
'AnonymousUserMixin' object has no attribute 'email_address'
```


quis:
:+1:

PR #299, opened 2015-10-01
[#104308518] Prevent 500s if session expired on create-your-account-complete page

TheDoubleK:
See: https://www.pivotaltracker.com/story/show/104308518

Before:
screen shot 2015-10-01 at 16 16 43

After:
screen shot 2015-10-01 at 15 48 50


pcraig3:
Why does this happen, do we know?

TheDoubleK:
I can only think because people refresh the page after their session has expired. Or maybe because the bookmarked that last page and try to visit it again later.  I haven't tried extracting user journeys from the logs leading up to these errors or anything though.

pcraig3:
Well at that point everything is in the database anyway, right?

:+1: from me.

PR #300, opened 2015-10-05
Use shared service helpers from dmutils

robyoung:
Functions that are also required by the admin frontend have been moved to dmutils.


allait:
:+1:

PR #304, opened 2015-10-07
Upgrade to dmutils 9.0.0

robyoung:
This brings in alphagov/digitalmarketplace-utils#178 which changes the `DataAPIClient` to match the language we use around the supplier declaration.

## :sparkles: This has been tested locally with the functional tests :sparkles: 

Using the SSP tests with the framework set to 'open'.


allait:
:+1:

PR #306, opened 2015-10-15
Upgrade dmutils to 10.0.1

robyoung:
This contains a few [version bumps in dmutils](https://github.com/alphagov/digitalmarketplace-utils/compare/9.0.0...10.0.1).

The most significant is the 10.0.0 change https://github.com/alphagov/digitalmarketplace-utils/pull/182


quis:
I've been using alphagov/digitalmarketplace-utils#182 locally in this app for a couple of days. Works fine. :+1:

PR #309, opened 2015-10-19
Fix bug #105988832: Communications files erroring

robyoung:
When the signature of `get_draft_document_url` was changed the download supplier file route was not updated.


allait:
:+1:

PR #311, opened 2015-10-20
Add prototype of DOS declaration

quis:
The aim of merging this is to stop it from getting too stale.

It's behind a feature flag so that it won’t show up in the live app, but is easy to keep pushing to Heroku to preview the content.

--

It adds:
- latest content (https://github.com/alphagov/digitalmarketplace-frameworks/compare/b1dd57d47a01807ac4aff2c4e693ed2680a19af3...7e11c1a5b814e86b8bb234ee7a5c9156233cbcd4)
- smarter question numbering (alphagov/digitalmarketplace-utils#186)
- a template filter for replacing a question reference (`termsAndConditions`) with a question number (`23`)
- section descriptions (https://github.com/alphagov/digitalmarketplace-utils/pull/188)
- a temporary ✎ link which goes to the content file on Github


config.py
```diff
@@ -64,6 +64,7 @@ class Config(object):
     FEATURE_FLAGS_EDIT_SERVICE_PAGE = False
     FEATURE_FLAGS_GCLOUD7_OPEN = False
     FEATURE_FLAGS_G7_CLARIFICATIONS_CLOSED = False
+    FEATURE_FLAGS_SHOW_DOS_DECLARATION = False
```
TheDoubleK: Could we have it "true" in Development?  Seems like we want to see all the stuff that's still being worked on?

app/main/views/frameworks.py
```diff
@@ -103,6 +103,43 @@ def framework_services(framework_slug):
     ), 200
 
 
+@main.route('/frameworks/digital-outcomes-and-specialists/declaration/<string:section_id>',
```
TheDoubleK: This will want to work from the existing route `'/frameworks/<string:framework_slug>/declaration/<string:section_id>` - I appreciate this is intended for prototyping, but is there a danger that too much will go in here that won't easily generalise to that route?  There doesn't look to be anything here that couldn't work in that general route - is there some other reason for hard-coding the framework in here?

TheDoubleK:
Not really related to this PR, but it seems weird having the two optional "cyber security" questions - they are Yes/No. Anyone who has it will hit "Yes" and anyone who doesn't will leave it blank to avoid saying "No".
If it's optional then why are we asking it at all?

app/__init__.py
```diff
@@ -76,6 +76,14 @@ def refresh_session():
     def markdown_filter(data):
         return Markup(markdown(data))
 
+    @application.template_filter('question_references')
+    def question_references_filter(data, get_question):
+        return re.sub(
+            r"\[\[([^\]]+)\]\]",
```
TheDoubleK: Nice regex... maybe we could have a comment in the code to explain what it does? This feels quite obscure.

package.json
```diff
@@ -11,7 +11,7 @@
     "gulp": "3.8.7",
     "gulp-filelog": "0.4.1",
     "gulp-include": "1.1.1",
-    "gulp-sass": "1.3.3",
+    "gulp-sass": "2.0.4",
```
pcraig3: Last week I had to downgrade my version of node because `gulp-sass` 1.3.3  wasn't recent enough for it.
@tombye mentioned that Travis only runs node v0.10.* or something.  

Do you know if this runs everywhere it's supposed to?

TheDoubleK:
As a WIP I guess you already know it dies after the "Leftovers" page (declaration summary page to come...)

Also, I noticed this bug with the question numbering the other day - not the end of the world, but might be worth a look as the declaration is built out further): https://www.pivotaltracker.com/story/show/106191644

config.py
quis: Yeah, sure, d0efc84

app/main/views/frameworks.py
quis: The reason is so, in the prototype, you can click through the 5 pages without any validation. That's the only meaningful difference.

TheDoubleK: Ah OK, fair enough.

app/__init__.py
quis: 3d260a0

package.json
quis: `gulp-sass` 1.3.3 "requires" Node 0.12.5. But I'm running 0.10.35 locally and it's still working. And Node itself is at 4.2.1 now so shrug.

PR #312, opened 2015-10-21
𝘗𝘙𝘖𝘗𝘖𝘚𝘈𝘓: Restart the app when content changes

quis:
_This is just hacked together, but it works. If we do it for real it should probably go in utils or something._

Any changes to files in `app/content` now cause the app to restart, in the same way that changing a `.py` file causes the app to restart.

This means:
- that when you `npm install` you don't need to restart the app
- you can quickly try out content changes locally and see the changes with just
  a refresh of the page

### DIFF
--- a/application.py
+++ b/application.py
@@ -3,7 +3,23 @@
 import os
 import re
 from app import create_app
-from flask.ext.script import Manager, Server
+from flask.ext.script import Manager, Server as _Server
+
+
+def extra_files():
+    for dirname, dirs, files in os.walk('./app/content'):
+        for filename in files:
+            filename = os.path.join(dirname, filename)
+            if os.path.isfile(filename):
+                print(filename)
+                yield filename
+
+
+class Server(_Server):
+    def __init__(self, *args, **kwargs):
+        if kwargs.get('use_reloader', True):
+            kwargs.setdefault('extra_files', []).extend(extra_files())
+        super(Server, self).__init__(*args, **kwargs)
 
 application = create_app(
     os.getenv('DM_ENVIRONMENT') or 'development'
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -263,12 +263,14 @@ gulp.task('test', function () {
 gulp.task('watch', ['build:development'], function () {
   var jsWatcher = gulp.watch([ assetsFolder + '/**/*.js' ], ['js']);
   var cssWatcher = gulp.watch([ assetsFolder + '/**/*.scss' ], ['sass']);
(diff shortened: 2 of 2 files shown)


application.py
```diff
@@ -3,7 +3,25 @@
...
+
+def get_extra_files():
+    def extra_files():
+        for dirname, dirs, files in os.walk('./app/content'):
+            for filename in files:
+                filename = os.path.join(dirname, filename)
+                if os.path.isfile(filename):
+                    print(filename)
+                    yield filename
+    return extra_files()
```
robyoung: Why do we need the nested function here? We just call it on return.

quis: My still-hacky understanding of how generators work. Squashed it out in 63edd74.

gulpfile.js
```diff
@@ -263,12 +263,14 @@ gulp.task('test', function () {
 gulp.task('watch', ['build:development'], function () {
   var jsWatcher = gulp.watch([ assetsFolder + '/**/*.js' ], ['js']);
   var cssWatcher = gulp.watch([ assetsFolder + '/**/*.scss' ], ['sass']);
+  var frameworkWatcher = gulp.watch([ bowerRoot + '/digitalmarketplace-frameworks' + '/**/*.yml' ], ['copy:frameworks']);
```
allait: We should probably use `sspContentRoot` here in case the repository name changes again in the future

application.py
```diff
@@ -3,7 +3,23 @@
...
-from flask.ext.script import Manager, Server
+from flask.ext.script import Manager, Server as _Server
+
+
+def extra_files():
+    for dirname, dirs, files in os.walk('./app/content'):
+        for filename in files:
+            filename = os.path.join(dirname, filename)
+            if os.path.isfile(filename):
+                print(filename)
```
allait: This sounds like a lot of output for `content`, I think we can drop the print

application.py
```diff
@@ -3,7 +3,23 @@
...
+                yield filename
+
+
+class Server(_Server):
```
allait: I'd go with a different class name since we don't really need to pretend that it's the original `Server` class

allait:
This is great :sparkling_heart:  and should be immediately useful for frontend apps, API (json schemas) and search API (mappings), so I say we move it to `dmutils/flask_init.py` right away! This way each app could pass in the additional folders list to `init_manager`.

There's a not on the werkzeug docs about reloader working better with `watchdog` installed, so it might be worth adding it to `requirements_for_tests.txt` files

quis:
Closing in favour of https://github.com/alphagov/digitalmarketplace-utils/pull/193

PR #313, opened 2015-10-21
[#105005134] Register for DOS

TheDoubleK:
Completes this story: https://www.pivotaltracker.com/story/show/105005134

Further design for displaying both DOS and G-Cloud messages on the same page is forthcoming.  For now I have simply put the DOS messages above the existing G-Cloud 7 messages.

I have also changed registering interest to only happen on a POST request from the button in the "register interest" message - for G7 interest was registered with all GET requests to the framework dashboard. 

Also removes the dependency on `'GCLOUD7_OPEN' is active_feature` from the dashboard template - all messaging is now conditional on the status of the frameworks in the databse, not on a feature flag.

Prototype design for before registering:
http://dm-prototype.herokuapp.com/states/account?dos.framework=open&dos.application=null

Screenshot before registering:
screen shot 2015-10-21 at 11 14 34

Prototype design for after registering: 
http://dm-prototype.herokuapp.com/states/account?dos.framework=open&dos.application=some

Screenshot after registering:
screen shot 2015-10-21 at 11 15 01

Depends on:
- [x] https://github.com/alphagov/digitalmarketplace-api/pull/272
- [x] https://github.com/alphagov/digitalmarketplace-utils/pull/190

### DIFF
--- a/app/assets/scss/_framework-application.scss
+++ b/app/assets/scss/_framework-application.scss
@@ -118,13 +118,24 @@
 
   .summary-item-heading {
     padding-top: 0;
+    padding-right: 0;
   }
 
+  input.button-save { 
+    margin-top: 0px;
+    margin-bottom: 8px;
+  }
+
+  a {
+    @include bold-24;
+  }
+ 
   p {
     margin-bottom: 5px;
 
     a {
       display: block;
+      @include core-19;
     }
   }
 }
--- a/app/main/helpers/frameworks.py
+++ b/app/main/helpers/frameworks.py
@@ -1,6 +1,5 @@
 from flask import abort
 from flask_login import current_user
-from dmutils.audit import AuditTypes
 from dmutils.apiclient import APIError
 import re
 from operator import add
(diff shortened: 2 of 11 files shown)


quis:
Maybe a bigger question than this pull request, but are we doing the thing where you can register interest before it's actually open? If so I think the "start" button is confusing—it was last time when we had a link that said something like "register" but there was nothing you could do beyond it.

TheDoubleK:
Good point - a question for Cath I think.  This messaging only appears if the framework is open, so at the moment it makes sense.  But if we want people to "register" in advance then this isn't the right thing.

quis:
@TheDoubleK Based on the discussion this morning it sounds like this is OK. We might want to think about having a separate "DOS is coming" message for suppliers who create accounts in advance.

quis:
There's quite a lot in here which isn't very generic, eg templates called `_dos_messages`, variables called `g7…` in the view. Do we care, or are we just trying to get this out?

quis:
I'd rather see the template look like
```
for framework in frameworks:
   if framework is pending:
       include generic-template-for-pending-frameworks.html
   if framework is open:
       include generic-template-for-open-frameworks.html
```

quis:
Based on real life discussion have decided that this is :+1: for now, and will do some work to make it less generic as part of [#106103994](https://www.pivotaltracker.com/story/show/).

PR #314, opened 2015-10-22
Lowercase lot updates

allait:
#### Show full lot name in service list tables
Instead of showing slug abbreviation display the full lot name
in service lists. This is a temporary solution before services
are split between different lot pages.

#### Get lot question directly from loader for G7 submission page
G7 draft creation requires the lot question form field to be rendered,
since lot was removed from the service summary sections in the last
framework update the question needs to be loaded separately.


app/main/views/services.py
```diff
@@ -214,7 +214,7 @@ def create_new_draft_service(framework_slug):
     lot = request.form.get('lot', None)
 
     if not lot:
-        question = content_loader.get_builder(framework_slug, 'edit_submission').get_question('lot')
+        question = content_loader.get_question(framework_slug, 'services', 'lot')
```
TheDoubleK: This breaks with the current API because it still has lots in the schema - trying to create a new draft gives, e.g.  `{'lot': "'paas' is not one of ['PaaS']"}`

TheDoubleK:
Looks good - the service listing page still works for both old and new API, so given that no-one is creating drafts this should be good to merge I think :+1:

app/main/views/services.py
allait: Yes, that's one of the two things that break (the other one being lot names disappear in the service tables). Both are pretty hard to avoid without doing multi-step API deploy and since they don't seem too critical in a short window I'm just going to try and deploy things close to each other.

tombye:
I'm getting a few template errors on the following pages:

- `/suppliers/frameworks/g-cloud-7/submissions/create` 
- `/suppliers/frameworks/g-cloud-7/services`

They're to do with the `framework_slug` argument not being passed to some routes by the `url_for` helper.

I found adding the argument to the following lines fixed it:

- https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/templates/frameworks/services.html#L139
- https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/templates/services/service_submission.html#L32

Think it was left out when we made the framework less hard-coded.

PR #315, opened 2015-10-22
Watch for changes to content

quis:
Implements and depends on:
- [x] https://github.com/alphagov/digitalmarketplace-utils/pull/193

Adds https://pypi.python.org/pypi/watchdog as a requirement to make watching for file changes use system events, which is faster

### DIFF
--- a/application.py
+++ b/application.py
@@ -3,7 +3,6 @@
 import os
 import re
 from app import create_app
-from flask.ext.script import Manager
 from dmutils import init_manager
 
 application = create_app(
@@ -15,8 +14,7 @@ application.jinja_options = {
     ]
 }
 
-manager = Manager(application)
-init_manager(manager, 5003, ['./app/content/frameworks'])
+init_manager(application, 5003, ['./app/content/frameworks'])
 
 if __name__ == '__main__':
     manager.run()
--- a/application.py
+++ b/application.py
@@ -3,7 +3,8 @@
 import os
 import re
 from app import create_app
-from flask.ext.script import Manager, Server
+from flask.ext.script import Manager
+from dmutils import init_manager
 
 application = create_app(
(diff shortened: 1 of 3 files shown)


tombye:
I might be missing something but where is watchdog used here? I can see it included as a dependancy but can't see the module imported anywhere.

Also, if it's to be used outside of tests it should probably be in `requirements.txt` which the others include.

tombye:
Ah, I see, the Werkzeug reloader will use it if it's installed:

http://werkzeug.pocoo.org/docs/0.10/serving/?highlight=run_simple#reloader

Another question: why remove `Flask-script` if we're still using `Manager` from it?

quis:
@tombye You're right about the dependency on `Flask-script`, although it does get imported by `dmutils` so nothing breaks. But I think https://github.com/alphagov/digitalmarketplace-utils/pull/195 cleans things up a bit.

quis:
# Have flask init init the manager

Depends on and implements alphagov/digitalmarketplace-utils#195

application.py
```diff
@@ -14,8 +14,7 @@
     ]
 }
 
-manager = Manager(application)
-manager.add_command("runserver", Server(port=5003))
+init_manager(application, 5003, ['./app/content/frameworks'])
```
allait: Should be `manager = init_manager(...)`, otherwise `manager.run()` won't work. But it looks like https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/flask_init.py#L80 doesn't return `manager` either, so we need to update utils first.

tombye:
I've added some updates to brings in the changes from alphagov/digitalmarketplace-utils#195 but with a version of `init_manager` what returns a `Manager` instance (7db91ec).

allait:
:+1:

PR #316, opened 2015-10-29
README fix: point to a page that exists

tommorris:
Similar to [this](https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/138) pull request for admin-frontend, and as per @allait's comment, it is a good idea to point new developers to a page that actually exists rather than 404s.


quis:
:+1:

PR #318, opened 2015-11-02
G7 display agreements

robyoung:
- Add an environment variable for the base assets URL since the assets domain is not just for documents.
- Build a path for an agreements document based on the framework slug and the supplier ID of the current user. The agreement documents contain the supplier ID (for easy referencing later) and also live in a directory named the supplier ID (for easy browsing of the bucket).
- Add a function to get a document URL that works in a similar way to the draft document URL function but only replaces the host and scheme with the asset URL if the environment variable is set. This allows the URLs to work in development even if a reverse proxy is not being used.
- Add link to framework agreement page. This replaces the declaration status API call on the framework dashboard with a call to the supplier framework info endpoint that includes both the declaration status and the flags for whether a supplier is on the framework and they have returned their signed agreement.

## Depends on
- [x] Sign off by Cath
- [x] Sign off by Ralph
- [ ] :sparkles: functional tests :sparkles: 
- [x] https://github.com/alphagov/digitalmarketplace-aws/pull/159
- [x] https://github.com/alphagov/digitalmarketplace-utils/pull/201

### DIFF
--- a/app/main/helpers/frameworks.py
+++ b/app/main/helpers/frameworks.py
@@ -17,6 +17,10 @@ OPTIONAL_FIELDS = set([
 ])
 
 
+def get_agreement_document_path(framework_slug, supplier_id, document_name):
+    return '{0}/agreements/{1}/{1}-{2}'.format(framework_slug, supplier_id, document_name)
+
+
 def frameworks_by_slug(client):
     framework_list = client.find_frameworks().get("frameworks")
     frameworks = {}
--- a/app/main/helpers/frameworks.py
+++ b/app/main/helpers/frameworks.py
@@ -194,6 +194,32 @@ def get_declaration_status(data_api_client, framework_slug):
         return declaration.get('status', 'unstarted')
 
 
+def get_supplier_framework_info(data_api_client, framework_slug):
+    try:
+        return data_api_client.get_supplier_framework_info(
+            current_user.supplier_id, framework_slug
+        )['frameworkInterest']
+    except APIError as e:
+        if e.status_code == 404:
+            return None
+        else:
+            abort(e.status_code)
+
+
+def get_declaration_status_from_info(supplier_framework_info):
+    if supplier_framework_info is None:
+        return 'unstarted'
+    else:
+        return supplier_framework_info['declaration'].get('status', 'unstarted')
+
+
+def get_supplier_on_framework_from_info(supplier_framework_info):
+    if supplier_framework_info is None:
+        return False
+    else:
+        return bool(supplier_framework_info.get('onFramework'))
+
+
 def g_cloud_7_is_open_or_404(data_api_client):
     if data_api_client.get_framework_status('g-cloud-7').get('status', None) != 'open':
         abort(404)
--- a/app/main/helpers/frameworks.py
+++ b/app/main/helpers/frameworks.py
@@ -17,10 +17,6 @@ OPTIONAL_FIELDS = set([
 ])
 
 
-def get_agreement_document_path(framework_slug, supplier_id, document_name):
-    return '{0}/agreements/{1}/{1}-{2}'.format(framework_slug, supplier_id, document_name)
-
-
 def frameworks_by_slug(client):
     framework_list = client.find_frameworks().get("frameworks")
     frameworks = {}
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -18,10 +18,10 @@ from ...main import main, content_loader
 from ..helpers import hash_email
 from ..helpers.frameworks import get_error_messages_for_page, get_first_question_index, \
     get_error_messages, get_declaration_status, get_last_modified_from_first_matching_file, \
-    register_interest_in_framework
+    register_interest_in_framework, get_agreement_document_path
 from ..helpers.services import (
     get_draft_document_url, get_service_attributes, get_drafts,
-    count_unanswered_questions
+    count_unanswered_questions, get_document_url
 )
 
 
@@ -70,7 +70,6 @@ def framework_dashboard(framework_slug):
             'supplier_pack': get_last_modified_from_first_matching_file(key_list, 'g-cloud-7-supplier-pack.zip'),
             'supplier_updates': get_last_modified_from_first_matching_file(key_list, 'g-cloud-7-updates/')
         },
-
         **template_data
     ), 200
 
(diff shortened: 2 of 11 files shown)


TheDoubleK:
I get an error running this locally when I hit the link to "Download your application result letter", i.e. at `/suppliers/frameworks/g-cloud-7/agreements/result-letter.pdf` - would be nice if that didn't die if the file doesn't exist in S3.

robyoung:
If the file doesn't exist in S3 it will fail with a 404. However, the S3 bucket isn't configured by default so it needs to be provided by an environment variable. I thought about pre-connecting to all buckets on application startup but that would mean we'd need to have them configured for development. I'm not sure that's a bad idea.

TheDoubleK:
I think `/suppliers/frameworks/g-cloud-7/agreement` still needs a bit of CSS massaging - the spacing doesn't look right here:
screen shot 2015-11-03 at 12 13 37

and the bottom horizontal line shouldn't be here:
screen shot 2015-11-03 at 12 11 07

app/templates/frameworks/agreement.html
```diff
@@ -0,0 +1,132 @@
...
+                    },
+                ]
+            %}
+            {% include "toolkit/documents.html" %}
+            {% endwith %}
+        </div>
+        
+        <div class="padding-bottom framework-section">
+            <h3>2. Read your framework agreement</h3>
+            <p>If you have a question about your framework agreement, <a href="#">contact CCS</a>.</p>
```
TheDoubleK: It would be nice to find a URL/email address for this link now, as it might easily be forgotten if it goes in empty now.

app/main/helpers/frameworks.py
```diff
@@ -190,6 +190,32 @@ def get_declaration_status(data_api_client, framework_slug):
...
+        return 'unstarted'
+    else:
+        return supplier_framework_info['declaration'].get('status', 'unstarted')
+
+
+def get_supplier_on_framework_from_info(supplier_framework_info):
+    if supplier_framework_info is None:
+        return False
+    else:
+        return bool(supplier_framework_info.get('onFramework'))
```
TheDoubleK: I don't think you need the cast to bool here - getting a boolean value from the JSON should just work.  At least, it works here dereferencing the value directly rather than calling `get()`: https://github.com/alphagov/digitalmarketplace-scripts/blob/fix-lots-in-export/dmscripts/generate_framework_agreement_data.py#L62

app/main/views/frameworks.py
```diff
@@ -33,8 +35,9 @@
 @flask_featureflags.is_active_feature('GCLOUD7_OPEN')
 def framework_dashboard(framework_slug):
     template_data = main.config['BASE_TEMPLATE_DATA']
-    # TODO add a test for 404 if framework doesn't exist
     framework = data_api_client.get_framework(framework_slug)['frameworks']
+    if not framework:
```
TheDoubleK: Is this check necessary here?  If the framework doesn't exist then I don't think this line will ever be reached - I think either it will 404 from an HTTPError from the API client, or there will be a KeyError trying to dereference 'frameworks' if nothing was returned?

But I might be wrong.

app/main/views/frameworks.py
```diff
@@ -363,3 +382,32 @@ def framework_updates_email_clarification_question(framework_slug):
...
+        framework=framework,
+        supplier_framework=supplier_framework,
+        **template_data
+    ), 200
+
+
+@main.route('/frameworks/<framework_slug>/agreement', methods=['POST'])
+@login_required
+def upload_framework_agreement(framework_slug):
+    pass
```
TheDoubleK: This could return a `501: Not Implemented` response rather than break.

But it doesn't really matter as I know it's going to be filled in already anyway.

app/templates/frameworks/agreement.html
```diff
@@ -0,0 +1,132 @@
+{% extends "_base_page.html" %}
+
+{% block page_title %}{{ framework.name }} Framework Agreement– Digital Marketplace{% endblock %}
```
TheDoubleK: Should be a space before the dash here.

app/templates/frameworks/agreement.html
```diff
@@ -0,0 +1,132 @@
...
+        <div class="padding-bottom framework-section">
+            <h3>2. Read your framework agreement</h3>
+            <p>If you have a question about your framework agreement, <a href="#">contact CCS</a>.</p>
+        </div>
+
+        <div class="padding-bottom framework-section">
+            <h3>3. Sign your framework agreement</h3>
+            <p class='padding-bottom-small'>To digitally sign your framework agreement, you need to:</p>
+            <ol class='number-list indent-list-large loose-list padding-bottom-small'>
+                <li>Use <strong>Adobe Reader</strong> to open your G-Cloud 7 framework agreement.
```
TheDoubleK: Should be `{{ framework.name }}` instead of G-Cloud 7 here.

app/templates/frameworks/agreement.html
```diff
@@ -0,0 +1,132 @@
...
+                with
+                type = "save",
+                label = "Send document to CCS"
+                %}
+                {% include "toolkit/button.html" %}
+                {% endwith %}
+            </form>
+        </div>
+        
+        <a href="{{url_for('.framework_dashboard', framework_slug=framework.slug)}}">Return to your G-Cloud application</a>
```
TheDoubleK: Should also be `{{ framework.name }}` instead of G-Cloud here.

app/templates/frameworks/dashboard.html
```diff
@@ -56,9 +56,46 @@ <h2 class="summary-item-heading">{{ framework.name }} is closed for applications
...
     <ul class="browse-list">
+
+      {% if framework.status == 'standstill' and application_made %}
+        <li class="browse-list-item framework-application-section">
+          <div class="grid-row">
+            <div class="column-two-thirds">
+              <a class="browse-list-item-link" href="{{ url_for('.download_agreement_file', framework_slug=framework.slug, document_name='result-letter.pdf') }}">
+                <span>Download your application result letter</span>
+              </a>
+              <p>This letter informs you if your {{ framework.name }} has been successful.</p>
```
TheDoubleK: Missing the word `application` after `{{ framework.name }}` in this sentence I think.

tests/app/main/test_frameworks.py
```diff
@@ -260,6 +255,222 @@ def test_last_updated_does_not_exist(self, data_api_client, s3):
 
             self._assert_last_updated_times(doc, last_updateds)
 
+    def test_returns_404_if_framework_does_not_exist(self, data_api_client, s3):
+        with self.app.test_client():
+            self.login()
+            data_api_client.get_framework.return_value = {'frameworks': {}}
```
TheDoubleK: Not sure that this is the response you'd get from the API for a non-existent framework - I think it will 404.

app/main/helpers/frameworks.py
robyoung: I've cast to bool here to make it clear that this function is expected to be used as a boolean function.

app/main/views/frameworks.py
robyoung: This gets cleaned up in the next PR.

TheDoubleK:
:+1:

PR #321, opened 2015-11-02
Add declaration content for Digital Outcomes and Specialists

quis:
## Update dependencies and frameworks content
- frameworks and utils to handle the content
- latest Node SASS to work on newest Node (for Heroku)

### Smarter question numbering and cross referencing, which
- implements alphagov/digitalmarketplace-utils#186
- adds a template filter for looking up questions

### Add section descriptions
- this should mean that we don’t have to publish separate guidance

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -9,8 +9,6 @@ from dmutils.user import User
 
 from config import configs
 
-from markdown import markdown
-
 
 data_api_client = apiclient.DataAPIClient()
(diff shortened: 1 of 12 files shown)


robyoung:
That's a pretty complex filter flow, it would be good to have at least one test in there.

quis:
@robyoung Fair point, added tests in 0e9378a

robyoung:
:+1:

PR #322, opened 2015-11-02
Basic DOS service flow

allait:
Requires:
- [x] https://github.com/alphagov/digitalmarketplace-api/pull/281
- [x] https://github.com/alphagov/digitalmarketplace-frameworks/pull/110

### Add service submission page with a list of framework lots
Splits framework services into two pages: list of lots and individual
lot pages with lot services.

Both pages are only displayed if framework status is either open or
pending. For pending frameworks page is only displayed to suppliers
who have submitted an application. Feature flag check is removed.

Lot pages check the 'one service limit' flag and:

* Display the list of lot services for lots with multiple services.
  This is similar to the previous submission services page, except
  the services are filtered by lot.

* For one service lots: creates the service for given lot and supplier
  if it doesn't exist and redirects to the service page right away.

Note: creating an empty draft for one-service lots is an action taken
as a result of a GET request (which are not covered by the CSRF
protection), but this is the only approach we have given the current
design of the lots page.

### Rewrite draft create page to use the first manifest section
Since draft lot is set as part of the URL the first service submission
page can be different for each lot.

The draft creation endpoint will look up the first editable section for
the given lot and use the submitted form data to create the draft service,
similar to how draft updates work at the moment: data is sent to the API
with a list of page questions to validate required answers, any API errors
returned are displayed on the form page, otherwise the draft is created
and user is redirected to the new draft page.

`create_new_draft_service` calls are updated with the breaking changes
in dmutils 12.0.0.

### Add lot slug to service submission URLs
Moves all draft-related endpoints to new URLs that include lot slugs
and updates the tests with lowercase lots, lot names and new URLs.

On redirects and in template links lot_slug is taken from service
data, so while it's not validated right now in most endpoints, if
the initial lot slug doesn't match the service the redirect or link
click will route to the correct lot URL.

#### Remove remaining uses of get_framework_status
`get_framework_status` was replaced with `get_framework` in most
calls, the only remaining instances were in unused helper function
and test mocks that were patching the wrong method.

With this change `get_framework_status` is no longer used by the
supplier frontend and both dmutils.apiclient method and the API
endpoint could be removed.

#### Replace GCLOUD7_OPEN with framework status checks
Adds `get_framework` helper to get framework data from the API and
check that status is 'open' if called with `open_only=True` or
either 'open' or 'pending' if `open_only=False`.

Also adds lot_slug checks by looking up the lot slug in framework
lots.

If framework is in a different state or lot doesn't exist returns
404.

Document redirect URLs are currently not covered by framework checks.

#### Don't filter services without serviceName
With removal of the lot selection form draft creation starts with
the first editable section, which should be serviceName for g-cloud
and listed services, so there's no need to filter drafts without a
service name.

#### Use lot name fallback for services without serviceName
One-per-lot services don't have a serviceName field, so we use
lot name instead.

### DIFF
--- a/app/main/helpers/frameworks.py
+++ b/app/main/helpers/frameworks.py
@@ -216,11 +216,6 @@ def get_supplier_on_framework_from_info(supplier_framework_info):
         return bool(supplier_framework_info.get('onFramework'))
 
 
-def g_cloud_7_is_open_or_404(data_api_client):
-    if data_api_client.get_framework_status('g-cloud-7').get('status', None) != 'open':
-        abort(404)
-
-
 def question_references(data, get_question):
     return re.sub(
         r"\[\[([^\]]+)\]\]",  # anything that looks like [[nameOfQuestion]]
--- a/app/main/helpers/frameworks.py
+++ b/app/main/helpers/frameworks.py
@@ -19,7 +19,7 @@ OPTIONAL_FIELDS = set([
 
 def get_framework(client, framework_slug, open_only=True):
     framework = client.get_framework(framework_slug)['frameworks']
-    allowed_statuses = ['open'] if open_only else ['open', 'pending']
+    allowed_statuses = ['open'] if open_only else ['open', 'pending', 'standstill']
     if framework['status'] not in allowed_statuses:
         abort(404)
 
--- a/app/main/helpers/frameworks.py
+++ b/app/main/helpers/frameworks.py
@@ -17,6 +17,20 @@ OPTIONAL_FIELDS = set([
 ])
 
 
+def get_framework(client, framework_slug, open_only=True):
+    framework = client.get_framework(framework_slug)['frameworks']
+    allowed_statuses = ['open'] if open_only else ['open', 'pending']
+    if framework['status'] not in allowed_statuses:
+        abort(404)
+
+    return framework
+
+
+def get_framework_and_lot(client, framework_slug, lot_slug, open_only=True):
+    framework = get_framework(client, framework_slug, open_only)
+    return framework, get_framework_lot(framework, lot_slug)
+
+
 def frameworks_by_slug(client):
     framework_list = client.find_frameworks().get("frameworks")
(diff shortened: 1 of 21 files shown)


robyoung:
The copy on the page that lists the lots doesn't match the prototype. Has it been checked with Roz and Ralph?

robyoung:
This should have been a comment on the API pull request but I only just hit it. There is no migration (or automated method) adding the Digital Outcomes and Specialists lots.

robyoung:
screen shot 2015-11-04 at 17 15 28
I think these questions need to be passed through the markdown filter.

robyoung:
On the Digital outcomes supplier information page this question appears twice and only one can be selected.
screen shot 2015-11-04 at 17 18 23

robyoung:
Not sure if it's part of this story or part of the pricing one but prices don't work for digital outcomes. They do not get registered as entered on the overview page and the same price is pre-filled for all capabilities.

robyoung:
The flash from deleting a digital outcome was not shown to me until entered user research studios. Also, I think the lot name should be used at the title rather than the framework name. I guess this would have an impact on G-Cloud 7.
screen shot 2015-11-04 at 17 38 46

robyoung:
The questions for the user research studio lot look like specialist roles.

allait:
@robyoung fixed the flash message and the services page heading.

> The copy on the page that lists the lots doesn't match the prototype. Has it been checked with Roz and Ralph?

Lot descriptions are loaded from the frameworks lot question, so they'll change whenever content changes. I've copied whatever was in the prototype at the moment.

> I think these questions need to be passed through the markdown filter to process the links.

This could be a separate story if the final content ends up with markdown bits in question names.

> On the Digital outcomes supplier information page this question appears twice and only one can be selected.

This sounds like an issue in the manifest, it should be resolved when we make a final questions list.

robyoung:
:+1:

PR #323, opened 2015-11-03
Remove markdown package and filter definition.

pcraig3:
As part of [an earlier pull request](https://github.com/alphagov/digitalmarketplace-utils/pull/192), a markdown filter was added to flask apps that import dmutils.

Means we can still use the `markdown` filter in our jinja2 templates, but we shouldn't be duplicating it in the `create_app` function.

*Might* have forgotten to do this earlier.


quis:
:+1:

PR #325, opened 2015-11-04
Refactored framework validation to make it modular

tommorris:
Previously framework validation rules were specifically for G-Cloud 7. To support DOS, we need to have the ability to have multiple validators. This lays the groundwork for this by separating out G7 framework rules into one class.

### DIFF
--- a/app/main/helpers/validation.py
+++ b/app/main/helpers/validation.py
@@ -1,0 +1,169 @@
+from functools import reduce
+from operator import add
+import re
+import six
+from werkzeug.datastructures import ImmutableOrderedMultiDict
+
+EMAIL_REGEX = r'^[^@^\s]+@[^@^\.^\s]+(\.[^@^\.^\s]+)+$'
+
+
+def get_validator(framework, content, answers):
+    """
+    Retrieves a validator by slug contained in the framework dictionary.
+    """
+    if framework is None:
+        raise ValueError("a framework dictionary must be provided")
+    if framework is not None:
+        validator_cls = VALIDATORS.get(framework['slug'])
+        return validator_cls(content, answers)
+
+
+class DeclarationValidator(object):
+    email_validation_fields = []
+    character_limit = None
+    optional_fields = set([])
+
+    def __init__(self, content, answers):
+        self.content = content
+        self.answers = answers
+
+    def get_error_messages_for_page(self, section):
+        all_errors = self.get_error_messages()
+        page_ids = section.get_question_ids()
+        page_errors = ImmutableOrderedMultiDict(filter(lambda err: err[0] in page_ids, all_errors))
+        return page_errors
+
+    def get_error_messages(self):
+        raw_errors_map = self.errors()
+        errors_map = list()
+        for question_number, question_id in enumerate(self.all_fields()):
+            if question_id in raw_errors_map:
+                validation_message = self.get_error_message(question_id, raw_errors_map[question_id])
+                errors_map.append((question_id, {
+                    'input_name': question_id,
+                    'question': "Question {}".format(question_number + 1),
+                    'message': validation_message,
+                }))
+
+        return errors_map
+
+    def get_error_message(self, question_id, message_key):
+        for validation in self.content.get_question(question_id).get('validations', []):
+            if validation['name'] == message_key:
+                return validation['message']
+        default_messages = {
+            'answer_required': 'Answer required'
+        }
+        return default_messages.get(
+            message_key, 'There was a problem with the answer to this question')
+
+    def errors(self):
+        raise NotImplementedError("only a subclass should be used")
+
+    def all_fields(self):
+        return reduce(add, (section.get_question_ids() for section in self.content))
+
+    def fields_with_values(self):
+        # empty list types will not be included so only string types need to be considered
+        return set(key for key, value in self.answers.items()
+                   if not isinstance(value, six.string_types) or len(value) > 0)
+
+    def errors(self):
+        errors_map = {}
+        errors_map.update(self.character_limit_errors())
+        errors_map.update(self.formatting_errors(self.answers))
+        errors_map.update(self.answer_required_errors())
+        return errors_map
+
+    def answer_required_errors(self):
+        req_fields = self.get_required_fields()
+        filled_fields = self.fields_with_values()
+        errors_map = {}
+
+        for field in req_fields - filled_fields:
+            errors_map[field] = 'answer_required'
+
+        return errors_map
+
+    def character_limit_errors(self):
+        errors_map = {}
+        for question_id in self.all_fields():
+            if self.content.get_question(question_id).get('type') in ['text', 'textbox_large']:
+                answer_length = len(self.answers.get(question_id, ""))
+                if self.character_limit is not None and answer_length > self.character_limit:
+                    errors_map[question_id] = "under_character_limit"
+
+        return errors_map
+
+    def formatting_errors(self, answers):
+        errors_map = {}
+        if self.email_validation_fields is not None and len(self.email_validation_fields) > 0:
+            for field in self.email_validation_fields:
+                if not re.match(EMAIL_REGEX, self.answers.get(field, '')):
+                    errors_map[field] = 'invalid_format'
+        return errors_map
+
+    def get_required_fields(self):
+        try:
+            req_fields = self.required_fields
+        except AttributeError:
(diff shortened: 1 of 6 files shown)


allait:
I don't think we really need nested packages in `helpers`, we could just move validators to `helpers/validation.py` and keep the `helpers/frameworks.py` instead of moving code to `__init__.py`.

app/main/helpers/frameworks/__init__.py
```diff
@@ -0,0 +1,70 @@
+import re
+from flask import abort
+from flask_login import current_user
+from dmutils.apiclient import APIError
+from werkzeug.datastructures import ImmutableOrderedMultiDict
+
+from app.main.helpers.frameworks.validation import G7Validator
```
allait: `from .validation ...`

quis:
I’d squash those commits together. Don’t think they add anything individually.

app/main/helpers/frameworks/validation.py
```diff
@@ -0,0 +1,178 @@
...
+            required_fields.add('SQ3-1k')
+
+        # If you answered No to question 26 (established in the UK)
+        if 'SQ5-2a' in self.answers and not self.answers['SQ5-2a']:
+            required_fields.add('SQ1-1i-i')
+            required_fields.add('SQ1-1j-i')
+
+        return required_fields
+
+validators['g-cloud-7'] = G7Validator
```
allait: `VALIDATORS = { 'g-cloud-7': G7Validator }` should make it clear that it's a constant and shouldn't be modified.

app/main/helpers/frameworks/validation.py
```diff
@@ -0,0 +1,178 @@
...
+
+def get_validator(framework):
+    """
+    Retrieves a validator by name, otherwise returns EmptyValidator.
+    """
+    if framework is not None:
+        framework_name = framework.get('slug', None)
+        return validators.get(framework_name, EmptyValidator)
+    else:
+        return EmptyValidator
```
allait: This silences a possible error, I don't think we need this

app/main/helpers/frameworks/validation.py
```diff
@@ -0,0 +1,178 @@
...
+    "SQ1-1q-i", "SQ1-1q-ii", "SQ1-1q-iii", "SQ1-1q-iv", "SQ1-1cii", "SQ1-1i-ii",
+    "SQ1-1j-i", "SQ1-1j-ii", "SQ4-1c", "SQ3-1k", "SQ1-1i-i"
+])
+
+G7_EMAIL_VALIDATION_FIELDS = set(['SQ1-1o', 'SQ1-2b'])
+
+validators = {}
+
+
+def get_validator(framework):
```
allait: This could take `content` and `answers` right away and return an instance instead of a class, similar to how a factory would normally behave

app/main/helpers/frameworks/validation.py
```diff
@@ -0,0 +1,178 @@
...
+    Retrieves a validator by name, otherwise returns EmptyValidator.
+    """
+    if framework is not None:
+        framework_name = framework.get('slug', None)
+        return validators.get(framework_name, EmptyValidator)
+    else:
+        return EmptyValidator
+
+
+class FrameworkValidator(object):
```
allait: This is only used for framework declarations, so `FrameworkValidator` is a bit ambiguous.`FrameworkDeclarationValidator` or just `DeclarationValidator`?

app/main/helpers/frameworks/validation.py
```diff
@@ -0,0 +1,178 @@
...
+    """
+    pass
+
+
+class G7Validator(FrameworkValidator):
+    """
+    Validator for G-Cloud 7.
+    """
+    optional_fields = G7_OPTIONAL_FIELDS
+    email_validation_fields = G7_EMAIL_VALIDATION_FIELDS
```
allait: I'd inline the constants here, the module is no longer about G7, but this class is, so it seems like a good place to store those values.

app/main/helpers/frameworks/validation.py
```diff
@@ -0,0 +1,178 @@
...
+            if validation['name'] == message_key:
+                return validation['message']
+        default_messages = {
+            'answer_required': 'Answer required'
+        }
+        return default_messages.get(
+            message_key, 'There was a problem with the answer to this question')
+
+    def errors(self):
+        return {}
```
allait: Since it's meant to be overriden, `errors` should `raise NotImplementedError` to avoid the possibility of accidentally calling a class that doesn't actually do any validation.

app/main/helpers/frameworks/validation.py
```diff
@@ -0,0 +1,178 @@
...
+    character_limit = G7_CHARACTER_LIMIT
+
+    def errors(self):
+        errors_map = {}
+        errors_map.update(self.character_limit_errors())
+        errors_map.update(self.formatting_errors(self.answers))
+        errors_map.update(self.answer_required_errors())
+        return errors_map
+
+    def answer_required_errors(self):
```
allait: I think this could be easily moved to the parent class (were the rest of the `_errors` methods are defined) if we make `required_fields` a property.

tommorris:
@allait I've made changes to reflect all those comments and refactored as @quis suggested.

app/main/helpers/validation.py
```diff
@@ -0,0 +1,185 @@
...
+    character_limit = 5000
+
+    def errors(self):
+        errors_map = {}
+        errors_map.update(self.character_limit_errors())
+        errors_map.update(self.formatting_errors(self.answers))
+        errors_map.update(self.answer_required_errors())
+        return errors_map
+
+    def answer_required_errors(self):
```
allait: We can move this method to `DeclarationValidator` now

app/main/helpers/validation.py
```diff
@@ -0,0 +1,185 @@
...
+            req_fields = set(self.all_fields())
+
+        #  Remove optional fields
+        if self.optional_fields is not None:
+            req_fields -= set(self.optional_fields)
+
+        return req_fields
+
+
+class EmptyValidator(DeclarationValidator):
```
allait: Don't need this anymore

app/main/helpers/validation.py
```diff
@@ -0,0 +1,185 @@
+from functools import reduce
+from operator import add
+import re
+import six
+from werkzeug.datastructures import ImmutableOrderedMultiDict
+
+EMAIL_REGEX = r'^[^@^\s]+@[^@^\.^\s]+(\.[^@^\.^\s]+)+$'
+
+VALIDATORS = {}
```
allait: I still think we should assign VALIDATORS only once at the end of the file

app/main/helpers/validation.py
tommorris: @allait: The G7 Validator calls back to the super implementation to retrieve the required fields.

app/main/helpers/validation.py
tommorris: @allait I'm not sure we can. `DeclarationValidator`'s `errors` method throws `NotImplementedError`. If we move `errors` from `G7Validator`, then we lose that behaviour.

app/main/helpers/validation.py
tommorris: @allait: We could. The only thing I'm concerned about with that is that the registration of the slug-to-class mapping would not be done directly next to the class. If we had, say, 5 or 10 frameworks, it wouldn't be intuitively obvious where the mapping is done.

app/main/helpers/validation.py
allait: I'm proposing to move `answer_required_errors`, which wouldn't be called in the parent class since it's only called in `errors`

app/main/helpers/validation.py
allait: I'm not sure I understand what you mean. We don't need `EmptyValidator` class since we're no longer returning it from `get_validator`.

app/main/helpers/validation.py
allait: @tommorris writing a special case class for each framework is a temporary solution, we don't want to be doing this for future frameworks, so hopefully we won't end up with 5 or more classes. At any rate, I'd rather have a single place which listed all available validators than have it spread out throughout the file.

tommorris:
@allait: rebased and fixed the issues raised.

allait:
:+1:

PR #326, opened 2015-11-05
BUGFIX: Supplier framework with null declaration

robyoung:
When a supplier first registers interest in a framework the declaration
is null. This was causing the framework dashboard to blow up because we
were expecting the declaration to be a dictionary. This handles the case
when it is null.


quis:
:+1:

PR #328, opened 2015-11-06
Tidy up various things

quis:
## Services page was missing a heading…

…because it wasn’t considering the 'standstill' state.

This commit:
- fixes that
- changes the 'empty' message to be in the correct tense for each state

image

## Fix typo


TheDoubleK:
:+1:

PR #329, opened 2015-11-06
Re-add final instruction to upload signed document

TheDoubleK:
With words from Ralph.

screen shot 2015-11-06 at 16 31 08


quis:
:+1:

PR #331, opened 2015-11-09
Pass framework name to email template

TheDoubleK:
The template was previously getting a framework rather than the framework name.

This also removes confusing messaging  "0 draft services were not submitted" for suppliers who did not leave any services in draft.

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -481,7 +481,7 @@ def upload_framework_agreement(framework_slug):
     try:
         email_body = render_template(
             'emails/framework_agreement_uploaded.html',
-            framework=framework,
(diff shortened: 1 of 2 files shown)


app/main/views/frameworks.py
```diff
@@ -481,7 +481,7 @@ def upload_framework_agreement(framework_slug):
     try:
         email_body = render_template(
             'emails/framework_agreement_uploaded.html',
-            framework=framework,
+            framework_name=framework.get('name', 'Unknown framework'),
```
robyoung: I think this should probably be `framework['name']` if it doesn't have a name this should fail completely.

TheDoubleK: Harsh but fair I guess.

robyoung:
:+1:

PR #332, opened 2015-11-09
Add lot counts to submitted lots page

quis:
This implements: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/178

For each lot, it tells the user how many draft and complete services they have submitted.

Right now it’s fairly specific to the G-Cloud 7 sitation (without actually having `g-cloud-7` anywhere). We’ll need to handle some different states when DOS comes along.

--

image


allait:
:+1:

PR #333, opened 2015-11-09
Use the declaration supplier name for agreements

robyoung:
The files for the framework agreements need to use the registered
company name from the declaration.


TheDoubleK:
:+1:

PR #334, opened 2015-11-09
Update content.

pcraig3:
Update [digitalmarketplace-frameworks](https://github.com/alphagov/digitalmarketplace-frameworks) content, and update the supplier app so it's consistent with the new changes.

Specifically:

- Change `maxLengthInWords` to `max_length_in_words` in the form field macro
- Remove the `2answers-type2` macro which has been removed from the content

Doesn't really depend on: 
- [ ] https://github.com/alphagov/digitalmarketplace-frameworks/pull/113

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -7,6 +7,6 @@
     "jquery-details": "https://github.com/mathiasbynens/jquery-details/archive/v0.1.0.tar.gz",
     "digitalmarketplace-frontend-toolkit": "git://github.com/alphagov/digitalmarketplace-frontend-toolkit#v11.4.3",
     "govuk_template": "https://github.com/alphagov/govuk_template/releases/download/v0.14.1/jinja_govuk_template-0.14.1.tgz",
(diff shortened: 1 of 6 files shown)


bower.json
```diff
@@ -7,6 +7,6 @@
     "jquery-details": "https://github.com/mathiasbynens/jquery-details/archive/v0.1.0.tar.gz",
     "digitalmarketplace-frontend-toolkit": "git://github.com/alphagov/digitalmarketplace-frontend-toolkit#v11.4.3",
     "govuk_template": "https://github.com/alphagov/govuk_template/releases/download/v0.14.1/jinja_govuk_template-0.14.1.tgz",
-    "digitalmarketplace-frameworks": "git://github.com/alphagov/digitalmarketplace-frameworks#38fbfa0"
+    "digitalmarketplace-frameworks": "git://github.com/alphagov/digitalmarketplace-frameworks#895b2a6"
```
quis: This dependency should be updated with a version, not a SHA from now on.

PR #335, opened 2015-11-09
Update content.

pcraig3:
Update [digitalmarketplace-frameworks](https://github.com/alphagov/digitalmarketplace-frameworks) content, and update the supplier app so it's consistent with the new changes.

Specifically:

- Change `maxLengthInWords` to `max_length_in_words` in the form field macro
- Remove the `2answers-type2` macro which has been removed from the content

Depends on:
- [x] https://github.com/alphagov/digitalmarketplace-frameworks/pull/113


quis:
:+1:

PR #338, opened 2015-11-17
Fix supplier declaration completion

robyoung:
If the declaration validator is just run over the data that is submitted on the current form then it will never be submitted. When the current submitted data is merged with the existing answers it must be revalidated to determine whether the submission is complete or not.


quis:
It’s pretty confusing that `answers` is more up to date than `latest_answers`. Variables could be named better?

tests/app/main/helpers/validation/test_dos_declaration.py
```diff
@@ -0,0 +1,269 @@
...
+    "conspiracyCorruptionBribery": True,
+    "publishContracts": True,
+    "unfairCompetition": True,
+    "organisationSize": "micro",
+    "safeguardPersonalData": True,
+    "appropriateTradeRegistersNumber": "Blah",
+    "GAAR": False,
+    "contactEmailContractNotice": "Blah@example.com",
+    "status": "complete",
+    "sufficientInformation": True,
```
TheDoubleK: `sufficientInformation` has gone from content now.

TheDoubleK:
If you can make Travis happy then this broken version looks fine to me.

TheDoubleK:
:+1:

PR #339, opened 2015-11-17
Fix bug with display of current services

quis:
There is an edge case where:
- a supplier has services
- a supplier is not 'on' a framework

In this case the link to view services should not be shown.

In actuality, the link was being shown.

This is because the logic was checking for different things when deciding to
- show the list of current frameworks and service counts
- showing the view link

This commit changes the code to check for the same thing (`frameworks.live`) for both purposes.


robyoung:
:+1:

PR #340, opened 2015-11-17
Fixing agreement uploads

robyoung:
## Set extension name on signed agreement upload
Set the file extension correctly when uploading signed framework agreements.

## Verify that signed agreements are not empty
We have been getting some zero byte files uploaded as signed agreement documents.

## Depends on
- [x] https://github.com/alphagov/digitalmarketplace-utils/pull/207


TheDoubleK:
Reviewed - looks good to go once utils is merged, Travis happy, etc.

PR #341, opened 2015-11-17
Fix `return to summary` link while editing services

pcraig3:
At some point, our `return to service summary` links stopped displaying
properly.  Got them back and added a test.

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/107279234)

***

### Whoops

screen shot 2015-11-17 at 15 33 06

### Fixed

screen shot 2015-11-17 at 15 32 46


quis:
I had noticed this but decided not to fix it because the feature isn’t live. But now you have. So :tada: :+1:.

PR #343, opened 2015-11-18
Add message to supplier dashboard when framework is coming

quis:
image

### DIFF
--- a/app/templates/suppliers/_frameworks_coming.html
+++ b/app/templates/suppliers/_frameworks_coming.html
@@ -1,0 +1,14 @@
+{% for framework in frameworks.coming %}
+  {% if framework.slug == 'digital-outcomes-and-specialists' %}
+    {%
+      with
+      messages = [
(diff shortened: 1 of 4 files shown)


app/templates/suppliers/_frameworks_coming.html
```diff
@@ -0,0 +1,12 @@
+{% for framework in frameworks.coming %}
+  {%
+    with
+    messages = [
+      "We’ll email you when you can apply to Digital Outcomes and Specialists.",
```
TheDoubleK: This should either be generic, e.g. `{{ framework.name }}` instead of  `Digital Outcomes and Specialists` - and similarly the URL should use `{{ framework.slug }}, or if it is DOS-only the conditional should be something more like `if DOS in frameworks.coming ...`

quis:
@TheDoubleK Have added a check for DOS in the template and updated the tests.

pcraig3:
Looks good :+1:

TheDoubleK:
Yep - do it :+1:

PR #346, opened 2015-11-23
Show status of application by lot

quis:
Shows the status of a suppliers application to a framework on the page which lists the lots.

--

image

--

_Notes for reviewing:_
- _You won’t be able to follow the logic in the code—I’d be more interested to know if you think there are any obvious test cases I’ve missed_

--

This should work for existing G7 applications, future DOS applications and any other framework that fits our current model.

The exact messaging depends on:
- whether the lot has a limit of one service
- status of framework
- number of draft and complete services
- status of supplier declaration
- name of the lot

…which makes it quite complicated.

This commit adds a couple of tests to the existing tests for this endpoint. It then adds a _lot_ of explicit tests for the helper functions that generate the status messages. This is the only way I could think to do this, because the code itself is not elegant.

The wording itself has been reviewed by Sarah, Sue and Ralph, and Sue has done some guerilla testing of it.

### DIFF
--- a/app/main/helpers/frameworks.py
+++ b/app/main/helpers/frameworks.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 import re
 
 from flask import abort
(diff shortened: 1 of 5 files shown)


app/main/helpers/frameworks.py
```diff
@@ -122,3 +123,127 @@ def count_drafts_by_lot(drafts, lot):
...
+    framework_status,
+    lot_name,
+    unit,
+    unit_plural
+):
+
+    if not drafts_count and not complete_drafts_count:
+        return []
+
+    framework_is_open = 'open' == framework_status
```
TheDoubleK: This kind of thing is where Java-style brackets win for clarity.  Don't know if PEP8 will even allow them though.

quis: PEP8 will permit it. 0f587d8.

app/main/helpers/frameworks.py
```diff
@@ -122,3 +123,127 @@ def count_drafts_by_lot(drafts, lot):
...
+            } if declaration_complete else {
+                'title': u'Marked as complete'
+            }
+
+
+def get_status_for_multi_service_lot_and_service_type(
+    count, services_status, framework_is_open, declaration_complete, unit, unit_plural
+):
+
+    singular = 1 == count
```
TheDoubleK: Also here if you're doing it :)

tests/app/main/helpers/test_frameworks.py
```diff
@@ -0,0 +1,294 @@
...
+        ),
+        (
+            [True, 0, 1, None, 'open'],
+            [{
+                'title': u'Marked as complete',
+                'hint': u'You can edit it until the deadline'
+            }]
+        ),
+        (
+            [True, 0, 1, 'complete', 'open'],
```
TheDoubleK: What about the 'started' state for a declaration? Is that still a thing?  If so, could add that to the test cases.

quis: Good point. Added a test in ff6a507.

tests/app/main/helpers/test_frameworks.py
```diff
@@ -0,0 +1,294 @@
...
+            [True, 1, 0, None, 'open'],
+            [{
+                'title': u'Started but not complete',
+                'type': u'quiet'
+            }]
+        ),
+        (
+            [True, 0, 1, None, 'open'],
+            [{
+                'title': u'Marked as complete',
```
TheDoubleK: Will there be messaging elsewhere to say they still need to complete the declaration? A banner like last time?

quis: We have this:
image

I assumed there was already a banner on this page specifically if they hadn’t done the declaration… doesn’t seem to be the case so I’ll add that.

tests/app/main/helpers/test_frameworks.py
```diff
@@ -0,0 +1,294 @@
...
+            [True, 1, 0, None, 'standstill'],
+            [{
+                'title': u'Not completed',
+                'type': u'quiet'
+            }]
+        ),
+        (
+            [True, 0, 1, None, 'standstill'],
+            [{
+                'title': 'Marked as complete'
```
TheDoubleK: Do we want something to say `(but) Not submitted` here?

TheDoubleK: I know the words have been OK'd already, sorry to stick my nose in.

tests/app/main/helpers/test_frameworks.py
TheDoubleK: That looks angry enough to me as it is.

TheDoubleK:
Not sure what else you have "in progress" with this, but looks good.

It might be too late in the day to change it now, but looking at the screenshot I'm not sure about the word "service" for the labs and participants lots (i.e. in "1 draft service" and "You've completed this service"). We think of them as services, but if I had a user research studio I probably wouldn't think of it as a service - "1 draft studio" or "1 draft facility" might be more descriptive.

But really, let's stick with service at this point.

quis:
@TheDoubleK The screenshot was well out of date. I’ve updated it now. We say "1 complete lab" instead of "service" already; you’re right.

TheDoubleK:
I'm not getting the updated screenshot, but I'll take your word for it.

tests/app/main/helpers/test_frameworks.py
quis: Added a banner in 24e7001.

> **Add banner warning you to complete the declaration**
> 
> We had this for G-Cloud on the "list of services page".
> It was never added to this page when this page was created.
> This commit copies it from the list of services page.

TheDoubleK:
:+1: unless you have anything else to add.

TheDoubleK:
Should wait for Travis though...

quis:
Sue had some minor concerns about “This will be submitted”. If anyone comes up with something soon then I will change it, otherwise merge now, change later.

PR #347, opened 2015-11-23
Make use of draft service validation errors

robyoung:
After https://github.com/alphagov/digitalmarketplace-api/pull/297 validation errors will come back in API responses when getting a single draft service. This change makes use of them to decide whether or not to show the 'mark as complete' button. This is required because not all of the validation rules can be determined just by whether any required fields are unfilled.

### DIFF
--- a/app/templates/services/service_submission.html
+++ b/app/templates/services/service_submission.html
@@ -87,7 +87,7 @@
     {% if service_data.status == 'submitted' and declaration_status == 'complete' %}
       <span class="service-status-published">This service is marked as complete and will be submitted at {{ deadline|safe }}</span>
     {% endif %}
-    {% if service_data.status == 'not-submitted' and unanswered_required == 0 and framework.status == 'open' %}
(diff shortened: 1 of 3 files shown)


pcraig3:
Looks good once the other one is merged :+1:

app/templates/services/service_submission.html
```diff
@@ -87,7 +87,7 @@ <h2 class="temporary-message-heading" id="temporary-message-heading">
     {% if service_data.status == 'submitted' and declaration_status == 'complete' %}
       <span class="service-status-published">This service is marked as complete and will be submitted at {{ deadline|safe }}</span>
     {% endif %}
-    {% if service_data.status == 'not-submitted' and unanswered_required == 0 and framework.status == 'open' %}
+    {% if service_data.status == 'not-submitted' and can_mark_complete and framework.status == 'open' %}
```
pcraig3: +1 for the variable name change.

PR #349, opened 2015-11-25
Remember submitted data on declaration failure

robyoung:
When the declaration validation fails the submitted data should be remembered and presented back into the form.


allait:
:+1:

PR #352, opened 2015-11-25
[#108938292] Update DOS closing date

robyoung:
https://www.pivotaltracker.com/story/show/108938292


quis:
Added hard-coded dates to the isht list so :+1:

PR #354, opened 2015-11-26
Add new statuses for services on framework dashboard

quis:
Based on talking this through with @ralph-hawkins.

## This:
- cuts down the number of status we show you by not talking about drafts on this page
- makes the statuses work for lots which have a limit of one service, as well as lots that have a limit of more than one service
- tells you which lots you are submitting to

image

image

## Also
- Tidy and simplify the dashboard template (to pave the way for the meat of this PR)
- Use new browse list status
- Move declaration above services

### DIFF
--- a/app/assets/scss/_framework-application.scss
+++ b/app/assets/scss/_framework-application.scss
@@ -21,23 +21,23 @@
   //padding-right: 1px;
 }
 
-%framework-application-section,
-.framework-application-section {
+.framework-dashboard {
 
-  border-top: 1px solid $grey-2;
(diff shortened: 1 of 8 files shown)


pcraig3:
Mostly looks awesome.  Only thing I noticed (aside from the updates page) was a strange `/frameworks/digital-outcomes-and-specialists` page when `framework.status == 'pending'`

screen shot 2015-11-26 at 17 30 10

quis:
@pcraig3 Maybe you didn’t run the frontend build/install?

PR #355, opened 2015-11-26
[#107188798] Edit remove subsections

TheDoubleK:
See: https://www.pivotaltracker.com/story/show/107188798

NOTE: This now depends on: 
 - [x] https://github.com/alphagov/digitalmarketplace-api/pull/304

This adds a "Remove" link to subsections that have been added (e.g. specialists), but doesn't let you remove the last one.

screen shot 2015-11-26 at 18 16 51

screen shot 2015-11-26 at 18 17 00

screen shot 2015-11-27 at 16 15 27

### DIFF
--- a/app/main/helpers/services.py
+++ b/app/main/helpers/services.py
@@ -74,3 +74,9 @@ def get_next_section_name(content, current_section_id):
         return content.get_section(
             content.get_next_editable_section_id(current_section_id)
         ).name
+
+
+def get_containing_section(content, question):
+    for section in content:
+        if question in section.questions:
+            return section
--- a/app/main/helpers/services.py
+++ b/app/main/helpers/services.py
@@ -74,9 +74,3 @@ def get_next_section_name(content, current_section_id):
         return content.get_section(
             content.get_next_editable_section_id(current_section_id)
         ).name
-
-
-def get_containing_section(content, question):
-    for section in content:
-        if question in section.questions:
-            return section
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -7,8 +7,8 @@ from ...main import main, content_loader
 from ..helpers.services import (
     is_service_modifiable, is_service_associated_with_supplier,
     get_signed_document_url, count_unanswered_questions,
-    get_next_section_name
-)
+    get_next_section_name, get_containing_section)
+
 from ..helpers.frameworks import get_framework_and_lot, get_declaration_status
 
 from dmutils.apiclient import APIError, HTTPError
@@ -505,6 +505,67 @@ def update_section_submission(framework_slug, lot_slug, service_id, section_id, 
                                 lot_slug=draft['lot'],
                                 service_id=service_id,
                                 _anchor=section_id))
+
+
+@main.route('/frameworks/<framework_slug>/submissions/<lot_slug>/<service_id>/remove/<question_slug>',
+            methods=['GET', 'POST'])
+@login_required
+def remove_subsection(framework_slug, lot_slug, service_id, question_slug):
+    if request.args.get("confirm") and request.method == "POST":
+        # Remove the section
+        try:
+            draft = data_api_client.get_draft_service(service_id)['services']
+        except HTTPError as e:
+            abort(e.status_code)
+
+        if not is_service_associated_with_supplier(draft):
+            abort(404)
+
+        content = content_loader.get_manifest(framework_slug, 'edit_submission').filter(draft)
+
+        question_to_remove = content.get_question_by_slug(question_slug)
+        fields_to_remove = [subquestion.id for subquestion in question_to_remove.questions]
+        containing_section = get_containing_section(content, question_to_remove)
+
+        all_section_fields = []
+        for question in containing_section.questions:
+            for subquestion in question.questions:
+                all_section_fields.append(subquestion.id)
+        all_section_responses = [id for id in all_section_fields if id in draft]
+
+        remaining_fields = [field for field in all_section_responses if field != fields_to_remove]
+        if len(remaining_fields) > 0:
+            # It isn't the last one so remove it
+            for field in fields_to_remove:
+                draft.pop(field, None)
+            try:
+                data_api_client.update_draft_service(
+                    service_id,
+                    draft,
+                    current_user.email_address
+                )
+            except HTTPError as e:
+                abort(e.status_code)
+        else:
+            # You can't remove the last one
+            flash('remove_last_attempted', 'error')
+
+        return redirect(
+            url_for('.view_service_submission',
+                    framework_slug=framework_slug,
+                    lot_slug=lot_slug,
+                    service_id=service_id,
+                    confirm_remove=None
+                    ))
+    else:
+        # Show page with "Are you sure?" message and button
+        return redirect(
+            url_for('.view_service_submission',
+                    framework_slug=framework_slug,
+                    lot_slug=lot_slug,
+                    service_id=service_id,
+                    confirm_remove=question_slug
+        ))
 
 
 def _update_service_status(service, error_message=None):
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -554,7 +554,7 @@ def remove_subsection(framework_slug, lot_slug, service_id, section_id, question
         section_responses = [field for field in containing_section.get_field_names() if field in draft]
         fields_remaining_after_removal = [field for field in section_responses if field not in fields_to_remove]
 
-        if len(fields_remaining_after_removal) > 0:
+        if draft['status'] == 'not-submitted' or len(fields_remaining_after_removal) > 0:
             # Show page with "Are you sure?" message and button
             return redirect(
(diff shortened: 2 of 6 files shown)


quis:
I think the wording of the second banner could do with some love.

app/main/views/services.py
```diff
@@ -503,6 +503,62 @@ def update_section_submission(framework_slug, lot_slug, service_id, section_id,
                                 service_id=service_id))
 
 
+@main.route('/frameworks/<framework_slug>/submissions/<lot_slug>/<service_id>/remove/<question_slug>',
```
allait: Should we keep the URLs similar to section edits with `/remove/<section_id>/<question_slug>`? That way if we ever add a way to clear section contents too we won't have a URL clash.

app/main/views/services.py
```diff
@@ -503,6 +503,62 @@ def update_section_submission(framework_slug, lot_slug, service_id, section_id,
...
+            abort(404)
+
+        content = content_loader.get_manifest(framework_slug, 'edit_submission').filter(draft)
+        question_to_remove = content.get_question_by_slug(question_slug)
+        containing_section = get_containing_section(content, question_to_remove)
+        fields_to_remove = question_to_remove.form_fields
+        section_responses = [field for field in containing_section.get_field_names() if field in draft]
+        fields_remaining_after_removal = [field for field in section_responses if field not in fields_to_remove]
+
+        if len(fields_remaining_after_removal) > 0:
```
allait: I don't think we really need these checks here – it's easier to send the request and let the API return an error if "at leas one" validation fails. This will require us to handle the API errors, but gets rid of the validation logic duplication (that can get more complicated depending on the section). We also don't really care about removing last section on a not-submitted draft.

app/main/views/services.py
TheDoubleK: Good point.  I did try to think of a good reason to keep the longer version of the URL - this is it.

TheDoubleK:
Rebased for the umpteenth time.

app/main/helpers/services.py
```diff
@@ -74,3 +74,9 @@ def get_next_section_name(content, current_section_id):
         return content.get_section(
             content.get_next_editable_section_id(current_section_id)
         ).name
+
+
+def get_containing_section(content, question):
```
allait: Do we still need this now that the URL includes the section_id? It seems we can do the same thing we do in the edit section endpoint

app/main/views/services.py
```diff
@@ -506,6 +507,76 @@ def update_section_submission(framework_slug, lot_slug, service_id, section_id,
...
+
+    content = content_loader.get_manifest(framework_slug, 'edit_submission').filter(draft)
+    question_to_remove = content.get_question_by_slug(question_slug)
+    containing_section = get_containing_section(content, question_to_remove)
+    fields_to_remove = question_to_remove.form_fields
+
+    if request.args.get("confirm") and request.method == "POST":
+        # Remove the section
+        for field in fields_to_remove:
+            draft[field] = None
```
allait: I don't think we should be submitting the full draft as an update – we should be sending the question fields with `None` values only, since it's the only thing we're changing. The API should be able to figure out the rest. The only thing we could do here is check if none of the fields are in the draft and then skip sending the request, but since we hide the delete button and there's no harm in sending the request we can safely skip that.

app/main/views/services.py
```diff
@@ -506,6 +507,76 @@ def update_section_submission(framework_slug, lot_slug, service_id, section_id,
...
+            if e.status_code == 400:
+                # You can't remove the last one
+                flash({
+                    'remove_last_attempted': containing_section.name,
+                    'service_name': question_to_remove.label
+                }, 'error')
+            else:
+                abort(e.status_code)
+
+    else:
```
allait: I don't think we need this whole block anymore since the "last section" check is taken from the API response after the delete attempt failed

pcraig3:
Seems okay to me.  Only comment is what we've been talking about on Slack.

`'digital specialists'` vs `"Digital Specialists"`

TheDoubleK:
Latest update:
screen shot 2015-11-30 at 14 25 02

TheDoubleK:
And there's more... agreed with @ralph-hawkins and Sarah.

screen shot 2015-11-30 at 14 41 02

app/main/helpers/services.py
TheDoubleK: Good point - yes if we're doing the long URL we should use the extra parameter.

app/main/views/services.py
TheDoubleK: The reason for the checks here is to avoid showing the "Are you sure?" message when clicking "Remove" on the last item. In that case the app needs to decide to show the "You can't remove the last one" without calling the API at all.

TheDoubleK: Because we shouldn't make people click two things before telling them they can't delete it.

allait: That's an interesting point. We need to show the "Are you sure?" dialog mostly because the initial "Remove" is a link and we can't use a GET request for the action.

But I don't think duplicating this check here is a good solution. We depend on the API response because it's one place that will be updated if this logic ever changes. By having a check here we either make the API response handling redundant (like it is right now) or go back to telling people they can't delete something after the confirmation (if the conditions diverge).

At the same time, this is only a problem for submitted drafts (which I think this block doesn't check), since there's no reason to prevent deletion if the draft wasn't marked as complete. So since the action should be rare to begin with, I'm still suggesting we just drop this check.

app/main/views/services.py
```diff
@@ -507,6 +508,74 @@ def update_section_submission(framework_slug, lot_slug, service_id, section_id,
...
+        draft = data_api_client.get_draft_service(service_id)['services']
+    except HTTPError as e:
+        abort(e.status_code)
+
+    if not is_service_associated_with_supplier(draft):
+        abort(404)
+
+    content = content_loader.get_manifest(framework_slug, 'edit_submission').filter(draft)
+    question_to_remove = content.get_question_by_slug(question_slug)
+    containing_section = content.get_section(section_id)
```
allait: This should be in the opposite order: find the section, get `section.get_question_by_slug` if the is not None. This is also missing the `editable` checks, which means it could make it possible to delete read-only fields (I think the API will complain, but I think this could still be a problem). I'd copy the relevant bit from `edit_service_submission` since that's likely to do the right thing.

scripts/run_app.sh
```diff
@@ -11,7 +11,7 @@ export DM_DATA_API_URL=${DM_DATA_API_URL:=http://localhost:5000}
 export DM_DATA_API_AUTH_TOKEN=${DM_DATA_API_AUTH_TOKEN:=myToken}
 export DM_API_AUTH_TOKEN=${DM_API_AUTH_TOKEN:=myToken}
 
-export DM_SUBMISSIONS_BUCKET=${DM_SUBMISSIONS_BUCKET:=digitalmarketplace-submissions-dev-dev}
+export DM_SUBMISSIONS_BUCKET=${DM_SUBMISSIONS_BUCKET:=digitalmarketplace-documents-dev-dev}
 export DM_COMMUNICATIONS_BUCKET=${DM_COMMUNICATIONS_BUCKET:=digitalmarketplace-communications-dev-dev}
```
allait: I think this should be `digitalmarketplace-documents-dev-dev` too

scripts/run_app.sh
```diff
@@ -11,7 +11,7 @@ export DM_DATA_API_URL=${DM_DATA_API_URL:=http://localhost:5000}
...
 export DM_ASSETS_URL=${DM_ASSETS_URL:=http://asset-host}
```
allait: I've left this comment on Rob's PR, but I think it got merged without it, so I'll copy it here:
> I think the assets URL should be the digitalmarketplace-documents-dev-dev S3 URL: https://${DM_DOCUMENTS_BUCKET}.s3-eu-west-1.amazonaws.com, since we're using a real S3 bucket for dev we can actually link directly to it from services

app/main/views/services.py
TheDoubleK: @ralph-hawkins insisted that they shouldn't be shown a button if we know it will fail...

allait: We've talked about this a bit more with @ralph-hawkins and it seems this is a good middle ground between making a proper "validate update" endpoint in the API and just dropping the validation if this turns out to be a rare edge case, so
I suggest we make sure this (attempt to remove the last subsection on a complete draft) can be tracked by the analytics.

We still need the draft status check so that it doesn't prevent removing subsections if the draft wasn't submitted.

TheDoubleK:
"Submitted" check added and virtual pageview in.

I have no idea how to test the virtual pageview, but it's done the same way as for the "complete" button.

TheDoubleK:
Everything is green and all comments addressed I think.  Please can we merge this and raise bugs later? :)

allait:
:+1: 

<sub>Although we _could_ do with some tests for the non-edit_questions sections and missing section/questions</sub>

PR #356, opened 2015-11-27
Update submissions to use new submissions bucket

robyoung:


allait:
`scripts/run_app.sh` doesn't use `test_env.sh`, so it needs to be updated too

allait:
:+1:

PR #359, opened 2015-11-27
Better buttons

quis:
## Line buttons up with the column they sit in

image

image

## Make delete button label lot-specific

This makes the text of the delete button name the lot, if the lot has a limit of one service. This makes sense because deleting the services effectively deletes the whole lot, and talking about these things as services is weird.

image

### DIFF
--- a/app/assets/scss/_service_submission.scss
+++ b/app/assets/scss/_service_submission.scss
@@ -4,6 +4,7 @@
 
 .align-with-heading .button-save {
     margin-top: 0px;
+    width: 100%;
 }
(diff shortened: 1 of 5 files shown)


TheDoubleK:
Tested - buttons look good (better). :+1: 

I have heard talk that the button for removing labs should say "Remove this lab" rather than "Remove this service", but maybe not one for this PR.

quis:
Yes. Ralph and I have been talking about that. We’d need to figure what the best way of storing that information would be.

PR #360, opened 2015-11-27
Make clarification emails non-G7 specific

allait:
### Make clarification emails non-G7 specific
Removes the `G7_CLARIFICATIONS_CLOSED` feature flag in favour of the
framework data `clarificationQuestionsOpen` flag returned by the API.
This requires a switch in `if ... else ...` order to make conditions
more readable.

Changes to clarification emails sent to zendesk:
* Email subject changed from "Clarification question" to "<framework>
  clarification question"
* Email from address changed from `suppliers@...` to
  `suppliers+<framework-slug>@...`

Changes to application emails:
* Subject will use the current framework name in "<framework>
  application question"
* Tags changed from "g7-application-question" to "application-question"

#### Add framework slug to clarification email logs and audit events

### DIFF
--- a/app/main/helpers/frameworks.py
+++ b/app/main/helpers/frameworks.py
@@ -252,4 +252,4 @@ def get_status_for_multi_service_lot_and_service_type(
 def has_one_service_limit(lot_slug, framework_lots):
     for lot in framework_lots:
         if lot['slug'] == lot_slug:
-            return lot['one_service_limit']
+            return lot['oneServiceLimit']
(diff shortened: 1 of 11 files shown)


app/main/helpers/frameworks.py
```diff
@@ -252,4 +252,4 @@ def get_status_for_multi_service_lot_and_service_type(
 def has_one_service_limit(lot_slug, framework_lots):
     for lot in framework_lots:
         if lot['slug'] == lot_slug:
-            return lot['one_service_limit']
+            return lot['oneServiceLimit']
```
quis: Is this changing in a breaking way? I’ve been using the :snake: :briefcase: one in a few places :disappointed:.

allait: The API responds with both for now but the intention was to drop the old one some time soon. I've updated everything I could find in the supplier-frontend, were you using it somewhere else?

quis: I think it’s just in some of my PRs at the moment. I’ll see if I can hunt them down at change them.

robyoung:
Looks good :+1:

PR #361, opened 2015-11-27
Fix breadcrumbs on lots with a limit of one service

quis:
If a lot has a limit of one service it doesn’t make sense to link to the lot from the service page—it just links to the current page, which we don’t do anywhere else.

This commit adds a condition to hide the link.

Fixes [109072944](https://www.pivotaltracker.com/story/show/109072944).


TheDoubleK:
Yep - nice one :+1:

PR #362, opened 2015-11-27
Linkifying URLs in summary tables

tommorris:
Requires the merging of [PR 192](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/192) on frontend toolkit.

If there's a URL, people should be able to click on it.

### DIFF
--- a/app/templates/suppliers/_information.html
+++ b/app/templates/suppliers/_information.html
@@ -24,7 +24,7 @@
   {% endcall %}
   {% call summary.row() %}
     {{ summary.field_name('Website') }}
-    {{ summary.text(supplier.contact.website) }}
+    {{ summary.external_link(supplier.contact.website, supplier.contact.website)}}
   {% endcall %}
   {% call summary.row() %}
     {{ summary.field_name('Address') }}
--- a/bower.json
+++ b/bower.json
@@ -5,7 +5,7 @@
     "jquery": "1.11.2",
     "hogan": "3.0.2",
     "jquery-details": "https://github.com/mathiasbynens/jquery-details/archive/v0.1.0.tar.gz",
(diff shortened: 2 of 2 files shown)


allait:
Is there a story for this?
I'm not sure what's the value of linking the supplier to their own website. And whether we have any rules about external links (eg "nofollow" or special styling) that apply to supplier pages.

robyoung:
I'm not sure we want to be linking off to random sites at all from a gov.uk domain. I'd want to double check with the GOV.UK content team.

quis:
It’s from a URL that only suppliers can see, to a URL that they have inputted, so I don’t have any objection on that front.

We should mark it as an external link though, there are [mixins in the GOV.UK frontend toolkit](https://github.com/alphagov/govuk_frontend_toolkit/blob/fbe3ce944b7ebd22e3e965d6f19265c4035354a3/stylesheets/_typography.scss#L269) for doing this.

robyoung:
That's a fair point that they are the only ones that can see it.

quis:
Which makes you wonder, why do we ask for it at all? We never show it to buyers.

tommorris:
There's no story for it - it just seems very strange that we have a bare URL on the supplier page so I submitted a simple fix for it. If we present a link, people should be able to click on it, otherwise that's a usability fail.

TheDoubleK:
I think the "why do we ask for it" comes down to the will-they?/won't-they? get a supplier information page on the marketplace stuff from a way back.  We now have the supplier A-Z pages, but I think it was felt that giving them a link to their sites from GOV.UK would be a bit too much.

tommorris:
The build fails at the moment, but once [the PR on frontend-toolkit](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/192) gets merged and released, it should merge.

quis:
Merged the toolkit PR and rebased.

PR #363, opened 2015-11-30
Jump links

quis:
## Make 'Answer required' jump to the exact question

This worked in previous SSPs. It got lost somewhere along the way.

## Jump to summary section after answering a question

It is jarring to be taken back to the top of a long page after answering a question, and having to scroll all the way down to find where the answer is played back (and potentially where the next section of questions you want to answer are).

This commit adds an anchor to jump between these pages more seamlessly.

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -500,10 +500,15 @@ def update_section_submission(framework_slug, lot_slug, service_id, section_id, 
                                 service_id=service_id,
                                 section_id=next_section))
     else:
-        return redirect(url_for(".view_service_submission",
-                                framework_slug=framework['slug'],
-                                lot_slug=draft['lot'],
-                                service_id=service_id))
+        return redirect("{}#{}".format(
+            url_for(
+                ".view_service_submission",
+                framework_slug=framework['slug'],
+                lot_slug=draft['lot'],
+                service_id=service_id
+            ),
+            section_id
+        ))
 
 
 def _update_service_status(service, error_message=None):
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -500,15 +500,11 @@ def update_section_submission(framework_slug, lot_slug, service_id, section_id, 
                                 service_id=service_id,
                                 section_id=next_section))
     else:
-        return redirect("{}#{}".format(
-            url_for(
-                ".view_service_submission",
-                framework_slug=framework['slug'],
-                lot_slug=draft['lot'],
-                service_id=service_id
-            ),
-            section_id
-        ))
+        return redirect(url_for(".view_service_submission",
+                                framework_slug=framework['slug'],
+                                lot_slug=draft['lot'],
+                                service_id=service_id,
+                                _anchor=section_id))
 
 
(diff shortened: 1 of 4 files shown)


app/main/views/services.py
```diff
@@ -500,10 +500,15 @@ def update_section_submission(framework_slug, lot_slug, service_id, section_id,
                                 service_id=service_id,
                                 section_id=next_section))
     else:
-        return redirect(url_for(".view_service_submission",
-                                framework_slug=framework['slug'],
-                                lot_slug=draft['lot'],
-                                service_id=service_id))
+        return redirect("{}#{}".format(
+            url_for(
```
allait: `url_for` has a special keyword argument `_anchor` for URL anchors, so we can avoid a manual `.format` (http://flask.pocoo.org/docs/0.10/api/#flask.url_for)

quis: Clever Flask. Added in e394036.

app/templates/services/_base_service_page.html
```diff
@@ -17,45 +17,47 @@
...
-              {% call summary.field() %}
-                <span class="summary-item-field-answer-required">{{ question.empty_message }}</span>
-              {% endcall %}
-            {% else %}
-              {{ summary[question.type](question.value, question.assurance) }}
-            {% endif %}
-            {% if section.edit_questions %}
-              {{ summary.edit_link('Add' if question.is_empty else 'Edit', url_for(".edit_service_submission", framework_slug=framework.slug, lot_slug=service_data.lot, service_id=service_id, section_id=section.id, question_slug=question.slug)) }}
-            {% endif %}
+        <div id="{{ section.slug }}">
```
allait: Is there a benefit of having the whole section wrapped in the element with id as opposed to the heading only?
My first though was adding the id to the heading itself, which might make it easier to do the same thing on similar pages.

quis: It would be cleaner to have `section.heading` take `id` as an optional parameter. Maybe I should not be lazy.

TheDoubleK:
Will this fix this bug? Or is this something else?
https://www.pivotaltracker.com/story/show/109072494

app/templates/services/_base_service_page.html
```diff
@@ -39,7 +39,7 @@
             {% if question.answer_required %}
               {% call summary.field() %}
                 {% if framework and framework.status == 'open' %}
-                    <a href="{{ url_for(".edit_service_submission", framework_slug=framework.slug, lot_slug=service_data.lot, service_id=service_id, section_id=section.id) }}">Answer required</a>
+                    <a href="{{ url_for(".edit_service_submission", framework_slug=framework.slug, lot_slug=service_data.lot, service_id=service_id, section_id=section.id) }}#{{ question.id }}">Answer required</a>
```
allait: This could use `_anchor` too

allait:
:+1:

quis:
@TheDoubleK Should do :tada:

allait:
@TheDoubleK @quis I don't think that bug is related. It's about the link from field name in the validation errors block to the question form field

PR #364, opened 2015-11-30
Remove G-Cloud 7 deadline and declaration link

pcraig3:
This pull request makes a few small changes where I noticed problems while clicking through the app.

1. Services that were marked as complete before the framework declaration was finished would direct users to the G7 declaration.
2. Otherwise, once the (DOS) declaration was completed, the old G7 deadline would be displayed.
3. Added time of day into the `DOS_CLOSING_DATE`

##### `DOS_CLOSING_DATE`
screen shot 2015-11-30 at 13 26 07

### Notes

Should we move the DOS closing date into `digitalmarketplace-frameworks`?

There's still [one reference to `g_cloud_7_essentials` still in `_event.js`](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/808f074085346aaf70baa74491cc7f3d1210f428/app/assets/javascripts/analytics/_events.js#L5)?  Do we change it or get rid of it?  I don't know what it is.

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -385,7 +385,7 @@ def view_service_submission(framework_slug, lot_slug, service_id):
         can_mark_complete=not validation_errors,
         delete_requested=delete_requested,
         declaration_status=get_declaration_status(data_api_client, framework['slug']),
-        deadline=current_app.config['G7_CLOSING_DATE'],
+        deadline=current_app.config['DOS_CLOSING_DATE'],
         **main.config['BASE_TEMPLATE_DATA']), 200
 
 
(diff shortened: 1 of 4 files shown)


quis:
The `g_cloud_7_essentials` is used in the code that tracks how far down someone has scrolled on the supplier declaration. We decided not to fix it for DOS because we didn’t use this data. Could delete that functionality entirely but extra work so…

pcraig3:
Okay, so it's fine if we just leave it in?

quis:
@pcraig3 Yeah, I think so.

:+1:

PR #365, opened 2015-11-30
Make "Save and continue" button return to summary

allait:
Removes the secondary "Save and return to summary" button and makes
all service "Save and continue" buttons return to the summary page
with the current section slug anchor.

Removes `next_section` and `return_to_summary` mentions from service
pages.

### DIFF
--- a/app/main/helpers/services.py
+++ b/app/main/helpers/services.py
@@ -67,10 +67,3 @@ def parse_document_upload_time(data):
     match = re.search("(\d{4}-\d{2}-\d{2}-\d{2}\d{2})\..{2,3}$", data)
     if match:
         return datetime.strptime(match.group(1), "%Y-%m-%d-%H%M")
-
-
-def get_next_section_name(content, current_section_id):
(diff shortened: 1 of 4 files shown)


TheDoubleK:
We are going to end up reverting this again the next time a G-Cloud framework comes along because going back and forth for forty pages of questions (e.g. adding a PasS service) will be awful this way round.

TheDoubleK:
Would it be madness to have the button behaviour conditional on the framework?  Could be a flag in the frameworks table, or just a `'dos' or 'gcloud'` in the view for now?

allait:
@TheDoubleK I don't think we want behaviour conditional on framework. It might be something about manifest size and structure, but as far as I understand two buttons are confusing and secondary button is rarely used, suppliers use summary page quite a lot, and with the "scroll to section" it might be not that bad. We might have to look at analytics a bit more to decide if we actually want it for G-Cloud again.

robyoung:
This seems fine right now, if we need to bring it back for G-Cloud we can solve it more thoroughly then.
:+1:

PR #366, opened 2015-11-30
Make contextual help make sense

quis:
## Get rid of required answer hint text
This is replaced by…

## Show section description on summary page

Brings in and depends on:
- [x] alphagov/digitalmarketplace-frameworks#147
- [x] alphagov/digitalmarketplace-frontend-toolkit#196
- [x] alphagov/digitalmarketplace-utils#220

### DIFF
--- a/app/templates/services/_base_edit_section_page.html
+++ b/app/templates/services/_base_edit_section_page.html
@@ -16,6 +16,12 @@
         {% include 'toolkit/page-heading.html' %}
       {% endwith %}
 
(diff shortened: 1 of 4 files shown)


allait:
I think we need to check that section descriptions are displayed on the edit section pages. I think until now they were only used by the declaration.

quis:
@allait Added section descriptions to services here: 5438aa8

allait:
:+1:

PR #372, opened 2015-12-02
Add spacing for question heading linebreaks

pcraig3:
Added some spacing between linebreaks in our really long question headings.
Only targeting paragraph tags in headings (excluding the last one), so the rendering for the majority of question headings doesn't change.

I'm making the change here instead of the frontend toolkit because the markup in each is different.
The supplier app passes all question headings through a markdown filter so the `<p>` tags only exist in this app.

Markup in the supplier app:
```html
<span class="question-heading ">
  <p>Where will your agile coach work?</p>
</span>
```

Markup in the frontend toolkit:
```html
<span class="question-heading ">
  Where will your agile coach work?
</span>
```

Adding CSS in the frontend toolkit that would only be interpreted in the supplier app because of a different way of rendering forms seemed like the wrong thing to me.

***

### Before 
24a0fc80-9780-11e5-9475-aae7e6c7cb47

### Now
screen shot 2015-12-02 at 16 02 19

### DIFF
--- a/app/assets/scss/_forms.scss
+++ b/app/assets/scss/_forms.scss
@@ -7,3 +7,11 @@
 a.button-save {
   @include inline-block;
 }
+
+.question-heading p {
+  margin-bottom: 10px;
+
+  &:last-child {
+    margin-bottom: 0;
+  }
+}


quis:
Think you could argue the toolkit/app thing either way. This is :+1: for now.

PR #373, opened 2015-12-03
Content from content update

ralph-hawkins:
* Added back links to service summary pages
* Lots of minor content updates
* Put the apply to framework section in a two-thirds grid


app/templates/frameworks/updates.html
```diff
@@ -75,13 +75,13 @@
     <div class="column-one-whole updates-document-tables">
       {% set sections = [
         {
-          "heading": framework.name + " communications",
-          "empty_message": "No communications have been sent out",
+          "heading": " Communications",
```
allait: Why do we need a leading space here?

quis:
Going to look at fixing the tests here…

quis:
Tests fixed in 41004a2, rebased.

app/templates/suppliers/_frameworks_open.html
```diff
@@ -1,39 +1,42 @@
...
+          with
+          items = [{
+            "title":
+            "Continue your {} application".format(framework.name),
+            "link": url_for(".framework_dashboard", framework_slug=framework.slug),
+            "body": "Deadline: {}".format(framework.dates.deadline)
+          }]
         %}
-            {% include "toolkit/button.html" %}
+          {% include "toolkit/browse-list.html" %}
```
allait: This is unrelated to this PR since it was here before, but is this ok to use `browse-list` like this? Usually, we'd pass in a list of items and call it once, but this calls browse-list with a single item for each framework, which creates a different layout. Are we doing this on purpose?

allait:
:+1:

app/templates/suppliers/_frameworks_open.html
quis: We have to do it this way because we’re interviewing browse list items and _apply now_ buttons, eg if there were 5 open frameworks and you’d starting applying to the 2nd and 4th ones. Does mean the markup isn’t optimal.

PR #374, opened 2015-12-03
Make question headings not bold

quis:
Brings in:
- [x] https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/200


allait:
:+1:

PR #375, opened 2015-12-03
Change status text for draft studios

quis:
Came out of yesterday’s content review. This is more consistent with the other lots.


pcraig3:
:+1:

PR #376, opened 2015-12-03
Consolidate framework dates in the content repo

robyoung:
Dates consistently change between frameworks. This moves all dates into
the digitalmarketplace-frameworks repo as messages. Ultimately it would
be good if these were stored in the API against a framework. However,
until we have a clear understanding of what dates are required, what
they mean, and how we want to manage them consolidating them in one
place is a good first step.


pcraig3:
:+1:

PR #377, opened 2015-12-03
Show label for contributor email field

tombye:
This field was using the `single-question-page`
class. This class currently assumes the page
heading will be a duplicate of the field label so
hides it.

The field was also using a different markup
pattern to other pages with one question (for
example, those in the 'create a new supplier
flow'). The markup for this pattern was missing
some span tags, causing the label to be hidden
normally but shown if validation messages were
present.

This change removes the `single-question-page`
class to ensure the label is shown and swaps the
markup out to use the [toolkit textbox pattern](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/forms/textbox.html)
(which has markup that works both for the field
in its base and invalid states).

The page now looks like this:

image

image

### DIFF
--- a/app/templates/auth/submit-email-address.html
+++ b/app/templates/auth/submit-email-address.html
@@ -27,72 +27,61 @@
 
 {% block main_content %}
 
-<div class="single-question-page">
+{% with messages = get_flashed_messages(with_categories=true) %}
+{% if messages %}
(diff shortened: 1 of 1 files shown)


quis:
Tested locally, :+1:

PR #378, opened 2015-12-03
Bump utils to 15.6.1

pcraig3:
Brings in
- [#221](https://github.com/alphagov/digitalmarketplace-utils/pull/221)
  which fixes rendering for questions with assurances

None are breaking changes, so there shouldn't be anything else to do.


robyoung:
:+1:

PR #379, opened 2015-12-04
Fix deadline markup escaping bug

robyoung:
Marking a string as safe does not work when building a string with .format.

### DIFF
--- a/app/templates/suppliers/_frameworks_open.html
+++ b/app/templates/suppliers/_frameworks_open.html
@@ -8,7 +8,7 @@
             "title":
             "Continue your {} application".format(framework.name),
             "link": url_for(".framework_dashboard", framework_slug=framework.slug),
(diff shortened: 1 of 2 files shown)


app/templates/suppliers/_frameworks_open.html
```diff
@@ -8,7 +8,7 @@
             "title":
             "Continue your {} application".format(framework.name),
             "link": url_for(".framework_dashboard", framework_slug=framework.slug),
-            "body": "Deadline: {}".format(framework.dates.framework_close_date|safe)
```
TheDoubleK: But Ralph only just added that colon!

TheDoubleK:
Looks good :+1:
And a pesky looking bug - hope you didn't lose too much time to it.

quis:
Attaching it to the framework feels like a nicer solution irrespective of the bug. Nice.

PR #381, opened 2015-12-04
Update utils and content versions

allait:
Includes ordered error labels, fix for min/max price error link ids
and "Apply to privide" lot links text.


TheDoubleK:
:+1: although there are further version bumps to both things coming soon by the look of it.

PR #383, opened 2015-12-04
Use lot name for summary of completed lots

quis:
This comes from the API.

Previously used lot label from the content, which has changed to start with
'Apply to provide…'

image

--

Fixes https://www.pivotaltracker.com/story/show/109529008

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -55,8 +55,6 @@ def framework_dashboard(framework_slug):
     declaration_status = get_declaration_status_from_info(supplier_framework_info)
     supplier_is_on_framework = get_supplier_on_framework_from_info(supplier_framework_info)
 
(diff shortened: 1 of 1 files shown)


app/main/views/frameworks.py
```diff
@@ -80,7 +80,10 @@ def framework_dashboard(framework_slug):
             "complete": len(complete_drafts)
         },
         completed_lots=[{
-            'name': lot['label'],
+            'name': [
+                l['name']
+                for l in framework['lots'] if l['slug'] == lot['value']
```
allait: We can get rid of `lot_question` now and just use `framework['lots']`

quis:
@allait Changed to use framework from API and squashed into f1bb1cc.

allait:
:+1:

PR #384, opened 2015-12-04
Fix declaration validation

robyoung:
This fixes two bugs;
- Character limit fails on empty text fields
- Email validation fails when empty

This also adds default messages for character limit and invalid format. These are not ideal but we do not have any custom validation messages in the content repo for these questions so this is better than nothin.


allait:
:+1:

robyoung:
Nice, I'm nicking that.

PR #385, opened 2015-12-04
Send application started welcome email to supplier users

allait:
[#107188306](https://www.pivotaltracker.com/story/show/107188306)

Sends a welcome email to all supplier users. Email is sent using
the updated `dmutils.email.send_email`: each recipient will get
a separate message with only one 'to' address.

The request will proceed without returning an error if send_email call
raises an exception, since if framework interest is registered a page
refresh would show the dashboard anyway.

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -10,7 +10,6 @@ import six
 
 from dmutils.apiclient import APIError
 from dmutils.audit import AuditTypes
-from dmutils import flask_featureflags
 from dmutils.email import send_email, MandrillException
 from dmutils.formats import format_service_price, datetimeformat
 from dmutils import s3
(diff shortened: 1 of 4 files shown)


app/main/views/frameworks.py
```diff
@@ -44,10 +43,25 @@ def framework_dashboard(framework_slug):
...
 
     if request.method == 'POST':
+        register_interest_in_framework(data_api_client, framework_slug)
+        supplier_users = data_api_client.find_users(supplier_id=current_user.supplier_id)
+
         try:
-            register_interest_in_framework(data_api_client, framework_slug)
-        except APIError as e:
-            abort(e.status_code)
+            email_body = render_template('emails/dos_application_started.html')
```
robyoung: Ultimately this should be loaded by framework slug, we should probably add it to the SSFL.

allait: Email template is specific to dos, we'd have to make it part of the frameworks content to make this generic

robyoung: No, I mean using the framework slug in the file path of the email template. Or indeed, frameworks content works too.

robyoung:
:+1:

app/main/views/frameworks.py
robyoung: This is not a blocker though.

PR #386, opened 2015-12-04
Make section descriptions look the same

quis:
In both the declaration and the 'service essentials' sections. In other words, don’t have it scoped to the declaration.

Also sets the type size explicitly for safety.


ralph-hawkins:
:+1:

PR #388, opened 2015-12-05
[#109538382] Clarifications "from" should be framework-specific

TheDoubleK:
See: https://www.pivotaltracker.com/story/show/109538382

Previously all clarification questions were being sent "from" `G-Cloud 7 Supplier`.
Instead the framework in the "from" should match the framework that the question is about.

### DIFF
--- a/tests/app/main/test_frameworks.py
+++ b/tests/app/main/test_frameworks.py
@@ -1197,7 +1197,7 @@ class TestSendClarificationQuestionEmail(BaseApplicationTest):
                 "MANDRILL",
                 "Test Framework clarification question",
                 "suppliers+g-cloud-7@digitalmarketplace.service.gov.uk",
-                "G-Cloud 7 Supplier",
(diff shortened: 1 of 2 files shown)


tests/app/main/test_frameworks.py
```diff
@@ -1197,7 +1197,7 @@ def _assert_clarification_email(self, send_email, is_called=True, succeeds=True)
                 "MANDRILL",
                 "Test Framework clarification question",
                 "suppliers+g-cloud-7@digitalmarketplace.service.gov.uk",
```
quis: Should check with Nicole what email address she wants this to come from so her filters work.

TheDoubleK: This is only in the test case.  Actual emails are paramaterised by framework.

allait:
:+1: 

As far as I remember, there are no filters set up on the sender email address or name, but it might be worth checking with Nicole.

PR #390, opened 2015-12-07
Fix declaration validation

TheDoubleK:
Two things fixed here:
  * Validation was treating `None` values as though they were valid answers for required strings
  * Conditional validation for Q72 (`licenceOrMemberRequiredDetails`) had the wrong condition on it

Fixes these three bugs:
https://www.pivotaltracker.com/story/show/109608322
https://www.pivotaltracker.com/story/show/109608580
https://www.pivotaltracker.com/story/show/109608626


allait:
:+1:

PR #391, opened 2015-12-07
Don't send welcome email to deactivated users

allait:
Is a user was removed from the contributors list they should not
receive the application started email.


pcraig3:
:+1:

PR #393, opened 2015-12-08
Redirect when logged in

robyoung:
## Redirect logged in users on /login

If a logged in user navigates to the login page they should be
redirected to the supplier dashboard.

## Change redirect flow for create-users
This commit changes the redirect flow for the `/suppliers/create-user` route. Before this change if a logged in user clicked the create user link for a different email address they would be redirected to the `/suppliers/login` route rather than the supplier dashboard. Since changing the `/suppliers/login` route to redirect logged in users to the supplier dashboard this different flow just adds in another redirect step.

With this change we redirect the user to the same place (the supplier dashboard) but show them an error message telling them what has gone wrong.

The tests associated with this flow have been updated to not follow redirects and we instead verify the location header and cookies are set. The reason for this is that the supplier dashboard has a lot of dependencies which detract from the meaning of the create user tests.

### DIFF
--- a/app/main/views/login.py
+++ b/app/main/views/login.py
@@ -253,10 +253,12 @@ def create_user(encoded_token):
         # valid supplier account exists
         if user.role == 'supplier' and not is_registered_to_another_supplier and not is_inactive_or_locked:
             # valid supplier account exists and is logged in
-            if not current_user.is_anonymous() and current_user.email_address == user.email_address:
+            if not current_user.is_authenticated():
+                return redirect(url_for('.render_login'))
+            else:
+                if current_user.email_address != user.email_address:
+                    flash('You are trying to create a user while logged in as a different user', 'error')
                 return redirect(url_for('.dashboard'))
-
-            return redirect(url_for('.render_login'))
 
         return render_template(
             "auth/update-user.html",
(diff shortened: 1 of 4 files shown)


app/main/views/login.py
```diff
@@ -251,11 +253,13 @@ def create_user(encoded_token):
         # valid supplier account exists
         if user.role == 'supplier' and not is_registered_to_another_supplier and not is_inactive_or_locked:
             # valid supplier account exists and is logged in
-            if not current_user.is_anonymous() and current_user.email_address == user.email_address:
+            if not current_user.is_authenticated():
+                return redirect(url_for('.render_login'))
+            else:
+                if current_user.email_address != user.email_address:
+                    flash('You are trying to create a user while logged in as a different user', 'error')
```
allait: Not sure if this message still makes sense once you're redirected to the dashboard. But I suppose it's better than nothing and this shouldn't happen to begin with.

allait:
:+1:

app/main/views/login.py
```diff
@@ -23,6 +23,8 @@
 
 @main.route('/login', methods=["GET"])
 def render_login():
+    if current_user.is_authenticated():
+        return redirect(url_for('.dashboard'))
```
allait: This should probably redirect to `next_url` if it's set and starts with `/suppliers`, similar to what POST endpoint does

robyoung: Good call, I'll update that.

PR #394, opened 2015-12-08
[#109632740] Updated lab pricing display

TheDoubleK:
See: https://www.pivotaltracker.com/story/show/109632740

This will bring in the updated lab price display from the utils update that it depends on:
 - [x] https://github.com/alphagov/digitalmarketplace-utils/pull/227


quis:
Tested locally, :+1:

PR #396, opened 2015-12-10
Update titles

robyoung:
https://www.pivotaltracker.com/story/show/109883600

## Pass framework into edit_service templates
The framework is required by the framework but it hasn't been passed in. This was raised while trying to change the page titles.

## Change page titles to remove supplier specifics

While a framework is open for submissions we are not allowed to see what specific suppliers are doing. This change removes supplier identifiable information from titles as these are sent to google and therefore are very easy for us to see.

### DIFF
--- a/app/templates/services/_base_service_page.html
+++ b/app/templates/services/_base_service_page.html
@@ -1,6 +1,6 @@
 {% extends "_base_page.html" %}
 
-{% block page_title %}{{ service_data.serviceName or service_data.lotName }} – Digital Marketplace{% endblock %}
+{% block page_title %}{{ framework.name }} {{ service_data.lotName }} – Digital Marketplace{% endblock %}
 
 {% block main_content %}
   <div class="grid-row">
--- a/app/templates/services/_base_service_page.html
+++ b/app/templates/services/_base_service_page.html
@@ -1,6 +1,6 @@
 {% extends "_base_page.html" %}
 
-{% block page_title %}{{ framework.name }} {{ service_data.lotName }} – Digital Marketplace{% endblock %}
+{% block page_title %}{{ service_data.serviceName or service_data.lotName }} – Digital Marketplace{% endblock %}
 
 {% block main_content %}
   <div class="grid-row">
--- a/app/templates/suppliers/dashboard.html
+++ b/app/templates/suppliers/dashboard.html
@@ -1,5 +1,5 @@
 {% extends "_base_page.html" %}
-{% block page_title %}{{ current_user.supplier_name }} – Digital Marketplace{% endblock %}
+{% block page_title %}Supplier dashboard – Digital Marketplace{% endblock %}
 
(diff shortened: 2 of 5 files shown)


TheDoubleK:
This seems a weird thing to do - the titles of pages should be based on end-user need, not our analytics need. It is only the main supplier dashboard page that has the supplier name in it, and that doesn't tell anyone anything about their submission or otherwise to the currently open procurement, so I don't see what the issue is anyway?

I guess most people don't actually look at the title much, and this is probably fine to change.  But it does seem a bit of an odd reasoning behind it.

app/templates/services/_base_service_page.html
```diff
@@ -1,6 +1,6 @@
 {% extends "_base_page.html" %}
 
-{% block page_title %}{{ service_data.serviceName or service_data.lotName }} – Digital Marketplace{% endblock %}
+{% block page_title %}{{ framework.name }} {{ service_data.lotName }} – Digital Marketplace{% endblock %}
```
quis: > ‘Digital Outcomes and Specialists Digital outcomes’ – Digital Marketplace

Yo dawg, I heard you like digital…

app/templates/suppliers/dashboard.html
```diff
@@ -1,5 +1,5 @@
 {% extends "_base_page.html" %}
-{% block page_title %}{{ current_user.supplier_name }} – Digital Marketplace{% endblock %}
+{% block page_title %}Supplier dashboard – Digital Marketplace{% endblock %}
```
quis: We don’t publicly call this a dashboard anywhere. In fact, we alternate between supplier name and ‘Your account’.

image

image

--

‘Your account’ is definitely better than ‘Supplier dashboard’.

app/templates/services/_base_service_page.html
robyoung: Can you think of a better alternative? Maybe just the lot name?

quis: 1. For editing existing services, keep as before.
2. Something generic for new services, like ‘Summary – Digital Marketplace’ or ‘Overview – Digital Marketplace’.

robyoung:
@TheDoubleK the problem is that through Google Analytics someone can tie those requests up with others that don't have the supplier name in the title. The question of not sending the title to Google Analytics is a good one, I'll speak to Ash about how much we depend on the titles for analytics.

app/templates/services/_base_service_page.html
robyoung: Submission or Overview doesn't really say what it's a summary of. What about 'Digital outcomes submission summary - Digital Marketplace'?

robyoung: I've spoken @cathrooney and this sounds ok.

TheDoubleK:
Looks good to me now :+1:

PR #397, opened 2015-12-10
Load custom timestamps on communications

robyoung:
The communications that have been copied over from the G7 bucket have
had timestamps set in a custom metadata field. This change loads those
metadata fields if available.


TheDoubleK:
:+1: as long as Travis goes green when utils version exists.

PR #398, opened 2015-12-11
Breadcrumb bugfix

TheDoubleK:
See: https://www.pivotaltracker.com/story/show/108242580
And: https://www.pivotaltracker.com/story/show/109717458
(They are the same bug)

As well as fixing breadcrumbs after a validation failure - as described in the bugs - this also adds a breadcrumb that was missing for multi-lot services.

Before:
screen shot 2015-12-11 at 12 47 51

After:
screen shot 2015-12-11 at 12 48 18

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -8,7 +8,7 @@ from ..helpers.services import (
     is_service_modifiable, is_service_associated_with_supplier,
     get_signed_document_url, count_unanswered_questions,
 )
-from ..helpers.frameworks import get_framework_and_lot, get_declaration_status
+from ..helpers.frameworks import get_framework_and_lot, get_declaration_status, has_one_service_limit
 
 from dmutils.apiclient import APIError, HTTPError
 from dmutils import s3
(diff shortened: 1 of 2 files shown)


app/main/views/services.py
```diff
@@ -419,6 +419,9 @@ def edit_service_submission(framework_slug, lot_slug, service_id, section_id, qu
         framework=framework,
         service_data=draft,
         service_id=service_id,
+        one_service_limit=[
+            l['oneServiceLimit'] for l in framework['lots'] if l['slug'] == draft['lot']
+            ][0],
```
pcraig3: Is it worth making this a small utility function somewhere? 
It looks a bit finicky.

pcraig3:
Breadcrumb disappears if you try to change the service name to an empty string.
Not sure if we care about that. Seems unlikely anyone gets into that state by themselves.

app/main/views/services.py
quis: Already done here: https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/6c5ea132391d99d464a7dceaee626484a53fd49b/app/main/helpers/frameworks.py#L255

TheDoubleK:
Yep I know about the empty name thing - but at the point you've submitted an empty string for the name then the service doesn't have a name so there's nothing to put in the breadcrumb... I did have a go using the previous name (i.e. the one that you deleted), but that was weird.  
I guess we could "New service" or "New lab" as a default?

app/main/views/services.py
TheDoubleK: OK I'll sort that out.  It was a pure copy-and-paste job from elsewhere that did the same thing I'm afraid.

TheDoubleK:
I think the blank breadcrumb if you submit a blank lab name is OK.  Have switched out that horrible logic to use the helper method everywhere it was needed.

pcraig3:
Just bringing it up so that you knew about it.
Submitting an empty name is never going to work, so I'm not sure how much value there is in keeping the breadcrumb.

PR #399, opened 2015-12-14
Brings in support for empty data fields

tombye:
Needed for this story: https://www.pivotaltracker.com/story/show/109270496

See: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/202


TheDoubleK:
:+1:

PR #402, opened 2015-12-29
Generalise notification message for password resets

pcraig3:
Used to say: 
> If the email address you've entered belongs to a supplier account, we'll send a link to reset the password.

Now, because Administrators might be using it as well, it says: 
> If the email address you've entered belongs to a Digital Marketplace account, we'll send a link to reset the password.

Also refactored the tests a bit.

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/110492418)


robyoung:
:+1:

PR #403, opened 2016-01-05
Remove status link from list of current services

quis:
Before | After
--- | ---
image | image

Having two links in the table was confusing.

This commit has one link, which goes to:
- the live service on the buyer app if the 'edit service' feature flag is off (
  because otherwise we’d be sending you to a pointless page)
- the page where you can turn your service off if the 'edit service' feature
  flag is on

It also removes the pretty colours from the statuses, and changes ‘private’ to
‘unavailable’ to match the admin app. We think there’s still value in
differentiating between services that have been removed by an admin and services
that a supplier has removed themselves.

***

https://www.pivotaltracker.com/story/show/110085882

### DIFF
--- a/app/templates/services/list_services.html
+++ b/app/templates/services/list_services.html
@@ -57,34 +57,25 @@
   ) %}
     {% call summary.row() %}
 
-      {% if 'EDIT_SERVICE_PAGE' is active_feature %}
-        {{ summary.service_link(
-            item.serviceName,
-            url_for('.edit_service', service_id=item.id)
-        ) }}
-      {% else %}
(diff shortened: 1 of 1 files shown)


allait:
It looks like we can get rid of one of the enabled/disabled and just have "Live" and "Removed", so I think we can map both "enabled" and "disabled" to "Remove" for now

quis:
## Map both “enabled” and “disabled” to “removed”

The two states are now just:
- live
- removed

allait:
I've rebased this. Is everyone ok with merging this?

robyoung:
Let's go! :+1:

PR #407, opened 2016-01-15
[#111771084] Only show edit links if framework is open & fix banner message

TheDoubleK:
Fixes this bug: https://www.pivotaltracker.com/story/show/111771084

Before:
screen shot 2016-01-15 at 17 07 37

After:
screen shot 2016-01-15 at 17 08 01

And now also fixes this bug: https://www.pivotaltracker.com/story/show/111770766
Before:
screen shot 2016-01-15 at 17 19 49

After:
screen shot 2016-01-15 at 17 18 15


robyoung:
:+1:

PR #408, opened 2016-01-18
Mock a mandrill call so that tests work offline

pcraig3:
Running our test suite without an internet connection would fail because our `send_email` method wasn't being mocked properly in one of the tests.  
Now it is. 👍


allait:
:+1:

pcraig3:
I am explicitly *not* going to deploy this because I'm going to be opening a new pull request soon with more changes.

PR #410, opened 2016-01-20
[#110062092] Allow a supplier to make one of their services unavailable to buy

pcraig3:
This pull request allows suppliers to remove their live services (although it's still hidden behind a feature flag).

The old behaviour allowed suppliers to change their services between 'live/public/published' and 'private/enabled', although it never went live. 
The new behaviour allows suppliers to make services unavailable to buy, but not to reactivate them. 

As such, the radio buttons that allowed suppliers to change the status of their service have been replaced (in the case of live services) with a button to remove them.
Services that have been removed (either by an administrator or a supplier) now have a notification header (very similar to the one used by the [buyer app](https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/202)), with an email address to get in contact if needed.

The difference between a service that has been removed by an administrator and a service that has been removed by a supplier is still maintained, but totally imperceptible on the front-end.  (Not that we're doing anything with it.)

Refactored quite a bit of code along the way, so it might be more helpful to read it commit by commit.

I've also [redone the functional tests](https://github.com/alphagov/digitalmarketplace-functional-tests/compare/pc_suppliers_suppress_services) where needed to take the various markup changes into account.

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/110062092)

***

### Service list

screen shot 2016-01-20 at 11 07 55

### Edit service (without section editing)
screen shot 2016-01-20 at 11 50 01

### Are you sure you want to remove your service
screen shot 2016-01-25 at 17 13 57

### Service list after removing service
screen shot 2016-01-20 at 11 50 10

### Edit service (that has been removed)

screen shot 2016-01-20 at 11 50 14

***

## Old

- [Service list (this is the same)](https://cloud.githubusercontent.com/assets/2454380/12447672/44242f80-bf69-11e5-85fe-b74420104a84.png)

- [Edit service](https://cloud.githubusercontent.com/assets/2454380/12447678/487ed170-bf69-11e5-9139-69d2a1857407.png)

- [Service list after removing service](https://cloud.githubusercontent.com/assets/2454380/12447684/4c87326c-bf69-11e5-84c6-64261e7716c4.png)

- [Edit service (that has been removed)](https://cloud.githubusercontent.com/assets/2454380/12447687/50a7ecc4-bf69-11e5-8a9f-5e2c46c14ec9.png)


TheDoubleK:
In other places with "destructive" actions like this (e.g. delete a draft service) we have the "are you sure?" button in a red message at the top - do we want that for removing a service?

TheDoubleK:
:boom: 
screen shot 2016-01-20 at 14 05 36
:boom:

app/main/views/services.py
```diff
@@ -40,69 +33,58 @@ def list_services():
...
-            "Sorry, but '{}' is not a valid status.".format(status)
-        )
-
-    try:
-        updated_service = data_api_client.update_service_status(
-            service.get('id'), status,
-            current_user.email_address)
+    updated_service = data_api_client.update_service_status(
+        service.get('id'),
+        'enabled',
```
TheDoubleK: Changing status to `enabled` in order to disable the service... how did we get to this point again? :)

Might it be worth going all the way to `disabled` if we're treating both states the same, just for clarity?  Or do we still want to maintain the distinction between services turned off by an admin and services turned off by the supplier?

app/templates/services/service.html
```diff
@@ -21,14 +21,33 @@
...
-  {% if service_data.status == "published" %}
-    <div class="column-one-third">
-      {% with label="View service", url="/g-cloud/services/" + service_id %}
-        {% include "toolkit/link-button.html" %}
+{% block before_heading %}
+  <div class="column-one-whole">
+    {% if service_data.status != 'published' and service_unavailability_information %}
+      {%
+        with
+        message = "If don’t know why this service was removed or you want to reinstate it, contact <a href='mailto:enquiries@digitalmarketplace.service.gov.uk'>enquiries@digitalmarketplace.service.gov.uk</a>.",
```
TheDoubleK: Missing a word here - should be "If *you* don't know why..."

app/templates/services/_edit-status.html
```diff
@@ -1,18 +1,22 @@
...
-  </form>
-{% endif %}
-{% if service_data.status == 'disabled' %}
-  <div class="question">
-    <h2 class="question-heading service-status-removed">
-      This service has been removed
-    </h2>
-    <p class="hint">
-      If you don't know why this service has been removed, contact
+{% if service_data.status == 'published' %}
```
TheDoubleK: This file could probably be renamed from `_edit-status.html` to something along the lines of `_remove-service.html`

app/main/views/services.py
pcraig3: We've used different names at different points.

| Database  | Interface (before) | Interface (now)
| ------------- | ------------- | ------------- |
| `published`  | Public | Live |
| `enabled`  | Private | Removed |
| `disabled`  | Removed | Removed |

Essentially, whenever we've had three values, two have been inverses of each other and the third one is an outlier. published + (enabled/disabled), and then (public/private) + removed.  Now we don't, which is nice.

Keeping the three statuses lets us see at a glance who turned off a service, so maybe that's worth keeping. Functionally, we don't do anything differently though.

robyoung: We had another discussion about this a week or two ago and the conclusion was that we only need two statuses one where the service is live and one where it is not. I don't think this actually turned into a plan about how we get to there though.

cc/ @cathrooney @allait

pcraig3:
After @superroz looks at the error messages, this is good to go :smiley:

TheDoubleK:
Looks good to me now :+1:

PR #411, opened 2016-01-20
fix failing 503 test

andrewchong:


allait:
I think I know why this was passing for everyone and Travis before. While the mock is clearly wrong, since there's no API running on Travis, the real `data_api_client` was returning the expected 503 after failing to connect to the API. `./scripts/run_tests.sh` doesn't set the Data API URL variable, so the test would fail to connect to the API even if it's running locally on the standard port.
But, if you launch the API and export the API URL variable for the supplier-frontend tests (which is what Andrew did following our README) the client connects to the API and returns 404 instead of 503.

:+1:, happy we caught this, but we should think about what we can change in the apiclient so that it never makes actual requests during test runs.

PR #412, opened 2016-01-20
Add agreement bucket to `run_app.sh`

pcraig3:
Running the app without `DM_AGREEMENTS_BUCKET` set causes the app to fail in various places (such as [the frameworks dashboard](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/main/views/frameworks.py#L91)).
Setting the variable resolves the issue.

### DIFF
--- a/scripts/run_app.sh
+++ b/scripts/run_app.sh
@@ -13,6 +13,7 @@ export DM_API_AUTH_TOKEN=${DM_API_AUTH_TOKEN:=myToken}
 
 export DM_SUBMISSIONS_BUCKET=${DM_SUBMISSIONS_BUCKET:=digitalmarketplace-documents-dev-dev}
 export DM_COMMUNICATIONS_BUCKET=${DM_COMMUNICATIONS_BUCKET:=digitalmarketplace-documents-dev-dev}
+export DM_AGREEMENTS_BUCKET=${DM_AGREEMENTS_BUCKET:=digitalmarketplace-documents-dev-dev}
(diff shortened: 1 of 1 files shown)


scripts/run_app.sh
```diff
@@ -13,6 +13,7 @@ export DM_API_AUTH_TOKEN=${DM_API_AUTH_TOKEN:=myToken}
 
 export DM_SUBMISSIONS_BUCKET=${DM_SUBMISSIONS_BUCKET:=digitalmarketplace-documents-dev-dev}
 export DM_COMMUNICATIONS_BUCKET=${DM_COMMUNICATIONS_BUCKET:=digitalmarketplace-documents-dev-dev}
+export DM_AGREEMENTS_BUCKET=${DM_AGREEMENTS_BUCKET:=digitalmarketplace-documents-dev-dev}
```
TheDoubleK: This bucket currently only has stuff in from the functional tests.  Dumping stuff here is probably OK though, it shouldn't affect test runs.

The bucket `digitalmarketplace-dev-documents` seems to have been the preferred dumping ground for the rest of random stuff from running things locally up until now though.

@allait and/or @robyoung might have an opinion which is better? Although it probably doesn't matter either way really.

allait: I haven't touched this in a while, but I think we used `dev-dev` because S3 uploader expected a bucket name format it could parse. If that changed we could probably use something different.
I don't think the name matters that much, but we should just choose one (or few, if we have different use cases) and delete all other dev buckets from S3 at some point.

TheDoubleK:
:+1: after a rebase.

PR #413, opened 2016-01-20
Only copy the frameworks folder to app/content

allait:
When copying the digitalmarketplace-frameworks contents from bower
components to app/content the build process used to copy all repo
files. Since we've added more non-content related files to the
repo (tests, scripts, etc.) that aren't used by the apps there's
no reason to add them to the app release package and they could be
causing issues in the future (eg being picked up by pep8 and pytest).


TheDoubleK:
Good stuff :+1:

PR #414, opened 2016-01-22
[#111536136] Remove update user endpoint and user promotions

allait:
### Remove update-account tests and update test assertions
Splits the user invite tests into sending the invitation and
creating the account. Adds the new expectation of 400 response
for existing accounts and removes the update-user tests since
the endpoint shouldn't be used anymore.

Also removes all the extra `with app.context` wrappers.

### Remove /update-user endpoint
Since we're no longer promoting old buyer acounts to supplier accounts
the update-user endpoint can be removed completely.

### Return 400 when creating a user that already exists
API returns 409 when trying to create a user account with an existing
email address. Since we don't have an error handler for 409, we return
400 instead.

`create-user.html` doesn't seem to handle the flash messages, so they're
removed from the view. Also `invalid_token` parameter seems to be unused.

### Return "account already exists" message if the user already exists
Removes account type change from buyer to supplier.

Removes the checks for logged in users: if the account already
exists there's nothing we can change about it, so it doesn't matter
if the user is already logged in or not. `/create-user` page won't
log you in after the initial registration, even if you're using the
same token you've used to register, since it's not a general "log in"
route - otherwise it's possible to use it to log in without knowing
the account password.

Changes messages for existing buyer and admin accounts to make
it possible to reuse error the template for buyer registrations.

### User create-user-error template for all error responses
Removes duplicate invalid token message from create-user template.

### For all existing accounts (including admin, buyer and accounts from the same supplier)
screen shot 2016-01-22 at 16 08 42

### For accounts with a different supplier_id
screen shot 2016-01-22 at 16 10 33

### Token error and repeated POST request 
screen shot 2016-01-22 at 16 07 23
error

### Locked account
screen shot 2016-01-22 at 16 15 30

### Deactivated account
screen shot 2016-01-22 at 16 15 02


app/templates/auth/create-user-error.html
```diff
@@ -0,0 +1,63 @@
...
+</p>
+
+{% elif user %}
+
+  {% if user.role == 'supplier' and token.get("supplier_id") != user.supplier_id %}
+
+  <header class="page-heading-smaller">
+    <h1>The details you provided are registered with another supplier.</h1>
+  </header>
+  <p>You were invited by ‘{{ token.get("supplier_name") }}’</p>
```
TheDoubleK: These two sentences are missing full stops (`You were invited by...` and `Your account is registered with...`).

app/templates/auth/create-user-error.html
```diff
@@ -0,0 +1,63 @@
...
+  </p>
+
+  {% else %}
+
+  <header class="page-heading-smaller">
+    <h1>Account already exists</h1>
+  </header>
+
+  <p>
+    <a href="{{ url_for('.render_login') }}">Log in</a> to your account.
```
TheDoubleK: An odd thing here is if you click this link while logged in with a different account from the one you're trying to create then you get taken straight to the supplier dashboard as the logged-in user, with no chance to swap to the new user account.

But this is an edge case only likely to be encountered by one of us testing login stuff, so fine I think.

TheDoubleK:
Reviewed and tested locally - :+1:  once those full stops are added.

PR #415, opened 2016-02-01
[#111771966] Change layout of submitted services

robyoung:
https://www.pivotaltracker.com/story/show/111771966

When a framework is in the pending / standstill state the list of lots that have been submitted to should be displayed differently.

## Before
screen shot 2016-02-01 at 17 10 40

## After
screen shot 2016-02-02 at 17 23 36

### DIFF
--- a/app/assets/scss/_lists.scss
+++ b/app/assets/scss/_lists.scss
@@ -27,6 +27,9 @@ ol {
 .list-bullet,
 .list-number {
   margin-top: 5px;
+}
+
+.list-number {
   margin-bottom: 20px;
 }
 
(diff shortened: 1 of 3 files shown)


pcraig3:
Looks dope :+1:

ralph-hawkins:
The appearance of the `View services` link is broken. The underline is about 10px too far below the word. We have custom underlines for `browse-list` links, so it something to do with that. It also looks like the spacing above `View services`is too small, which is probably related.

ralph-hawkins:
:+: (long-term we need to look at how this page works through all the stages of a framework)

PR #416, opened 2016-02-02
Enable suppliers removing their services

pcraig3:
@cathrooney asked for the feature (suppliers' ability to remove their own services) to go live, so here it goes.


TheDoubleK:
Cripes!
:+1: 

Although I'd be tempted to leave the old dates in, just so we can remind ourselves from time to time that it took 11 months to ship this feature.

pcraig3:
This used to be tied together with Editing Sections which still isn't going to be released, so I'll re-date those ones.

PR #417, opened 2016-02-02
Move login to buyer app

TheDoubleK:
Removes the login and password reset routes, and other associated bits and pieces. These will soon be all in the buyer app instead.

Depends on:
 - [x] https://github.com/alphagov/digitalmarketplace-utils/pull/235


pcraig3:
screen shot 2016-02-05 at 12 19 41

tumblr_n2o7arveuu1rewusno1_500

app/main/helpers/__init__.py
```diff
@@ -8,3 +11,13 @@ def hash_email(email):
     m.update(email.encode('utf-8'))
 
     return base64.urlsafe_b64encode(m.digest())
+
+
+def supplier_role_required(func):
```
allait: This is true whenever we require a logged in user, I'd just wrap `@login_required` with additional role check, give it the same `login_required` name and then fix the imports – just to make sure we're never missing role checks.

app/main/errors.py
```diff
@@ -1,4 +1,9 @@
-from flask import render_template
+import sys
+if sys.version_info[0] < 3:
```
allait: We usually just `try: ... except ImportError: ...` instead of checking version

app/main/helpers/__init__.py
TheDoubleK: I did it that way to start with but then decided to stick with the Flask `login_required`.  But yep, it's probably nicer to just have the one decorator.

My main worry was that it could be confusing once we also have `login_required` routes in the buyer app they will be calling different methods.

pcraig3:
:+1: :dinosaur:

PR #418, opened 2016-02-03
Change uploaded document filenames and set a Content-Disposition header

pcraig3:
Because including the supplier names in the filenames for
uploaded documents is quite finicky, we've opted to keep only the
`supplier_id` in the filename proper but suggest a different filename
(which includes the supplier_name) when the file is downloaded.
By setting a `Content-Disposition` header on the uploaded file, we:

- force the browser to download it
- suggest a different filename in the 'Save As' dialog

Note that since we use the same dmutils function for both viewing
and downloading files, and since all existing files are tied to the
old naming convention, we have to update the supplier uploading method
separately from the supplier downloading method.
Once existing files have been renamed, we can port the change over to
dmutils.

### As a tangible example.

If the old filename was:
`Supplier_Name-10101-framework-agreement.pdf`

The new filename will be:
`10101-framework-agreement.pdf`

But when you download it, the filename suggested will be:
`Supplier_Name-10101-framework-agreement.pdf`

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -14,9 +14,9 @@ from dmutils.formats import format_service_price, datetimeformat
 from dmutils import s3
 from dmutils.deprecation import deprecated
 from dmutils.documents import (
-    get_agreement_document_path, get_signed_url, get_extension,
-    file_is_less_than_5mb, file_is_empty,
-    get_countersigned_agreement_document_path)
+    get_agreement_document_path, get_signed_url, get_extension, file_is_less_than_5mb, file_is_empty,
+    get_countersigned_agreement_document_path, sanitise_supplier_name
+)
 
 from ... import data_api_client
 from ...main import main, content_loader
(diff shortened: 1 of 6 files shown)


app/main/views/frameworks.py
```diff
@@ -540,12 +535,24 @@ def upload_framework_agreement(framework_slug):
...
 
     agreements_bucket = s3.S3(current_app.config['DM_AGREEMENTS_BUCKET'])
-    legal_supplier_name = supplier_framework['declaration']['SQ1-1a']
     extension = get_extension(request.files['agreement'].filename)
-    path = get_agreement_document_path(
-        framework_slug, current_user.supplier_id, legal_supplier_name,
-        'signed-framework-agreement{}'.format(extension))
-    agreements_bucket.save(path, request.files['agreement'], acl='private')
+
+    # TODO: get rid of this; rewrite `.get_agreement_document_path()` function in dmutils
```
robyoung: Why not do the rewrite in `dmutils` and then just depend on it here?

app/main/views/frameworks.py
```diff
@@ -540,12 +535,24 @@ def upload_framework_agreement(framework_slug):
...
+    path = '{0}/agreements/{1}/{1}-signed-framework-agreement{2}'.format(
+        framework_slug,
+        current_user.supplier_id,
+        extension
+    )
+    agreements_bucket.save(
+        path,
+        request.files['agreement'],
+        acl='private',
+        download_filename='{}-{}-signed-framework-agreement{}'.format(
```
robyoung: We should be aware that if the supplier name is changed the agreements files will still be downloaded with the previous supplier name. That may be ok as it's the supplier name at the point that the agreement was uploaded.

allait:
:+1:

PR #419, opened 2016-02-04
Add environment variables for agreements and documents buckets

tombye:
The supplier app uses them but they were absent until now.


robyoung:
:+1:

PR #420, opened 2016-02-04
Build front-end code before running app

tombye:
Brings it into line with Buyer app (see [the same script](https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/master/scripts/run_app.sh#L8) in the buyer app).


TheDoubleK:
:+1:

PR #422, opened 2016-02-05
Don't show the "Sign and return your framework agreement" link if countersigned framework agreement exists

pcraig3:
### Fix if condition to display links properly
Even after judiciously refactoring the /frameworks/dashboard page, it
turned out that there was a stray `if` condition checking for a
template variable that was no longer being passed in.

This meant that, after uploading a signed framework agreement and
receiving a countersigned one back, that it would still be possible
to upload a *new* signed framework agreement -- anarchy, basically.

This bugfix removes that possibility.


TheDoubleK:
anarchy

TheDoubleK:
:+1:

PR #423, opened 2016-02-05
Add makefile

robyoung:
## Move default config to config.py
This way there is one place where config defaults are managed. They can still be overridden with environment variables.

## Add a Makefile
This follows the lead taken in `digitalmarketplace-api`

### DIFF
--- a/scripts/run_tests.sh
+++ b/scripts/run_tests.sh
@@ -5,12 +5,6 @@
 # NOTE: This script expects to be run from the project root with
 # ./scripts/run_tests.sh
 
-# Use default environment vars for localhost if not already set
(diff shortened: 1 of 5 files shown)


scripts/run_tests.sh
```diff
@@ -5,12 +5,6 @@
 # NOTE: This script expects to be run from the project root with
 # ./scripts/run_tests.sh
 
-# Use default environment vars for localhost if not already set
-. ./scripts/test_env.sh
-
-echo "Environment variables in use:"
-env | grep DM_
```
allait: This could still be useful to show overwritten variables

allait:
:+1:

PR #424, opened 2016-02-08
Drive run scripts from the Makefile

robyoung:
Now that we have a Makefile that captures these tasks these scripts
should be driven by that. These scripts are being left here for people
that are using them.

### DIFF
--- a/scripts/run_app.sh
+++ b/scripts/run_app.sh
@@ -1,13 +1,3 @@
 #!/bin/bash
 
-if [ -n "$VIRTUAL_ENV" ]; then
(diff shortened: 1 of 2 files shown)


TheDoubleK:
Not an issue with this PR, but the Makefile doesn't seem to be outputting things quite right...
```
$ ./scripts/run_tests.sh
Environment variables in use:
[ -z $VIRTUAL_ENV ] && [ ! -d venv ] && virtualenv venv || true
```

TheDoubleK:
Tests run and app runs :+1:

robyoung:
That's because there are no environment variables in use.

PR #426, opened 2016-02-10
Suppliers can view/remove G-Cloud 7 services

pcraig3:
Sorry for all the re-indenting.  This is essentially a one-line fix for suppliers viewing/removing G-Cloud 7 services.

### 	Load g-cloud-7 service editing manifest
Manifests are loaded initially and then retrieved when they're needed.
This means they'll fail if one tries to retrieve a manifest that wasn't loaded.
In this case, the `edit_service` manifest for G-Cloud 7 wasn't loaded, so if any supplier tried to view/remove one of their G7 services, they would get a 500-level (ContentNotFound) error.
Adding the manifest solves the problem.


allait:
Can we try this http://pytest.org/latest/fixture.html#parametrizing-a-fixture instead of loops within each test?

Otherwise, when g-cloud-6 goes away we'd have to update all those tests again.

pcraig3:
Yeah, that looks much better.

robyoung:
:+1:

PR #427, opened 2016-02-12
Reflect status code in errorhandler name

pcraig3:
Our 500-level errorhandler had a generic name instead of the standard 'Internal Server Error'.
Fixed it in the buyer app, so fixing it here as well.


TheDoubleK:
:+1:

PR #429, opened 2016-02-25
Submit brief clarification questions

allait:
Adds a form to submit brief clarification questions.

* Page is only available if brief is currently live.
* The check for supplier eligibility is not connected to the API so
  will always accept the supplier until the filtering is implemented.
* There's no separate "question sent" page, the message is shown on
  the page with the form right now.
* Email sent to the buyer user is sent from "do-not-reply@" with
  anonymised supplier name.
* Failure to send a confirmation to the supplier is ignored. So as long
  as the buyer email was accepted by Mandrill the app will respond with
  200.

:exclamation: This needs some content work on the error/success pages and email templates.

screen shot 2016-02-25 at 17 15 07

### DIFF
--- a/app/main/helpers/briefs.py
+++ b/app/main/helpers/briefs.py
@@ -1,0 +1,72 @@
+import six
+
+from flask import abort, current_app, render_template
+from flask_login import current_user
+
+from dmutils.email import send_email, hash_email, MandrillException
+
+
+def get_brief(data_api_client, brief_id, live_only=True):
+    brief = data_api_client.get_brief(brief_id)['briefs']
+    if live_only and brief['status'] != 'live':
+        abort(404)
+
+    return brief
+
+
+def check_supplier_is_eligible_for_brief(brief, supplier_id):
+    # TODO connect this with the API endpoint once it exists
+    # Should abort or render an error page if the check fails
+    pass
+
+
+def send_brief_clarification_question(brief, clarification_question):
+    # Email the question to brief owners
+    email_body = render_template(
+        "emails/brief_clarification_question.html",
+        brief_name=brief['title'],
+        message=clarification_question
+    )
+    try:
+        send_email(
+            to_email_addresses=get_brief_user_emails(brief),
+            email_body=email_body,
+            api_key=current_app.config['DM_MANDRILL_API_KEY'],
+            subject="{} clarification question".format(brief['title']),
+            from_email=current_app.config['CLARIFICATION_EMAIL_FROM'],
+            from_name="{} Supplier".format(brief['frameworkName']),
+            tags=["brief-clarification-question"]
+        )
+    except MandrillException as e:
+        current_app.logger.error(
+            "Brief question email failed to send. error={error} supplier_id={supplier_id} brief_id={brief_id}",
+            extra={'error': six.text_type(e), 'supplier_id': current_user.supplier_id, 'brief_id': brief['id']}
+        )
+
+        abort(503, "Clarification question email failed to send")
+
+    # Send the supplier a copy of the question
+    supplier_email_body = render_template(
+        "emails/brief_clarification_question_confirmation.html",
+        brief_name=brief['title'],
+        message=clarification_question
+    )
+    try:
+        send_email(
+            to_email_addresses=[current_user.email_address],
+            email_body=supplier_email_body,
+            api_key=current_app.config['DM_MANDRILL_API_KEY'],
+            subject="Your {} clarification question".format(brief['title']),
+            from_email=current_app.config['CLARIFICATION_EMAIL_FROM'],
+            from_name=current_app.config['CLARIFICATION_EMAIL_NAME'],
+            tags=["brief-clarification-question-confirmation"]
+        )
+    except MandrillException as e:
+        current_app.logger.error(
+            "Brief question supplier email failed to send. error={error} supplier_id={supplier_id} brief_id={brief_id}",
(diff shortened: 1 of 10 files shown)


app/main/views/briefs.py
```diff
@@ -0,0 +1,41 @@
...
+    check_supplier_is_eligible_for_brief(brief, current_user.supplier_id)
+
+    error_message = None
+    clarification_question_value = None
+
+    if request.method == 'POST':
+        clarification_question = request.form.get('clarification-question', '').strip()
+        if not clarification_question:
+            error_message = "Question cannot be empty"
+        elif len(clarification_question) > 5000:
```
robyoung: On the API side there is a 100 word limit and no character limit. We should decide what the limits should be and use the same.

app/main/views/briefs.py
```diff
@@ -0,0 +1,41 @@
...
+from ..helpers.briefs import (
+    get_brief,
+    check_supplier_is_eligible_for_brief,
+    send_brief_clarification_question,
+)
+from ...main import main
+from ... import data_api_client
+
+
+@main.route('/opportunities/<int:brief_id>/ask-a-question', methods=['GET', 'POST'])
```
robyoung: I like `ask-a-question`. Maybe just `answer-a-question` on the buyer side?

app/main/helpers/briefs.py
```diff
@@ -0,0 +1,72 @@
...
+
+def get_brief(data_api_client, brief_id, live_only=True):
+    brief = data_api_client.get_brief(brief_id)['briefs']
+    if live_only and brief['status'] != 'live':
+        abort(404)
+
+    return brief
+
+
+def check_supplier_is_eligible_for_brief(brief, supplier_id):
```
robyoung: :warning: really nitpicky semantics :warning: 
Based on how this is used I think this would be better named something like `ensure_supplier_is_eligible_for_brief`. Check implies to me that the fact is being checked but not necessarily acted upon.

app/main/views/briefs.py
allait: Oh, I forgot to make a note about limits. I've mentioned this to Ralph briefly, but I agree we need to decide what we should do.

I think we need an upper length limit to protect buyers from malicious emails. I'm not enforcing a word limit (even though there's a counter on the form) to avoid implementing it again before we're sure there's a need for it. 5000 character limit is what we used for framework clarification questions.

robyoung: OK, I'm going to add the 5000 character limit to the JSON schema as well then. I wish there was a 'resolved' button like in Google Docs.

robyoung:
:+1:

app/main/helpers/briefs.py
```diff
@@ -0,0 +1,72 @@
+import six
+
+from flask import abort, current_app, render_template
+from flask_login import current_user
+
+from dmutils.email import send_email, hash_email, MandrillException
+
+
+def get_brief(data_api_client, brief_id, live_only=True):
```
pcraig3: [Looks familiar](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/pc_supplier_reply_responses/app/main/helpers/briefs.py#L16).

app/templates/briefs/clarification_question.html
```diff
@@ -0,0 +1,99 @@
...
+        }
+    ],
+      lede = "There was a problem with your submitted question"
+    %}
+      {% include "toolkit/forms/validation.html" %}
+    {% endwith %}
+  {% endif %}
+
+  {% with
+    heading = "Ask a question about {}".format(brief.title),
```
pcraig3: We're using [smart curly Ralph single quotes](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/pc_supplier_reply_responses/app/main/views/briefs.py#L40) in the title for the supplier's brief response.

app/templates/briefs/clarification_question.html
```diff
@@ -0,0 +1,99 @@
...
+          name = clarification_question_name,
+          question = "Ask a question",
+          error = error_message,
+          value = clarification_question_value,
+          max_length_in_words = 100
+        %}
+          {% include "toolkit/forms/textbox.html" %}
+        {% endwith %}
+        <p>
+        </p>
```
pcraig3: `<p></p>`

tests/app/main/test_briefs.py
```diff
@@ -0,0 +1,134 @@
...
+class TestSubmitClarificationQuestions(BaseApplicationTest):
+    def test_submit_clarification_question_requires_login(self, data_api_client):
+        res = self.client.post('/suppliers/opportunities/1/ask-a-question')
+        assert res.status_code == 302
+        assert '/login' in res.headers['Location']
+
+    @mock.patch('app.main.helpers.briefs.send_email')
+    def test_submit_clarification_question(self, send_email, data_api_client):
+        self.login()
+        data_api_client.get_brief.return_value = {'briefs': {
```
pcraig3: Kev was using [some generated brief from dmutils](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/pc_supplier_reply_responses/tests/app/main/test_briefs.py#L46) for tests.  Not sure if it's easy to change  / worth changing your existing stuff (but it might be nice to be consistent).

pcraig3:
Oops, didn't notice @robyoung's upward-pointing thumbs. My comments aren't really critical, though it might be nice to figure out this quotes situation.

allait:
@pcraig3 I'm updating content on this right now, so comments are still welcome. Good catch on empty paragraph and API stub, will fix.

PR #430, opened 2016-02-29
Removed deprecated redirect for file download

pcraig3:
@kevkeenoy and I noticed these deprecated redirects, so we heroically deleted them and an associated test.

Seems fairly unlikely that anyone's still using them.

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -317,22 +317,6 @@ def download_agreement_file(framework_slug, document_name):
     return redirect(url)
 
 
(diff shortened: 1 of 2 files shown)


robyoung:
Is it worth doing a quick check in the logs to verify that they're not being used?

TheDoubleK:
I've searched the logs for both .zip and .pdf routes here and neither has been hit this year  - searching from 01/01/2016 onwards with the following patterns:
```
{ $.request = "*/suppliers/frameworks/g-cloud-7*zip" }
{ $.request = "*/suppliers/frameworks/g-cloud-7*pdf" }
```
screen shot 2016-02-29 at 14 46 09

robyoung:
:+1:

PR #431, opened 2016-02-29
[#112216471] Supplier responds to a brief -- frontend

pcraig3:
This story allows suppliers to submit responses to a briefs/opportunities (hereafter referred to as a 'brief').
All of the questions for a brief response are on one page; some questions are stored entirely in the content but others are a mixture of static content and dynamic (user-submitted) essential/nice to have questions from the brief itself.

Suppliers can only submit one response per brief. (If they try to submit another, they get a 404 page -- in future we might want to redirect them to another page and display a flash message.)

It *does not* check if suppliers are eligible to submit responses to a particular brief (this depends on the [only eligible suppliers can respond to a brief](https://www.pivotaltracker.com/story/show/112216509) story)

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/112216471)

#### boolean_list macro
Introduced a new [`boolean_list`](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/pc_supplier_reply_responses/app/templates/macros/toolkit_forms.html#L153) macro which receives a section object injected full of boolean_questions pulled from the buyer brief.  Loops through the boolean questions, calling the `boolean` macro for each one.

#### convert_to_boolean filter
Introduced a new jinja2 filter, [`convert_to_boolean`](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/pc_supplier_reply_responses/app/__init__.py#L78).  Logic is imported directly from dmutils.  We're using it to transform `"true"` and `"false"` strings to `True` and `False`  boolean values so that we can preserve submitted boolean questions if validation fails.

#### text macro
Updated the [`text`](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/pc_supplier_reply_responses/app/templates/macros/toolkit_forms.html#L1) form field macro to support smaller textboxes and textboxes with prices.

#### per-question validation
`boolean_list` questions are a kind of multiquestion in that they have nested questions, but, unlike in a multiquestion, all boolean questions must be answered for a `boolean_list` question to be valid.  @tombye pointed out that if a particular question wasn't answered, the error message should point to that question, not the entire boolean_list group.  [Wrote a big, fat loop to implements this](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/pc_supplier_reply_responses/app/main/views/briefs.py#L83).

#### inject brief questions into ContentQuestions
Since we're displaying all of the questions on one page, we decided to [put the buyer-submitted question
data into the `ContentSection` object](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/pc_supplier_reply_responses/app/main/helpers/briefs.py#L25).  It's not ideal, as there's a danger of potentially overwriting other keys unintentionally, as well as the fact that we don't really want the `ContentSection`
object to be any more complicated, but doing it this way allows us to keep our HTML template totally generic and avoid adding anything additional to the toolkit.

#### inject brief name into ContentSection
[Overwrite the default section name to include the name of the brief](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/pc_supplier_reply_responses/app/main/views/briefs.py#L40).  If a supplier is applying to lots of briefs, it might be helpful.

***

### brief response page
screen shot 2016-02-29 at 16 27 37

### brief response page (validation errors)
screen shot 2016-02-29 at 16 28 13

### brief response submitted
screen shot 2016-02-29 at 16 28 48

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -7,7 +7,6 @@ from flask_wtf.csrf import CsrfProtect
 import dmapiclient
 from dmutils import init_app, flask_featureflags
 from dmutils.user import User
-from dmutils.config import convert_to_boolean
 
 from config import configs
 
@@ -74,10 +73,6 @@ def create_app(config_name):
 
     application.add_template_filter(question_references)
     application.add_template_filter(parse_document_upload_time)
-
-    @application.template_filter('convert_to_boolean')
-    def convert_to_boolean_filter(data):
-        return convert_to_boolean(data)
 
     return application
 
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -7,6 +7,7 @@ from flask_wtf.csrf import CsrfProtect
 import dmapiclient
 from dmutils import init_app, flask_featureflags
 from dmutils.user import User
+from dmutils.config import convert_to_boolean
 
 from config import configs
 
@@ -73,6 +74,10 @@ def create_app(config_name):
 
     application.add_template_filter(question_references)
     application.add_template_filter(parse_document_upload_time)
+
+    @application.template_filter('convert_to_boolean')
+    def convert_to_boolean_filter(data):
+        return convert_to_boolean(data)
 
     return application
 
--- a/app/main/helpers/briefs.py
+++ b/app/main/helpers/briefs.py
@@ -6,12 +6,16 @@ import datetime
 from flask import abort, current_app, render_template
 from flask_login import current_user
 
-from dmutils.email import send_email, hash_email, MandrillException
+from dmutils.email import send_email, MandrillException
 
 
-def get_brief(data_api_client, brief_id, live_only=True):
+def get_brief(data_api_client, brief_id, allowed_statuses=None):
+    if allowed_statuses is None:
+        allowed_statuses = []
+
     brief = data_api_client.get_brief(brief_id)['briefs']
-    if live_only and brief['status'] != 'live':
+
+    if allowed_statuses and brief['status'] not in allowed_statuses:
         abort(404)
 
     return brief
@@ -80,3 +84,11 @@ def send_brief_clarification_question(brief, clarification_question):
 
 def get_brief_user_emails(brief):
     return [user['emailAddress'] for user in brief['users'] if user['active']]
+
+
+def inject_yes_no_questions_into_section_questions(section, brief):
+
+    # pass all of the brief yes/no questions into the appropriate ContentQuestions
+    for question in section.questions:
+        if question.type == 'boolean_list':
+            question.boolean_list_questions = brief[question.id]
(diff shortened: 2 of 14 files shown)


app/__init__.py
```diff
@@ -74,6 +75,10 @@ def refresh_session():
     application.add_template_filter(question_references)
     application.add_template_filter(parse_document_upload_time)
 
+    @application.template_filter('convert_to_boolean')
+    def convert_to_boolean_filter(data):
+        return convert_to_boolean(data)
```
robyoung: Can this just be `application.add_template_filter(convert_to_boolean)` like the ones above?

app/main/helpers/briefs.py
```diff
@@ -0,0 +1,30 @@
...
+
+    return brief
+
+
+def inject_yes_no_questions_into_section_questions(section, brief):
+
+    # pass all of the brief yes/no questions into the appropriate ContentQuestions
+    for question in section.questions:
+        if question.type == 'boolean_list':
+            question.boolean_list_questions = brief[question.id]
```
robyoung: This feels like it should be handled by `ContentQuestion`.

app/main/views/briefs.py
```diff
@@ -0,0 +1,123 @@
...
+                        value_we_want_to_be_true_or_false = None
+
+                    if not isinstance(value_we_want_to_be_true_or_false, bool):
+                        # Each non-boolean value is an error
+                        boolean_question_id = "{}-{}".format(question_id, index)
+                        new_errors[boolean_question_id] = {
+                            'input_name': boolean_question_id,
+                            'message': error_message['message'],
+                            'question': essential_question
+                        }
```
robyoung: This feels like it should be handled in `dmutils.content_loader` somewhere.

TheDoubleK: Rlly?

app/main/helpers/briefs.py
pcraig3: Brought it up with @allait and it doesn't really seem like we would get a lot from this at this juncture.  
This feels really particular to this particular field in this particular app, and it's also bound pretty tightly with the `boolean_list` macro (which looks for the `boolean_list_questions` attribute).

Passing a brief into a section also seems a bit funny.

Basically, we only need this for one field and we're only using this field in one place, and it feels like moving this code doesn't give us a useful generalisation.

app/main/views/briefs.py
pcraig3: Gonna refactor the `ContentSummaryQuestion` to get some of this logic in there. 
It's not going to be the ideal solution, but it'll work.

app/__init__.py
pcraig3: Killed it. :skull:

PR #432, opened 2016-02-29
Fix logging and unicode errors

allait:
### Update dmutils to 17.3.3 to fix logging and upload exceptions
Fixes log formatting errors in JSON logger and removes unicode
characters from supplier names to avoid Content-Disposition errors.


robyoung:
:+1:

PR #433, opened 2016-03-02
Don't inject the template data explicitly

TheDoubleK:
This is handled by `init_app` in utils now.

Pages look right locally and tests pass, so I believe this works.


allait:
:+1:

PR #434, opened 2016-03-03
Replace html with frontend tookit pattern in agreements section.

carolinegreen:
Replace html with frontend pattern list in agreements section.  Remove list scss file specific to frontend supplier app as lists are common to digitalmarketplace apps and this should be pulled in from the frontend toolkit.

Relies on - https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/225

Screenshot 1:

screen shot 2016-03-04 at 00 01 58

Screenshot 2:

screen shot 2016-03-03 at 23 59 45


TheDoubleK:
There seems to be a lot of extra space at the bottom of the page before the green button. I'm not sure if this is an issue? I think the old version looks a bit better maybe?

OLD:
screen shot 2016-03-09 at 12 19 25


NEW:
screen shot 2016-03-09 at 12 18 41

app/templates/frameworks/agreement.html
```diff
@@ -52,96 +52,130 @@
...
+                {% include "toolkit/instruction-list.html" %}
+              {% endwith %}
+            {% endif %}
+
+            {%
+            with
+            type = "save",
+            label = "Send document to CCS"
+            %}
+            {% include "toolkit/button.html" %}
```
TheDoubleK: This button "belongs" to the final item in the instruction list really (the choose-a-file element), so it seems odd to have the button completely removed from that.  I think this is where the extra space is coming from too.

carolinegreen: The extra space has been put in intentionally to the supplier app in both these commits:

https://github.com/alphagov/digitalmarketplace-supplier-frontend/commit/83bbd1b5c5fb276f9baa8e03c333b9e021ae4c51

https://github.com/alphagov/digitalmarketplace-supplier-frontend/commit/521c7a4ba36fc34866fb8dd393840e9521d77e58

TheDoubleK: But they're quite old commits - the spacing on the page changes with these new changes.

carolinegreen: Well if the spacing should be changed it should probably be changed uniformly, so I could just remove it from the css in the supplier app.

carolinegreen: screen shot 2016-03-09 at 14 15 38

carolinegreen:
The list itself has now had padding removed from the bottom and the margin above the form button is being kept in for the time being.  

This now relies on:
https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/229

pcraig3:
Bit funny [how the list style does numbers on the demo page](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/instruction-list.html), but doesn't really seem like an issue for our real-life code.

Looks better than before :+1:

PR #437, opened 2016-03-09
[#113081517] Two new footer links

TheDoubleK:
See: https://www.pivotaltracker.com/story/show/113081517

New links added look like this:
screen shot 2016-03-09 at 17 07 17


tombye:
Nice work with the link localising :+1:

PR #439, opened 2016-03-15
Remove custom version of gulp-jasmine-phantom

tombye:
We made a custom version to deal with the fact
that the version of Node that Travis uses in the VM
image for Python projects was below that required
for the NPM module (see related [stackoverflow question](http://stackoverflow.com/questions/32461456/travis-ci-can-the-version-of-node-installed-on-all-vm-images-match-the-latest)).

This is no longer the case (it's currently running
at 4.2 (using nvm) so the custom version is no
longer needed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: python
 python:
   - "2.7"
   - "3.4"
+before_install:
+  - nvm install 4.2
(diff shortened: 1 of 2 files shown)


TheDoubleK:
:+1:

PR #441, opened 2016-03-16
Add an audit event for sent brief clarification questions

allait:
Creates an audit event with the clarification question text when a brief question is submitted.

Reuses the existing audit event type from framework clarification questions. Since the related object type is different it is still simple to distinguish the clarification question types when listing events.


TheDoubleK:
Yep we want to keep these questions somewhere for sure :+1:

PR #442, opened 2016-03-17
Use toolkit list styles for frameworks dashboard

pcraig3:
Makes use of the new `list-no-bullet` class instead of putting a classname on individual `li` tags.

Previous `unbulleted-item` and `guidance-link` classes were both doing essentially the same thing.

Also brings in the styles we need for [the error page we're building in this branch](https://github.com/alphagov/digitalmarketplace-supplier-frontend/tree/ensure-eligible-suppliers).

Also also changes some of the templates to snake_case so as to be consistent.


robyoung:
:snake: :seat: :key: :baby: :+1:

PR #443, opened 2016-03-18
Fix supplier view of submissions

robyoung:
[#113766617](https://www.pivotaltracker.com/story/show/113766617)

Before this change the supplier view of their submissions (the lot selection page) does not make a lot of sense after the framework is no longer open. In the case of Digital Outcomes and Specialists the lot titles were things like 'Apply for Digital Specialists' and link through to add a service of that type. This doesn't make sense after the framework is no longer open for submissions.

This is a bit of a hack but it solving it more correctly does not seem like a priority right now.

## Before
screen shot 2016-03-23 at 11 26 30

## After
Here labs isn't shown because the user didn't create any draft services for it.
screen shot 2016-03-23 at 11 55 21


allait:
:+1:

PR #444, opened 2016-03-23
[#116127657] No clarification questions after deadline

robyoung:
Prevent clarification questions from being submitted after the deadline.
This change also bumps the apiclient to get the latest API stub for briefs which includes the `clarificationQuestionsAreClosed` flag.

- [x] Depends on https://github.com/alphagov/digitalmarketplace-apiclient/pull/22


allait:
Travis needs apiclient to be merged first, but this looks good :+1:

PR #445, opened 2016-03-29
[#112216509] Only eligible suppliers can respond to a brief

pcraig3:
This pull request throws up an error page if you try to apply for or submit a clarification question for an opportunity that you're not eligible for.  Error pages will tell you what you can't do and if (a) you're not on the framework, or (b) you don't provide the right service.  Hard to be more helpful than that without getting into DOS-specific messaging, which we don't want to do.

@superroz, @ralph-hawkins, and, most recently, @suedavis have all looked at it.

*** 
### Trying to apply, not on the framework
screen shot 2016-03-30 at 11 29 16

### Trying to apply, don't provide the right services
screen shot 2016-03-30 at 11 32 26

### Trying to submit a clarification question, not on the framework
screen shot 2016-03-30 at 11 31 46

### Trying to submit a clarification question, don't provide the right services
screen shot 2016-03-30 at 11 32 45

### DIFF
--- a/app/main/helpers/briefs.py
+++ b/app/main/helpers/briefs.py
@@ -28,10 +28,9 @@ def ensure_supplier_is_eligible_for_brief(brief, supplier_id):
     pass
 
 
-def has_supplier_already_submitted_a_brief_response(data_api_client, supplier_id, brief_id):
-
+def supplier_has_a_brief_response(data_api_client, supplier_id, brief_id):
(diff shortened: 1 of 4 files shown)


ralph-hawkins:
:+1: nice apostrophes :-)

superroz:
Can we replace 'therefore' with 'so’?

pcraig3:
Updated with new words.

superroz:
:+1:

PR #446, opened 2016-03-29
Fix return-to-brief link on clarification page

pcraig3:
Sarah found a bug trying to return to a brief after asking a clarification question.   Basically, the URL was pointing to the wrong place. 

Easy fix.  Didn't write any tests or anything because the page it links to is in another app.  But I've manually tested it works. 

[>> Bug on Pivotal](https://www.pivotaltracker.com/story/show/116545029)


allait:
:+1:

PR #447, opened 2016-03-31
Do not send email from unverified domains

robyoung:
Mandrill is going to start preventing emails from unverified domains.
This means that we will no longer be able to send emails with the From
field set to supplier email addresses. This change sets the From field
to a no-reply address on our domain and uses the Reply-To field for the
supplier email address.


allait:
:+1:

PR #449, opened 2016-04-11
Validate supplier responses

TheDoubleK:
Completes this story: https://www.pivotaltracker.com/story/show/112337127
And addresses this bug: https://www.pivotaltracker.com/story/show/117143511

Content updates and revisions needed and welcomed in comments

For the bug-fix: when suppliers try to create a second response to a brief instead of a 404 page they are now taken to their dashboard with the following flashed message:

screen shot 2016-04-11 at 17 06 37

For the story, if suppliers submit a response which does not have "Yes" to all the essential requirements they are taken to a response result page `/opportunities/<brief_id>/responses/result` that looks like this (more words to come in future):

screen shot 2016-04-11 at 17 05 45

The page has another two modes, neither of which are currently linked to, but which will make sense if users try to hack URLs:
 1. If they haven't yet submitted a response to the brief:
screen shot 2016-04-11 at 17 05 24

 2. If they already submitted a response with all "Yes" to essentials:
screen shot 2016-04-11 at 17 04 49


app/templates/briefs/view_response_result.html
```diff
@@ -0,0 +1,43 @@
...
+    
+    {% elif result_state == "submitted_ok" %}
+      <p>
+        You will hear whether you have been successful or not from the buyer in due course.
+      </p>
+    
+    {% elif result_state == "submitted_unsuccessful" %}
+      <p>
+        Thank you for your response to this opportunity.
+      </p>
```
ralph-hawkins: It'd be good if there was white space between the paragraphs.

app/templates/briefs/view_response_result.html
```diff
@@ -0,0 +1,43 @@
...
+      <p>
```
ralph-hawkins: It'd be good if there was white space between the paragraphs.

ralph-hawkins:
Can we add `secondary action links` taking the user back to page with all the opportunities.

suedavis68:
1. "You've already submitted a response to 'My Outcome' and can't submit another one."

2. "You don't meet all the essential requirements
Thank you for your response to this opportunity
As you don't meet all the essential requirements the buyer can't award you a contract for this work"

3. "You haven't submitted a response to this opportunity"

4. "You'll hear whether you've been successful or not from the buyer soon"

TheDoubleK:
Spacing and content added, should be ready to go now.

superroz:
I've got a few tweaks. Give me a shout if I've changed anything that you've already decided against for some reason.

"You've already **applied** for 'My Outcome' **so you** can't **apply again**."

"You don't meet all the essential requirements.
Thank you for **applying for **this opportunity.
**You** don't meet all the essential requirements **so** the buyer can't **add you to their shortlist**."

"You haven't **applied for** this opportunity."

"**The buyer will tell you** whether you've been successful or **not soon**."

TheDoubleK:
Updated - should be good to go.

app/main/views/briefs.py
```diff
@@ -130,9 +133,36 @@ def create_new_brief_response(brief_id):
...
+        return _render_not_eligible_for_brief_error_page(brief)
 
-    return redirect(url_for(".dashboard"))
+    brief_response = data_api_client.find_brief_responses(
+        brief_id=brief_id,
+        supplier_id=current_user.supplier_id
+    )['briefResponses']
+
+    if len(brief_response) == 0:
+        result_state = 'not_submitted'
```
allait: Why not just redirect the user to the "respond to brief" page? No one without a response should visit this page during a normal flow anyway.

app/templates/briefs/view_response_result.html
```diff
@@ -0,0 +1,52 @@
+{% extends "_base_page.html" %}
+
+{% block page_title %}Your response to ‘{{ brief.title }}’ - Digital Marketplace{% endblock %}
+
+{% block main_content %}
+
+<div class="grid-row">
+  <div class="column-two-thirds">
+    <header class="page-heading-smaller">
```
allait: Is there a reason why we're not using the toolkit template for this http://alphagov.github.io/digitalmarketplace-frontend-toolkit/page-heading.html?

app/main/views/briefs.py
```diff
@@ -67,8 +70,8 @@ def submit_brief_response(brief_id):
         return _render_not_eligible_for_brief_error_page(brief)
 
     if supplier_has_a_brief_response(data_api_client, current_user.supplier_id, brief_id):
-        # TODO redirect to summary of brief response page with flash message
-        abort(404)
+        flash(SUBMIT_SECOND_RESPONSE_ERROR_MESSAGE.format(brief['title']), 'error')
```
allait: So now that we have a result page anyway, can we send the supplier there instead of flashing the message? Otherwise there's no way for them to find out what their result was, since they don't remember applying to begin with.

allait:
Not part of this PR, but I've noticed this during review and I think it's worth fixing: GET method is called `submit_brief_response` and POST one is `create_new_brief_response`. We tend to use `submit_` for POST since it make more sense, and since they refer to the same url they should probably have the same name.

app/main/views/briefs.py
TheDoubleK: Yep that makes sense I guess.

app/templates/briefs/view_response_result.html
TheDoubleK: Because the base page I copied from to make this one doesn't use it either ;)
I'll fix it up.

app/main/views/briefs.py
TheDoubleK: That's also a better idea.  Although for now the result page isn't very helpful.

TheDoubleK:
Addressed all comments and rebased.

Users who have already submitted a response are now redirected to the result page with a flash message if they try to submit a second time.

app/main/views/briefs.py
```diff
@@ -59,16 +59,16 @@ def ask_brief_clarification_question(brief_id):
 
 @main.route('/opportunities/<int:brief_id>/responses/create', methods=['GET'])
 @login_required
-def submit_brief_response(brief_id):
+def brief_response(brief_id):
```
allait: This should still be `create_brief_response`

TheDoubleK: I started with `create_` but then checked our code elsewhere and our most usual pattern, e.g. in https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/main/views/suppliers.py, seems to be to have the `name_of_the_thing()` on the GET route and `submit_name_of_the_thing()` on the POST route everywhere:

allait: Yeah, fair enough. I got the impression we're doing `create_` fro looking at service submissions, but this works too.

allait:
Looks good 👍

PR #450, opened 2016-04-12
Show DOS service lot instead of name for one-lot services

allait:
Since digital-outcomes and other one-service lot servicesdon't have
a serviceName field we need to show the lot name instead, similar
to what we did in the service submission page title.


TheDoubleK:
:+1:

PR #451, opened 2016-04-12
Don't show link for DOS services on the supplier dashboard

allait:
### Drop EDIT_SERVICE_PAGE feature flag
Enabled for 2 months on all environments, should be safe to remove.
The fallback link for service page also assumed a g-cloud service.

### Don't show link for DOS services on the supplier dashboard
Since DOS services don't have an edit service manifest or a public service page on the buyer frontend to link to we don't make the service name a link for now.

There's no check in the view, so the page would 500 if visited, but at the same time there's no obvious way to find the DOS service ID required to manually construct a link.

### DIFF
--- a/tests/app/main/test_services.py
+++ b/tests/app/main/test_services.py
@@ -88,7 +88,7 @@ class TestListServices(BaseApplicationTest):
             assert_equal(res.location, 'http://localhost/login?next=%2Fsuppliers%2Fservices')
 
     @mock.patch('app.main.views.services.data_api_client')
-    def test_shows_services_edit_button_with_id(self, data_api_client):
+    def test_shows_service_edit_link_with_id(self, data_api_client):
         with self.app.test_client():
             self.login()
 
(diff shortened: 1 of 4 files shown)


tests/app/main/test_services.py
```diff
@@ -108,6 +108,48 @@ def test_shows_services_edit_button_with_id(self, data_api_client):
...
+                    'frameworkSlug': 'digital-outcomes-and-specialists'
+                }]
+            }
+
+            res = self.client.get('/suppliers/services')
+            assert_equal(res.status_code, 200)
+            data_api_client.find_services.assert_called_once_with(supplier_id=1234)
+
+            assert "Service name 123" in res.get_data(as_text=True)
+            assert "/suppliers/services/123" not in res.get_data(as_text=True)
```
pcraig3: Should you assert this *is* in the last test?

:space_invader:

pcraig3:
:+1: after rebase and maybe one more assertion

tests/app/main/test_services.py
allait: You mean the test before this one? They both shouldn't have a link, since both are DOS services. I'm only checking this here since the other test has a different purpose.

PR #452, opened 2016-04-12
Validate brief clarification question word limit

allait:
Adds word length validation to clarification questions in addition to the character length check. Word length check is using the same expression we use in JSON schemas.


carolinegreen:
:+1:

TheDoubleK:
It's worth checking with @ralphhawkins what the word and character limits should be - I think there was some talk that 100 might not be enough for complex questions.

allait:
Confirmed 100 words is the limit

PR #453, opened 2016-04-14
Question and answer session details

allait:
### Add question and answer session details page
Adds a page with brief "webinar" details only visible to eligible suppliers. The page displays Q and A session details from the brief.

If the details weren't set the link from the brief page should be hidden, and this view URL will show an empty page with a "Return to brief" link.

### Add CSS rule for .padding-bottom-small
Class is already used by frameworks/agreement.html template and the same class is defined and used in the buyer frontend.

### Fix ask brief clarification question header width
Moves header inside the 2/3 div and fixes the order of closing tags.

#### [#117296475] Update brief clarification question sent confirmation banner

### DIFF
--- a/app/assets/scss/application.scss
+++ b/app/assets/scss/application.scss
@@ -85,3 +85,7 @@ $path: "/suppliers/static/images/";
   @include core-19;
   margin: (1.5 * $gutter) 0 (2 * $gutter);
 }
+
+.padding-bottom-small {
+  padding-bottom: 10px;
+}
--- a/app/templates/briefs/clarification_question.html
+++ b/app/templates/briefs/clarification_question.html
@@ -26,7 +26,7 @@
   {% with messages = get_flashed_messages(with_categories=true) %}
     {% for category, message in messages %}
       {% if message == 'message_sent' %}
-        {% set message = "Your question has been sent. It will be posted with the buyer’s response on the ‘{}’ page.".format(brief.title) %}
+        {% set message = "Your question has been sent. The buyer will post your question and their answer on the ‘{}’ page.".format(brief.title) %}
       {% endif %}
       {%
         with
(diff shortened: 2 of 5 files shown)


app/assets/scss/application.scss
```diff
@@ -84,3 +84,7 @@ $path: "/suppliers/static/images/";
   @include core-19;
   margin: (1.5 * $gutter) 0 (2 * $gutter);
 }
+
+.padding-bottom-small {
```
TheDoubleK: I think I introduced this to the buyer frontend a little while ago (it was a thing in the prototype that I copied over but probably shouldn't have), and I've been trying to backtrack on it and removed it from a couple of places recently.  But it seems to be a useful thing (Caroline added a bunch more in buyer frontend recently), so let's stick with it for now.

app/templates/briefs/clarification_question.html
```diff
@@ -53,16 +53,17 @@
...
+      {% with
+        heading = "Ask a question about ‘{}’".format(brief.title),
+        smaller = true
+      %}
+        {% include 'toolkit/page-heading.html' %}
+      {% endwith %}
+
+
       <div class="hint">
         <p>Your question will be published with the buyers’ response by {{ brief.clarificationQuestionsClosedAt | dateformat }}.</p>
```
TheDoubleK: Not part of this PR, but it worries me that the text here isn't true.  The date is the last date a question can be asked - from discussions the other day it seems like the answer might not get posted until the day before the opportunity closes.

A proper fix might need to wait for content sign-off for this page (the prototype doesn't include this new feature yet:
http://dm-prototype.herokuapp.com/page/ask-question/template-brief?key=outcomes.1), but we could remove the misleading date now, just to force whoever adds a date back in to think about what the right date is.

TheDoubleK:
Changes for the new page all good - :+1: once we work out what to say about the date.

app/assets/scss/application.scss
carolinegreen: Oh, right, it is useful.  Can worry about replacing it later.

PR #456, opened 2016-04-18
Fix framework dashboard bug

TheDoubleK:
The decision for whether an application has been made to a framework currently checks whether there are complete drafts and a complete declaration.  This is fine for frameworks that are not yet live.

However, once a framework is made live we may at some point delete the old draft, at which point the page would show "You did not make an application" even though the supplier is on the framework.

This adds the logic that if the supplier is currently on the framework then they must have made an application to it.


pcraig3:
:+1:

PR #457, opened 2016-04-18
Pass price field value to the form template

allait:
screen shot 2016-04-18 at 17 02 10

Adds macro support for the 'price' field for pricing questions.

Updates dmutils and frameworks to bring in new email and dayRate field types.


TheDoubleK:
:+1:

PR #459, opened 2016-04-20
[#117922093] Hide boolean list question section if there are no questions

allait:
Hides the 'nice to have requirements' question header if there are not questions defined in the brief.


TheDoubleK:
Looks like it should work :+1:

PR #461, opened 2016-04-22
[#118112493] Update content and frontend toolkit versions to latest

TheDoubleK:
Fixes this bug: https://www.pivotaltracker.com/story/show/118112493

We already bumped content in buyer app to be in line with schemas, but supplier app is behind, so brief-responses don't match the schema in the API.


allait:
👍

PR #467, opened 2016-04-26
G8 service submission

TheDoubleK:
Completes this story: https://www.pivotaltracker.com/story/show/118178641
And this story: https://www.pivotaltracker.com/story/show/118178637

This seems to be everything needed to get G-Cloud service submissions looking reasonably peachy:
 * Show links for lots without any draft services for live frameworks
 * Reinstate the save-and-continue-to-next-question flow through submissions
 * Fix document download URLs (document uploads were fine, but the generated download links were wrong)

There is a bug (again! I'm sure we've fixed this about three times now...) with the validation wrapper around questions with assurance.  I've raised it as a bug rather than fix it right away, because there are other higher priorities, like getting the functional tests running again: https://www.pivotaltracker.com/story/show/118388103

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -4,7 +4,8 @@ from flask import render_template, request, redirect, url_for, abort, flash, cur
 from ... import data_api_client, flask_featureflags
 from ...main import main, content_loader
 from ..helpers import login_required
-from ..helpers.services import is_service_associated_with_supplier, get_signed_document_url, count_unanswered_questions
+from ..helpers.services import is_service_associated_with_supplier, get_signed_document_url, count_unanswered_questions, \
+    get_next_section_name
 from ..helpers.frameworks import get_framework_and_lot, get_declaration_status, has_one_service_limit
 
 from dmapiclient import HTTPError
@@ -240,7 +241,9 @@ def copy_draft_service(framework_slug, lot_slug, service_id):
                             framework_slug=framework['slug'],
                             lot_slug=draft['lot'],
                             service_id=draft_copy['id'],
-                            section_id=content.get_next_editable_section_id()))
+                            section_id=content.get_next_editable_section_id(),
+                            return_to_summary=1
+                            ))
 
 
 @main.route('/frameworks/<framework_slug>/submissions/<lot_slug>/<service_id>/complete', methods=['POST'])
@@ -388,8 +391,10 @@ def edit_service_submission(framework_slug, lot_slug, service_id, section_id, qu
         "services/edit_submission_section.html",
         section=section,
         framework=framework,
+        next_section_name=get_next_section_name(content, section.id),
         service_data=draft,
         service_id=service_id,
+        return_to_summary=bool(request.args.get('return_to_summary')),
         one_service_limit=has_one_service_limit(lot_slug, framework['lots'])
     )
 
@@ -452,17 +457,29 @@ def update_section_submission(framework_slug, lot_slug, service_id, section_id, 
             "services/edit_submission_section.html",
             framework=framework,
             section=section,
+            next_section_name=get_next_section_name(content, section.id),
             service_data=update_data,
             service_id=service_id,
             one_service_limit=has_one_service_limit(lot_slug, framework['lots']),
+            return_to_summary=bool(request.args.get('return_to_summary')),
             errors=errors
         )
 
-    return redirect(url_for(".view_service_submission",
-                            framework_slug=framework['slug'],
-                            lot_slug=draft['lot'],
-                            service_id=service_id,
-                            _anchor=section_id))
+    return_to_summary = bool(request.args.get('return_to_summary'))
+    next_section = content.get_next_editable_section_id(section_id)
+
+    if next_section and not return_to_summary and request.form.get('continue_to_next_section'):
+        return redirect(url_for(".edit_service_submission",
+                                framework_slug=framework['slug'],
+                                lot_slug=draft['lot'],
+                                service_id=service_id,
+                                section_id=next_section))
+    else:
+        return redirect(url_for(".view_service_submission",
+                                framework_slug=framework['slug'],
+                                lot_slug=draft['lot'],
+                                service_id=service_id,
+                                _anchor=section_id))
 
 
 @main.route('/frameworks/<framework_slug>/submissions/<lot_slug>/<service_id>/remove/<section_id>/<question_slug>',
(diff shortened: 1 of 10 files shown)


app/main/views/services.py
```diff
@@ -305,15 +308,15 @@ def delete_draft_service(framework_slug, lot_slug, service_id):
                                 delete_requested=True))
 
 
-@main.route('/frameworks/<framework_slug>/submissions/documents/<int:supplier_id>/<document_name>', methods=['GET'])
+@main.route('/frameworks/submissions/<framework_slug>/submissions/<int:supplier_id>/<document_name>', methods=['GET'])
```
allait: I think I got really confused with all the URLs in that Slack discussion. Looking at it now, I think we don't need the `submissions` part after `/frameworks/` since our common URL prefix for the views is `/frameworks/<framework_slug>/submissions`. So the URL could be just `/frameworks/<framework_slug>/submissions/<int:supplier_id>/<document_name>`. Which makes the URL nicer without the duplicate `submissions` bit, but I think it might create some possible ordering issues with `/suppliers/frameworks/<framework_slug>/submissions/<lot_slug>` views.


We could change it to something like `/documents/<framework_slug>/submissions/<int:supplier_id>/<document_name>` and move it away from framework URLs altogether instead.

TheDoubleK: I like `assets` - I thought it seemed odd to hang it off `frameworks` really.  Not that anyone will ever see the URL, but still... will fix now.

TheDoubleK:
URL changed to `/suppliers/assets/<framework_slug>/submissions/<int:supplier_id>/<document_name>`

TheDoubleK:
But now depends on this getting merged first: https://github.com/alphagov/digitalmarketplace-frameworks/pull/262

app/main/views/services.py
```diff
@@ -452,17 +457,29 @@ def update_section_submission(framework_slug, lot_slug, service_id, section_id,
...
-    return redirect(url_for(".view_service_submission",
-                            framework_slug=framework['slug'],
-                            lot_slug=draft['lot'],
-                            service_id=service_id,
-                            _anchor=section_id))
+    return_to_summary = bool(request.args.get('return_to_summary'))
+    next_section = content.get_next_editable_section_id(section_id)
+
+    if next_section and not return_to_summary and request.form.get('continue_to_next_section'):
+        return redirect(url_for(".edit_service_submission",
```
allait: I'm still not sure why we're restoring this. Our DOS submission flow was always going from section back to summary and the brief submission is going form individual question to the section summary. I'm not sure why we want G-Cloud flow to be different and how this would look if we decide to do one-question-per-page.

/cc @ralph-hawkins

TheDoubleK: It's a requirement from Cath, and I think it's the right thing for G-Cloud.
Trying to complete an IaaS service going back-and-forth is genuinely horrible - I did half of one before giving up and picking up the story to restore the "save and continue" flow, just to save my sanity.

TheDoubleK: As well as that I think the reasoning is that suppliers are already familiar with this flow.  We aren't changing anything else at all for G8, so why would we change this and throw them?

allait: This is a temporary solution for the third time though – next time DOS submissions open we'll have to switch it back again because we want "back to summary" for DOS. I think if we decide to keep using this flow for G-Cloud we need to support both DOS and G-Cloud ways at once instead of changing this code every iteration.

ralph-hawkins: Its a question of how people add this information in a single flow vs many sittings. 

We've tested pretty heavily  (DOS applications and buyer journey) with the pattern of returning to the overview each time.

If people are using data from old services, then they're possibly going to go through multiple questions.

Other than that we would probably expect people to want to make sure they’re answering things correctly.

As a principle I favour clarity and making sure people do the right thing over worrying about people getting bored to clicking too much.

Plus we have the copy service feature which would make more sense to want to target certain questions rather than running through the complete service.

If the button takes you to the next question then we need the secondary button to return you to the overview, which I don't think is a great solution. The one contextual-research session I sat in on, the user didn't use the secondary 'Save and return to service overview'. Using the nice big green button to save any work they've done and then the breadcrumb to get to the overview. 

@TheDoubleK I don't think your experience reflects a typical user‘s, getting through the questions as quickly as you can isn't something users want to do. 

I also doubt many users will remember g-cloud 7 well enough and we’re making a change from DOS which is more recent. And means we’re probably going to have to change this behaviour once again?

TheDoubleK: We need to get a call on this from Cath - restoring this flow is what's in the story. 

From what I understand users *do* want to be able to get through the questions - many of them will collate the answers to all the questions in a spreadsheet, talking to whoever is needed across the organisation to get the right answers, and then someone will sit and "data entry" them into the Marketplace.  If that's a flow that many suppliers use then we should support it.

TheDoubleK:
The only way I could think to maintain both possible flows was to add a column to the `frameworks` table and a property to the Framework object to specify `flow_type` or something like that.  But it seemed weird so I didn't do it right off.

I think for now it would be good to reinstate this flow and get G-8 SSP tests running on preview, and then put a story in the backlog to implement the two different flows in parallel, and discuss the best way to do that when that new story gets kicked off.

app/main/views/services.py
TheDoubleK: Or rather than the secondary grey button, could we have two green buttons to show that either will save work equally well?

ralph-hawkins: Green buttons are for the primary action, using two visually similar, equally weighted actions forces users make a decision on every page. Its unnecessary cognitive load.

We should giving users the simplest path through.

We have videos from all the contextual research we did during for g7. If you can wait, i'll try and find them and take a look, it'll hopefully shed some light.

cathrooney:
I am happy that for now we are reverting to the G7 behaviour and that we revisit this thoroughly next time.  We need to determine what patterns suit different types of services.  We may find that the questions asked for G9 are radically different to earlier G-Cloud iterations.

We have always intended to make as few changes as possible for G8.  The DOS pattern of always forcing users back to the overview falls over when a supplier has to fill in as much data as they have to for their G-Cloud services.  I'm unaware that we saw any complaints about this during G7 applications.  Therefore to my mind this means that, with the time and resources we have available, we leave our suppliers to get on with loading their services as they have done before on earlier iterations.

Any further work we consider for this will need to be prioritised against all our other DOS iteration / digital framework agreements /  supplier editing  work etc...

TheDoubleK:
@allait, @ralph-hawkins - so can we merge this as is?

ralph-hawkins:
I'm ok with it.

allait:
👍

PR #468, opened 2016-04-27
Add `question_advice` to question-rendering macros

TheDoubleK:
Completes this: https://www.pivotaltracker.com/story/show/117629721

This is already done in the buyer app, but we also need to show the advice in the supplier app.

Of questions rendered by the supplier app only the `boolean_list` questions Essentials and Nice-to-haves currently have `question_advice` set.  But I've added it to all question-rendering templates, in case other supplier questions have advice in future.

Screenshots of the relevant page once this is applied:
screen shot 2016-04-27 at 08 07 21

screen shot 2016-04-27 at 08 07 27

### DIFF
--- a/app/templates/macros/toolkit_forms.html
+++ b/app/templates/macros/toolkit_forms.html
@@ -3,7 +3,8 @@
     with
     question=question_content.question|question_references(get_question)|markdown,
     question_advice=question_content.question_advice,
-    hint=(question_content.hint or '')|question_references(get_question)|markdown,
+    hint=(question_content.hint or '')|question_references(get_question),
+    optional=question_content.optional,
     name=question_content.id,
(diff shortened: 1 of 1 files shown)


allait:
👍

pcraig3:
Couple things:

1. Probably worth removing the `|markdown` filter on hints because [the markdown filter is applied in the toolkit itself](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/toolkit/templates/forms/textbox.html#L42).
2. Seems like a good opportunity to pass in the `optional` flag.

-> [Precedent in the buyer app](https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/272/files#diff-712f48621dd493057d3678f370d01074)

PR #469, opened 2016-05-04
Split registered address into three fields

TheDoubleK:
Completes this story: https://www.pivotaltracker.com/story/show/118180095

Depends on updated content being merged: 
 - [x] https://github.com/alphagov/digitalmarketplace-frameworks/pull/265

This introduces a new macro to render multiquestions in the frontend.
It assumes all questions are `text` type.

Look and feel, including validation messaging and wrapping, all OK'd by @ralph-hawkins 

## Question renders like this:

screen shot 2016-05-04 at 13 54 23

## And if there are validation errors:

screen shot 2016-05-04 at 13 54 06

screen shot 2016-05-04 at 13 53 59

### DIFF
--- a/app/main/helpers/validation.py
+++ b/app/main/helpers/validation.py
@@ -36,12 +36,14 @@ class DeclarationValidator(object):
     def get_error_messages(self):
         raw_errors_map = self.errors()
         errors_map = list()
-        for question_number, question_id in enumerate(self.all_fields()):
+        for question_id in self.all_fields():
             if question_id in raw_errors_map:
+                question_number = self.content.get_question(question_id).get('number')
                 validation_message = self.get_error_message(question_id, raw_errors_map[question_id])
                 errors_map.append((question_id, {
                     'input_name': question_id,
-                    'question': "Question {}".format(question_number + 1),
+                    'question': "Question {}".format(question_number)
+                    if question_number else self.content.get_question(question_id).get('question'),
                     'message': validation_message,
                 }))
 
--- a/app/templates/macros/toolkit_forms.html
+++ b/app/templates/macros/toolkit_forms.html
@@ -172,19 +172,21 @@
 {% macro boolean_list(question_content, service_data, errors, question_number=None, get_question=None) -%}
   {% if question_content.boolean_list_questions %}
     <fieldset class="question question-boolean-list" id="{{ question_content.id }}">
+        <legend>
+          <span class="visually-hidden">{{ question_content.name }}</span>
+        </legend>
+
         <span class="question-heading">
           <h2>{{ question_content.question }}</h2>
         </span>
-      <legend>
-        <span class="visually-hidden">{{ question_content.name }}</span>
-      </legend>
+
       {% if question_content.question_advice %}
         <span class="question-advice" id="input-{{ name }}-question-advice">
           {{ question_content.question_advice|markdown }}
         </span>
       {% endif %}
       {% if question_content.hint %}
-        <p class="question-description">{{ question_content.hint|markdown }}</p>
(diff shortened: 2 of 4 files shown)


app/main/helpers/validation.py
```diff
@@ -36,12 +36,14 @@ def get_error_messages_for_page(self, section):
...
-        for question_number, question_id in enumerate(self.all_fields()):
+        for question_id in self.all_fields():
             if question_id in raw_errors_map:
+                question_number = self.content.get_question(question_id).get('number')
                 validation_message = self.get_error_message(question_id, raw_errors_map[question_id])
                 errors_map.append((question_id, {
                     'input_name': question_id,
-                    'question': "Question {}".format(question_number + 1),
+                    'question': "Question {}".format(question_number)
+                    if question_number else self.content.get_question(question_id).get('question'),
```
pcraig3: <del>The question names seem more helpful than the numbers, but both might be most helpful.</del>
Actually, I don't think there's a way of getting data about a parent question from a child question, so never mind.

app/templates/macros/toolkit_forms.html
```diff
@@ -209,3 +209,38 @@
...
+          </span>
+        
+        {{ question_content.question }}
+        
+      </span>
+
+  </label>
+  <legend>
+    <span class="visually-hidden">{{ question_content.name }}</span>
+  </legend>
```
pcraig3: [`<legend>` elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/legend) are meant to be used within a `<fieldset>` element, and this looks like a pretty good candidate for being a fieldset.

The other thing I've just noticed is that, according to the MDN guide:

> `<legend>`
> Permitted Parents: A `<fieldset>` whose first child is this `<legend>` element

This isn't how [the `boolean_list` template](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/templates/macros/toolkit_forms.html#L174) is structured.

If it's an easy change to make, do you mind swapping the `<span class="question-heading">` and the `<legend>` elements in the above template as well? (If not, I can do it.)

app/templates/macros/toolkit_forms.html
```diff
@@ -209,3 +209,38 @@
...
+  {% if question_content.question_advice %}
+        <span class="question-advice" id="input-{{ name }}-question-advice">
+          {{ question_content.question_advice|markdown }}
+        </span>
+  {% endif %}
+  {% if question_content.hint %}
+  <p class="question-description">{{ question_content.hint|markdown }}</p>
+  {% endif %}
```
pcraig3: We're using [`<span class="hint">` in the toolkit for the question hint](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/toolkit/templates/forms/textbox.html#L41).  Any reason not to here?

app/main/helpers/validation.py
TheDoubleK: No - I thought about this and it made my head hurt thinking how to get the parent question number.  I gave a sigh of relief when Ralph was happy with just the question name.

app/templates/macros/toolkit_forms.html
TheDoubleK: I just copied it from `boolean_list` - I'll change it both places if you reckon that will work?  To be honest we don't have any hints on our sole multiquestion, so the point is moot.

TheDoubleK:
Fixed up the legend and fieldset stuff, and used the toolkit `hint` style for hints.
Tested locally and looks the same as before and still works.

pcraig3:
Looks good to me.  👍 

<sub>Thanks for making those boolean list changes.</sub>

PR #471, opened 2016-05-06
Services not labs

TheDoubleK:
Depends on this version of the API running:
 - [x] https://github.com/alphagov/digitalmarketplace-api/pull/390

Fixes this bug: https://www.pivotaltracker.com/story/show/118986461

I've also done a bit of tidying up:
 * removing the `has_one_service_limit` function in favour of using the `lot['oneServiceLimit']` value directly
 * checking that the `framework_slug` and `lot_slug` in the URL match those of the draft being edited 
 * moving to using `lotSlug` rather than `lot` (they have identical values but we want to phase out using `lot` eventually)

This broke a lot more tests than I thought it would.

### DIFF
--- a/tests/app/helpers.py
+++ b/tests/app/helpers.py
@@ -86,19 +86,21 @@ FULL_G7_SUBMISSION = {
     "SQE2a": ["as a prime contractor, using third parties (subcontractors) to provide some services"]
 }
 
-EMPTY_G7_DRAFT = {
-    'id': 1,
-    'supplierId': 1234,
-    'supplierName': 'supplierName',
-    'frameworkName': 'G-Cloud 7',
-    'frameworkSlug': 'g-cloud-7',
-    'lot': 'scs',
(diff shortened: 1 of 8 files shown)


tests/app/helpers.py
```diff
@@ -86,6 +86,20 @@
     "SQE2a": ["as a prime contractor, using third parties (subcontractors) to provide some services"]
 }
 
+EMPTY_G7_DRAFT = {
```
allait: I'd make this a function that returns a new dictionary each time it's called (or even better, a py.test fixture). That gets rid of all the `.copy()` calls and makes sure that it or any of the nested items are never accidentally modified.

TheDoubleK: I went with making it a function in the end - couldn't see any immediate benefit to making it a py.test fixture right now.

allait:
👍

PR #473, opened 2016-05-10
Don't mark dates down

TheDoubleK:
Required to fix:
 - https://www.pivotaltracker.com/story/show/119086793
 - https://www.pivotaltracker.com/story/show/119075005

Depends on: 
 - [x] https://github.com/alphagov/digitalmarketplace-frameworks/pull/269
 - [x] Now also needs: https://github.com/alphagov/digitalmarketplace-frameworks/pull/272

The markdown filter wraps values in `<p>` tags but we generally want to use dates inline and the `<p>` breaks the page layouts in the frontend.  So we're not going to use markdown for date strings.

# BEFORE
screen shot 2016-05-10 at 13 33 45
---------------------------
screen shot 2016-05-10 at 13 35 33
---------------------------
screen shot 2016-05-10 at 13 40 24

# AFTER
screen shot 2016-05-10 at 13 34 32
---------------------------
screen shot 2016-05-10 at 13 34 51
---------------------------
screen shot 2016-05-10 at 13 41 15

### DIFF
--- a/app/templates/frameworks/_dashboard_lede.html
+++ b/app/templates/frameworks/_dashboard_lede.html
@@ -1,6 +1,6 @@
 {% if framework.status == 'open' %}
 <aside role="complementary" class="framework-application-status" aria-label="{{ framework.name }} status">
-  Deadline: <strong>{{ dates.framework_close_date|markdown }}</strong>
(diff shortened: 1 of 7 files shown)


tombye:
Looks quite excellent to me 👍

PR #475, opened 2016-05-10
[#118094925] Question title displayed twice on page

pcraig3:
As part of the story about removing duplicated text in question labels, I removed the `_single-question-page.scss` template and updated the toolkit version to the version that'll have the removed CSS.

I also removed the manually-built `.hint` text for the supplier creation journey and passed everything into the `question_advice` field.  Pure HTML can be passed through markdown filters, so it seemed okay to just whack everything in there.  The big win is that we can use more general styling as the markup is standardized.
It's a bit funny looking, but it works.  

Also got another non-breaking hyphen in. (yes!)

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/118094925)

***

## old -> new

New version has (fatter) non-breaking hyphen and more space between the last line and the input field.

duns


app/templates/briefs/clarification_question.html
```diff
@@ -63,16 +63,15 @@
...
-      <div class="hint">
-        <p>Your question will be published with the buyers’ response by {{ brief.clarificationQuestionsPublishedBy | dateformat }}.</p>
-        <p>All questions and answers will be posted on the Digital Marketplace. Your company name won’t be visible.</p>
-        <p>You shouldn’t include any confidential information in your question.</p>
-        <p>Read more about <a href="https://www.gov.uk/guidance/how-to-answer-supplier-questions-about-your-digital-outcomes-and-specialists-requirements">how supplier questions are managed</a>.</p>
-      </div>
-
       <form method="post">
         <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
+        {% set question_advice =
```
TheDoubleK: I don't get how these `question_advice` things get rendered?  They are set as a variable but then never passed in to anything?  What is this black magic?

app/templates/suppliers/companies_house_number.html
```diff
@@ -28,12 +28,13 @@
...
     <div class="column-two-thirds">
-      <div class="hint">
-        <p>This is a unique 8-character number given to your organisation by Companies House.</p>
-        <p><a href="https://beta.companieshouse.gov.uk/">Visit Companies House to get your number</a></p>
-      </div>
       <form method="POST" action="{{ url_for('.companies_house_number') }}">
         <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
+        {% set question_advice =
+            'This is a unique 8-character number given to your organisation by Companies House.
+            \n\n<a href="https://beta.companieshouse.gov.uk/">Visit Companies House to get your number</a>
```
TheDoubleK: You really like `\n\n` more than some plain HTML?

pcraig3: If you write pure HTML into this field, it gets interpreted really strangely by markdown when you start  using linebreaks.  This way, it's easier for humans to maintain, though I agree that it's not ideal.

TheDoubleK: Does it not work with passing in actual line breaks rather than `\n`?

pcraig3: No, it's doing all sorts of weird stuff.  It starts putting things in `<code>` blocks :/

TheDoubleK: That's to do with spacing at the start of lines.  It might be nicer to left-align the markdown and avoid having the `\n`?

TheDoubleK: I don't know... that will probably look weird in the template too.

pcraig3: So, you're right.  Doing that does work.

@Wynndow had a look and he was saying he thinks the `\n\n` is clearer because you're passing in the formatting information explicitly rather than needing to ensure on left-aligned text with the right number of line breaks separating them.

I think that's a reasonable viewpoint and I'm happy enough to leave the `\n`s even though it obviously looks a bit shit. 

What do you think?

TheDoubleK: I liked it how it was to start with, with the straight-up HTML text and none of this markdown that obviously looks a bit shit.

But I can tell that's a battle not worth fighting at this point :)

TheDoubleK: I'd quite like to hear an opinion from @tombye - not because I think he'll agree with me, but just because I've not seen this in templates before and he's likely to have a sensible viewpoint on it.

tombye: Both methods are a bit odd because of the things YAML implies from newlines in certain places.

The point of @pcraig3 's change seems to be to make it more explicit how the paragraphs will wrap the given text. That being so I'd plump for the method using literal `\n`s.

We should try and work with this method and see if it removes some of the confusion (or causes more?).

pcraig3: Is that a thumbs up, @tombye?

TheDoubleK:
OK`\n` let's give it a go `\n`:+1:`\n`

PR #476, opened 2016-05-12
Make `G-Cloud X` non-breaking on all pages

TheDoubleK:
Work done with @carolinegreen to address this issue: https://www.pivotaltracker.com/story/show/118404225

This adds an `after_request` handler that (for HTML responses only) searches for occurrences of `G-Cloud x` (where `x` is a digit) and replaces it with a non-breaking version of the same string.

We first tried wrapping the base Jinja template with a filter to do the same thing, but unless we make everything `|safe` Jinja escapes the HTML entities and it all looks a mess.  By restricting the `after_request` handler to HTML responses only this seems like a reasonable alternative. 

It seems pretty crazy to me, but I can't think of any compelling reasons why this will break things.  The regular expression is short and contains no greedy quantifiers or anything fancy, so should be efficient even over big pages I think.  It's also very specific, and I can't think of any cases where it would match something that we don't want to transform. (But you should have a good think about this too before agreeing to this madness.)

If we like this then we should add the same thing to the buyer app.  We decided against adding it to the  base `flask_init` in utils (it could fit right here: https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/flask_init.py#L49) because we don't necessarily want it operating in the API and admin apps, although with it limited to HTML responses only it might not be so bad there.  Opinions welcome.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -71,11 +71,13 @@ def create_app(config_name):
         session.permanent = True
         session.modified = True
 
+    G_CLOUD_REGEX = re.compile(r"(G-Cloud )(\d)")
+
     @application.after_request
     def dont_break_gcloud(response):
         if "text/html" in response.content_type:
-            page = response.get_data()
-            page = re.sub("(G-Cloud )(\d)", r'G&#8209;Cloud&nbsp;\2', page)
+            page = response.get_data(as_text=True)
+            page = G_CLOUD_REGEX.sub(r'G&#8209;Cloud&nbsp;\2', page)
             response.set_data(page)
         return response
 
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -71,6 +71,14 @@ def create_app(config_name):
         session.permanent = True
         session.modified = True
 
+    @application.after_request
+    def dont_break_gcloud(response):
+        if "text/html" in response.content_type:
+            page = response.get_data()
+            page = re.sub("(G-Cloud )(\d)", r'G&#8209;Cloud&nbsp;\2', page)
+            response.set_data(page)
+        return response
+
     application.add_template_filter(question_references)
     application.add_template_filter(parse_document_upload_time)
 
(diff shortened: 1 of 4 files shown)


allait:
I really don't like the after request middleware. While it does look like a very quick solution, rewriting the whole page has a performance cost (that will multiply whenever we decide to add more cases) and I think it can have too many unexpected consequences.

For example, it will not only rewrite the static page content, it will also rewrite form values (since they're part of HTML) when editing or when user response is rendered back with an error, so this will end up modifying the data that gets stored.

One possible solutions is adding unicode characters to the framework names in the database. Replacing current names with `u'G-Cloud\u2011\xa07'` does end up rendering them with non-breaking unicode hyphen and space, but it immediately exposes bits of code that don't handle unicode properly, eg mixing unicode values into normal strings in Python 2, which will require fixing things like this
```diff
         {%
           with
           items = [{
-            "title":
-            "Continue your {} application".format(framework.name),
+            "title": "Continue your %s application" | format(framework.name),
             "link": url_for(".framework_dashboard", framework_slug=framework.slug),
             "body": framework.deadline,
           }]
```  

This is generally a useful change, but I'm not sure if all of those cases are covered by tests and maybe changing this so close to G8 launch could be a bit disruptive.

TheDoubleK:
I agree with all your reservations about this approach, and that this probably isn't the time to fix this.  It also isn't just the database values that would need to be sorted - I think there are also occurrences in the frameworks content repository, but making changes there would mean needing to `|safe` a lot of things in templates.  

Closing, even though it will make Ralph sad.

carolinegreen:
Yeah, we were worried about this.  I wonder if pre-processing could be applied to the content loader though, as it doesn't really seem like a good approach to force people to type in non breaking spaces every single time G-Cloud is referenced.  If that were changed and the database changes were made, it would just leave G-Clouds which were typed directly into the application.

allait:
@carolinegreen yeah, we could do replacements when loading frameworks data since we know a lot more about context there. And it only happens on startup, so there's 0 performance considerations.

allait:
@TheDoubleK we don't need to `| safe` if we use unicode characters for non-breaking hyphen and space instead of HTML entities. We will need to ensure that we're handling unicode properly though and there might be some browser issues related to encoding (I don't know of any, but there must be a reason for HTML entities in the first place)

PR #480, opened 2016-05-17
Update to latest toolkit

pcraig3:
Bring in the latest toolkit and remove local version of `_reset.scss` because the latest toolkit has a copy of this file.

Update `application.scss` to reflect this.

No changes at all to presentation or functionality.


TheDoubleK:
:+1:

PR #482, opened 2016-05-19
Pull in latest content with G8 ZIP file size

TheDoubleK:
Depends on: 
 - [x] https://github.com/alphagov/digitalmarketplace-frameworks/pull/280


allait:
👍

PR #485, opened 2016-05-24
Update supplier confirmation email

Wynndow:
For [this](https://www.pivotaltracker.com/story/show/118095523) story in Pivotal.

The content of the email which is sent to a supplier after they ask a question of the buyer has been updated.

Also, a variable which was being assigned and passed in to the email template has been removed as it was no longer required after the content update.

### DIFF
--- a/app/templates/emails/brief_clarification_question_confirmation.html
+++ b/app/templates/emails/brief_clarification_question_confirmation.html
@@ -10,7 +10,7 @@
   <br />
   You asked:<br />
   <br />
-  {{ message }}
+  ‘{{ message }}’
   <br /><br />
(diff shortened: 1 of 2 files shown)


app/templates/emails/brief_clarification_question_confirmation.html
```diff
@@ -4,15 +4,15 @@
...
 </head>
 <body>
-  Dear {{ user_name }},<br />
+  Dear supplier,<br />
   <br />
   Your question about ‘{{ brief_name }}’ has been sent.<br />
   <br />
   You asked:<br />
   <br />
   {{ message }}
```
TheDoubleK: From the story it looks like the `message` here should also be in (typographical) quotes, the same as `brief_name` above.

allait:
👍

PR #487, opened 2016-06-02
[WIP] Track supplier questions

tombye:
Note: Do not merge until [#330](https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/330) on the buyer app is merged.

Adds analytics to track the submission of supplier questions raised on opportunities.

https://www.pivotaltracker.com/story/show/119634865

Submitting a question is a POST request so this allows us to submit a [virtual page view](https://developers.google.com/analytics/devguides/collection/analyticsjs/pages) to Google analytics to simulate a request to `/suppliers/opportunities/{brief id}/ask-a-question?submitted=true`.

PR #488, opened 2016-06-03
Tracking supplier applications to opportunities

tombye:
A successful application to an opportunity results in a redirect to the supplier's dashboard which has the same URL as normal. 

To track the application in analytics, this marks the request as the result of an application by adding a query string parameter to the URL.

https://www.pivotaltracker.com/story/show/119634783

### DIFF
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -151,7 +151,7 @@ def submit_brief_response(brief_id):
 
     if all(brief_response['essentialRequirements']):
         flash('Your response to ‘{}’ has been submitted.'.format(brief['title']))
-        return redirect(url_for(".dashboard"))
+        return redirect(url_for(".dashboard", applied_for_brief=brief['id']))
     else:
         return redirect(url_for(".view_response_result", brief_id=brief_id))
 
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -151,6 +151,7 @@ def submit_brief_response(brief_id):
 
     if all(brief_response['essentialRequirements']):
         flash('Your response to ‘{}’ has been submitted.'.format(brief['title']))
+        # applied_for_brief parameter is used to track brief applications by analytics
         return redirect(url_for(".dashboard", applied_for_brief=brief['id']))
     else:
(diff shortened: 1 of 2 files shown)


TheDoubleK:
Is there really no better way to do this? This feels horribly open to people manually messing about with URLs, and also weird to pass a parameter that's then ignored by the app.

TheDoubleK:
I think it was discussed previously that rather than back to the dashboard the redirect could be through to the "results page", the same as for suppliers who failed to tick yes for all essentials.  This route (`'/opportunities/<int:brief_id>/responses/result'`) has the brief ID in it so there would be no need to hack it in as a parameter.

Also, I think this soon-upcoming story: https://www.pivotaltracker.com/story/show/119262435 is quite likely to change things again.

tombye:
We've added unused query-string parameters before (to supplier declaration flows I think) and haven't seen problems (as far as I know from talking to Ash). It does also reflect the state change of the confirmation banner appearing in the page so isn't entirely outside of the intent of the response.

Do you have a specific concern with the particular flow in this context? If so, we could fake it with a virtual pageview, like we did with [the pull request for tracking supplier questions](https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/487) which would keep the URL clean but would still be vulnerable to monkeying with the URL.

TheDoubleK:
I don't have any really strong arguments against this, it just feels like a weird thing to do.  Is there a reason not to do a virtual pageview here?  Seems like it might be nice to do the analytics hacks the same way everywhere.

tombye:
OK, I've looked into using a virtual pageview and the problem with that approach is it relies on the banner being in the page and access to the brief id.

We can check for the banner but the brief id doesn't appear in the page in any way. One strategy would be to add it as meta data to the HTML of the banner but this would mean extending the front-end pattern for this. We could add the meta to a different part of the document but this would mean building code into the template to detect for flash which I think is going a bit far.

TheDoubleK:
Ok cool, thanks for looking at the options, sorry to send you off on one...
Let's go with this for now :+1:

TheDoubleK:
Brilliant, ta.

PR #490, opened 2016-06-13
[#118392929] Update error message if supplier isn't eligible

risicle:
Ok a few caveats here:

- In this case we've switched from using the `SupplierFramework.onFramework` to determine the supplier's "on framework"-ness to poking the api to see if the supplier has any published services on the lot or framework. However, there are still a few bits of this frontend that do use `onFramework` for various things. So I guess there's some possibility of an incongruence here. From what I can tell the only functional difference will be for suppliers who've had services removed for some reason. And most of the other places `onFramework` is used in this frontend seem to be during the pre-live phase. So maybe... it's... all... ok?
- Depends on `digitalmarketplace-apiclient` 3.8.0, which isn't merged yet.
- Tests had to be changed a bit, and I hope I still captured the _intent_ of them properly.

### DIFF
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -184,12 +184,21 @@ def view_response_result(brief_id):
 
 
 def _render_not_eligible_for_brief_error_page(brief, clarification_question=False):
-    sf = get_supplier_framework_info(data_api_client, brief['frameworkSlug'])
-    on_framework = sf.get('onFramework', False) if sf else False
+    common_kwargs = {
+        "supplier_id": current_user.supplier_id,
+        "framework": brief['frameworkSlug'],
+    }
+    has_framework_service = bool(data_api_client.find_services(**common_kwargs)["services"])
+    has_framework_lot_service = has_framework_service and bool(data_api_client.find_services(
+        **dict(common_kwargs, lot=brief['lotSlug'])
+    )["services"])
+    # if has_framework_lot_service is true, we can deduce that the problem is that the roles don't match.
+
     return render_template(
         "briefs/not_is_supplier_eligible_for_brief_error.html",
-        on_framework=on_framework,
         clarification_question=clarification_question,
+        has_framework_service=has_framework_service,
+        has_framework_lot_service=has_framework_lot_service,
         framework_name=brief['frameworkName'],
         lot=brief['lotSlug'],
         **dict(main.config['BASE_TEMPLATE_DATA'])
--- a/app/templates/briefs/not_is_supplier_eligible_for_brief_error.html
+++ b/app/templates/briefs/not_is_supplier_eligible_for_brief_error.html
@@ -4,40 +4,24 @@
 
 {% block main_content %}
 
-{% if clarification_question %}
+{% with inhibited_action="ask a question about" if clarification_question else "apply for" %}
 <div class="grid-row">
   <div class="column-two-thirds">
     <header class="page-heading-smaller">
-      <h1>You can’t ask a question about this opportunity</h1>
+      <h1>You can’t {{ inhibited_action }} this opportunity</h1>
     </header>
-    {% if on_framework %}
-      <p>
-        You can’t ask a question about this opportunity because you don’t provide
-        {{ "this digital specialist role" if lot == 'digital-specialists' else "a service in this category" }}
-        at this location.
-      </p>
-    {% else %}
-      <p>You can’t ask a question about this opportunity because you’re not a {{ framework_name }} supplier.</p>
-    {% endif %}
+    <p>
+      You can’t {{ inhibited_action }} this opportunity because
+      {% if has_framework_service %}
+        you didn’t say you could provide
+        {{ "this specialist role" if has_framework_lot_service else "services in this category" }}
+        when you applied to the Digital Outcomes and Specialists framework.
+      {% else %}
+        you’re not a {{ framework_name }} supplier.
+      {% endif %}
+    </p>
   </div>
 </div>
-{% else %}
-<div class="grid-row">
-  <div class="column-two-thirds">
-    <header class="page-heading-smaller">
-      <h1>You can’t apply for this opportunity</h1>
-    </header>
-    {% if on_framework %}
-      <p>
-        You can’t apply for this opportunity because you don’t provide
-        {{ "this digital specialist role" if lot == 'digital-specialists' else "a service in this category" }}
-        at this location.
-      </p>
-    {% else %}
-      <p>You can’t apply for this opportunity because you’re not a {{ framework_name }} supplier.</p>
-    {% endif %}
-  </div>
-</div>
-{% endif %}
+{% endwith %}
 
 {% endblock %}
--- a/app/templates/briefs/not_is_supplier_eligible_for_brief_error.html
+++ b/app/templates/briefs/not_is_supplier_eligible_for_brief_error.html
@@ -1,6 +1,6 @@
 {% extends "_base_page.html" %}
 
-{% block page_title %}Not eligible for brief – Digital Marketplace{% endblock %}
+{% block page_title %}Not eligible for opportunity – Digital Marketplace{% endblock %}
 
 {% block main_content %}
 
--- a/tests/app/main/test_briefs.py
+++ b/tests/app/main/test_briefs.py
@@ -36,16 +36,24 @@ processed_brief_submission = {
 }
 
 ERROR_MESSAGE_PAGE_HEADING_APPLICATION = 'You can’t apply for this opportunity'
-ERROR_MESSAGE_NOT_ON_FRAMEWORK_APPLICATION = \
+ERROR_MESSAGE_NO_SERVICE_ON_LOT_APPLICATION = \
+    'You can’t apply for this opportunity because you didn’t say you'\
+    ' could provide services in this category when you applied to the Digital Outcomes and Specialists framework.'
+ERROR_MESSAGE_NO_SERVICE_ON_FRAMEWORK_APPLICATION = \
     'You can’t apply for this opportunity because you’re not a Digital Outcomes and Specialists supplier.'
-ERROR_MESSAGE_DONT_PROVIDE_THIS_SERVICE_APPLICATION = \
-    'You can’t apply for this opportunity because you don’t provide'
+ERROR_MESSAGE_NO_SERVICE_WITH_ROLE_APPLICATION = \
+    'You can’t apply for this opportunity because you didn’t say you'\
+    ' could provide this specialist role when you applied to the Digital Outcomes and Specialists framework.'
 
 ERROR_MESSAGE_PAGE_HEADING_CLARIFICATION = 'You can’t ask a question about this opportunity'
-ERROR_MESSAGE_NOT_ON_FRAMEWORK_CLARIFICATION = \
+ERROR_MESSAGE_NO_SERVICE_ON_LOT_CLARIFICATION = \
+    'You can’t ask a question about this opportunity because you didn’t say you'\
+    ' could provide services in this category when you applied to the Digital Outcomes and Specialists framework.'
+ERROR_MESSAGE_NO_SERVICE_ON_FRAMEWORK_CLARIFICATION = \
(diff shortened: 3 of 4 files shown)


app/templates/briefs/not_is_supplier_eligible_for_brief_error.html
```diff
@@ -1,43 +1,27 @@
 {% extends "_base_page.html" %}
 
-{% block page_title %}Not eligible for brief – Digital Marketplace{% endblock %}
+{% block page_title %}Not eligible for opportunity – Digital Marketplace{% endblock %}
 
 {% block main_content %}
 
-{% if clarification_question %}
+{% with inhibited_action="ask a question about" if clarification_question else "apply for" %}
```
Wynndow: Is the assignment of this variable better done in the view rather than the template; extract as much logic as possible?

risicle: Quite possible, though done this way allows all the _language_ decisions remain together in the template.

What do people think? I'm easy.

@tombye you're a frontendy kinda guy...

pcraig3: Yeah, we tend to keep the words themselves in the template as much as possible.  Seems okay to me to leave it here.

risicle: I think I should start putting deliberately controversial things in my PRs so they get attention and get reviewed. Alternatively I'll start giving them clickbait-y titles...

app/main/views/briefs.py
```diff
@@ -183,12 +183,21 @@ def view_response_result(brief_id):
 
 
 def _render_not_eligible_for_brief_error_page(brief, clarification_question=False):
-    sf = get_supplier_framework_info(data_api_client, brief['frameworkSlug'])
-    on_framework = sf.get('onFramework', False) if sf else False
+    # let's poke the api to discover whether we have any services when queried with less restrictive terms
```
TheDoubleK: I don't think we need a comment here - the intent should be clear enough.

app/main/views/briefs.py
```diff
@@ -183,12 +183,21 @@ def view_response_result(brief_id):
...
+    common_kwargs = {
+        "supplier_id": current_user.supplier_id,
+        "framework": brief['frameworkSlug'],
+    }
+    has_framework_service = bool(data_api_client.find_services(**common_kwargs)["services"])
```
TheDoubleK: This probably is slightly more precise than checking `on_framework` (to catch the case of a supplier having all their services disabled), but the result will be exactly the same in 100% of cases as we haven't ever disabled a DOS service as far as I know.

app/main/views/briefs.py
```diff
@@ -183,12 +183,21 @@ def view_response_result(brief_id):
...
+    has_framework_lot_service = has_framework_service and bool(data_api_client.find_services(
```
TheDoubleK: I don't think this needs the `has_framework_service and`?  There is no way for that to be false and the second part true.

app/templates/briefs/not_is_supplier_eligible_for_brief_error.html
```diff
@@ -1,43 +1,27 @@
...
-        at this location.
-      </p>
-    {% else %}
-      <p>You can’t ask a question about this opportunity because you’re not a {{ framework_name }} supplier.</p>
-    {% endif %}
+    <p>
+      You can’t {{ inhibited_action }} this opportunity because
+      {% if has_framework_service %}
+        you didn’t say you could provide
+        {{ "this specialist role" if has_framework_lot_service else "services in this category" }}
```
TheDoubleK: I don't think this logic is quite right. `has_framework_lot_service` hasn't done any checks about the specialists offered by the supplier - that stuff is in the service data itself.

app/main/views/briefs.py
risicle: Yup - it's what was asked for by sarah.

app/main/views/briefs.py
risicle: Possibly, but a reader of the code who isn't intimately familiar with how things are used doesn't necessarily know that. Putting the `has_framework_service and` clause in there makes it totally unambiguous how that behaves whether you're filled in with the background or not.

app/templates/briefs/not_is_supplier_eligible_for_brief_error.html
risicle: I'll think about this...

risicle: Ah yes, I see what you mean. No, it hasn't done any checks about the role - it's _deduced_ that that is what the problem is from the fact that you have a valid service on the framework & lot, but still when you filter by _all_ the eligibility criteria, it's not showing up. And the role (I _think_) is the only remaining filtering criterion. I've not made this obvious and it's at least comment worthy.

TheDoubleK: Yes I think that's the thing - I'm pretty sure it's doing the right thing in terms of which message to show, but it's not clear from the code why.  I think the old version was marginally more obvious, but still relatively obscure. In both cases there is an assumption that if the user has reached this page then they have already failed the check for eligibility, so there isn't a need to do the full eligibility check again.  That's why the old test for `lot == digital-specialists` was enough to know that they don't provide the right specialist role - because we already know that the test here has failed: https://github.com/alphagov/digitalmarketplace-api/blob/master/app/brief_utils.py#L33

I think this is somewhere where a well-placed comment will help anyone coming back to this code in future.

app/main/views/briefs.py
risicle: Oh looking at it again actually it _is_ fairly obvious whether you know the background or not. Think I was just trying to make things "duh-duh" obvious.

risicle: Oh, no, I know why I did it. It's boolean shortcut logic which allows it to skip the second api request if the first is false.

app/main/views/briefs.py
risicle: Agreed

app/main/views/briefs.py
TheDoubleK: Ah yep OK that makes sense.

tests/app/main/test_briefs.py
```diff
@@ -120,6 +129,7 @@ def test_clarification_question_checks_supplier_is_eligible(self, data_api_clien
         self.login()
         data_api_client.get_brief.return_value = api_stubs.brief(status='live')
         data_api_client.get_brief.return_value['briefs']['frameworkName'] = 'Digital Outcomes and Specialists'
+        data_api_client.get_brief.return_value['briefs']['lot'] = 'digital-specialists'
```
TheDoubleK: `api_stubs.brief` has `digital-specialists` as the lot by default so there's no need to set it here.

To make it explicit for the test (which I agree is nice) you can pass it in to the stubs constructor, eg: `api_stubs.brief(status='live', lot_slug='digital-specialists')`

tests/app/main/test_briefs.py
```diff
@@ -219,39 +229,77 @@ def test_submit_clarification_question_requires_live_brief(self, data_api_client
         assert res.status_code == 404
 
     @mock.patch('app.main.helpers.briefs.send_email')
-    def test_submit_clarification_question_returns_error_page_if_ineligible_supplier_is_not_on_framework(
+    def test_submit_clarification_question_returns_error_page_if_supplier_has_no_services_on_lot(
             self, send_email, data_api_client):
         self.login()
         data_api_client.get_brief.return_value = api_stubs.brief(status='live')
         data_api_client.get_brief.return_value['briefs']['frameworkName'] = 'Digital Outcomes and Specialists'
+        data_api_client.get_brief.return_value['briefs']['lot'] = 'digital-specialists'
```
TheDoubleK: Same here - can do `api_stubs.brief(status='live', lot_slug='digital-specialists')`

tests/app/main/test_briefs.py
```diff
@@ -219,39 +229,77 @@ def test_submit_clarification_question_requires_live_brief(self, data_api_client
...
         assert not data_api_client.create_audit_event.called
 
     @mock.patch('app.main.helpers.briefs.send_email')
-    def test_submit_clarification_question_returns_error_page_if_ineligible_supplier_is_on_framework(
+    def test_submit_clarification_question_returns_error_page_if_supplier_has_no_services_on_framework(
             self, send_email, data_api_client):
         self.login()
         data_api_client.get_brief.return_value = api_stubs.brief(status='live')
         data_api_client.get_brief.return_value['briefs']['frameworkName'] = 'Digital Outcomes and Specialists'
+        data_api_client.get_brief.return_value['briefs']['lot'] = 'digital-specialists'
```
TheDoubleK: And here.

tests/app/main/test_briefs.py
```diff
@@ -219,39 +229,77 @@ def test_submit_clarification_question_requires_live_brief(self, data_api_client
...
+
         res = self.client.post('/suppliers/opportunities/1/ask-a-question', data={
             'clarification-question': "important question",
         })
+        doc = html.fromstring(res.get_data(as_text=True))
+
         assert res.status_code == 400
-        assert ERROR_MESSAGE_PAGE_HEADING_CLARIFICATION in res.get_data(as_text=True)
-        assert ERROR_MESSAGE_NOT_ON_FRAMEWORK_CLARIFICATION in res.get_data(as_text=True)
+        assert doc.xpath('normalize-space(//h1/text())') == ERROR_MESSAGE_PAGE_HEADING_CLARIFICATION
```
TheDoubleK: This `normalize-space` in the XPath is a revelation. Nice!  At some point we should fix up all our clunky matching elsewhere to use it.

TheDoubleK:
Reviewed - the logic here was much more of a headache than I initially thought (maybe it's just me though), sorry about that.

Minor comment left about the test fixture setup, but don't think it's essential to change it - you decide.
Looking good to merge either way now though I think :+1:

tests/app/main/test_briefs.py
risicle: Cool - why do you think it's not used for `frameworkName`?

risicle: @TheDoubleK Doing this reveals an interesting discrepancy - a serialization of a `Brief` includes both a `lot` and `lotSlug` attribute, and I have been using the `lot`. However, `lotSlug` _appears_ to be newer and preferred as `lot` isn't included everywhere (including in the stub). Guess I'll switch to using lotSlug?

tests/app/main/test_briefs.py
risicle: Yeah there are two ways I used it - as a "selector" and as a "getter". I preferred the "getter" way as it allowed me to do a regular `==` comparison in python which pytest does handy things with to reveal text diffs etc. There were some places where that wouldn't work and thats where i made do with using it as a selector.

tests/app/main/test_briefs.py
TheDoubleK: Actually yes the `frameworkName` is also set by default - see https://github.com/alphagov/digitalmarketplace-apiclient/blob/master/dmapiclient/api_stubs.py#L33.  My guess is that in the past when this test was written it didn't.

TheDoubleK: In terms of `lot` vs `lotSlug` - Yes, good spot, `lotSlug` is newer and preferred.

We had `lot` being passed around a lot, and sometimes it would be the slug and sometimes a (perhaps serialized) `Lot` object.  So we decided to move to `lotSlug` where it's only the slug being passed round and save `lot` for full Lot objects.

But then it turned out that some stuff in the ContentLoader is dependent on `lot` to know which things to load, and we haven't got round to changing it, because it would take a lot of time to fix properly.

So for the time being we have both `lot` and `lotSlug` in the serialization, but yes try to stick to using `lotSlug` in any new code.

TheDoubleK: But still I'm pretty sure there are `lot` parameters on methods that are actually lot slugs :/

PR #492, opened 2016-06-13
Fix issues with supplier question tracking

tombye:
Change needed for this story:

https://www.pivotaltracker.com/story/show/119634865

...and fixes problems with the code merged in https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/487.

The regexp that looks for the banner telling users their question was submitted didn't take whitespace into consideration so wasn't finding it. This didn't show up in the unit tests because the HTML mock used had no whitespace in its banner text:

## Banner text in the page

```
\n\s\s\s\sYour question has been sent. The buyer will post your question and their answer on the ‘Dissolve a company’ page.\n
```

## Banner test in our HTML mock

```
Your question has been sent. The buyer will post your question and their answer on the ‘Dissolve a company’ page.
```

### DIFF
--- a/app/assets/javascripts/analytics/_virtualPageViews.js
+++ b/app/assets/javascripts/analytics/_virtualPageViews.js
@@ -10,13 +10,13 @@
   };
 
   GOVUK.GDM.analytics.virtualPageViews = function() {
-    var $flashMessage;
+    var $messageSent;
 
     $('[data-analytics=trackPageView]').each(sendVirtualPageView);
     if (GOVUK.GDM.analytics.location.pathname().match(/^\/suppliers\/opportunities\/\d+\/ask-a-question/) !== null) {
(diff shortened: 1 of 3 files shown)


TheDoubleK:
This detecting the banner messages is starting to feel quite brittle - if the wording of the messages changes then it's unlikely anyone will remember to update the JavaScript at the same time...

Perhaps dummy url parameters are the way to go everywhere? They feel much less likely to break. 

What do you think?

tombye:
I think you're right. It does feel a bit too brittle to fire the analytics based on parts of the page that we could change for design/content reasons.

I've changed the JS so it's based on the data coming into the template instead.

TheDoubleK:
Nice - looks like it will work.  Let's let Andrew get his mitts on it for retesting  :+1:

PR #493, opened 2016-06-13
Bump govuk template

tombye:
Bumps the GOVUK template and fixes the problem this brings in (see https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/338).

https://www.pivotaltracker.com/story/show/120364875

### DIFF
--- a/app/assets/scss/_footer.scss
+++ b/app/assets/scss/_footer.scss
@@ -64,12 +64,12 @@
 
     hr {
       clear: both;
-      margin: $gutter 0 0;
+      margin: $gutter 0;
       border: 1px solid $border-colour;
       border-width: 1px 0 0 0;
 
       @include media(tablet) {
-        margin: 0;
+        margin-top: 0;
       }
     }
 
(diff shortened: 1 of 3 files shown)


pcraig3:
Noticed a couple of things:

### form fields
##### specifically, buttons and the text area get bigger in the new version

update

### footer
##### footer loses bottom padding in new version
update-2

tombye:
@pcraig3 Good point about the `padding-bottom` problem. I've added a fix for that and rebased. 

The button changes were made to bring everyone's padding into line with elements so we should have that in. I can't see any changes to textareas in the CSS this brings in, do you think it's a problem?

pcraig3:
New footer is flush with the sides of our columns.  Old footer wasn't.

footer

@tombye says this is better than before, and I kind of think so too.  All of our other footer links (and most of the text on the page) are left-justified to the edge of the column.

👍  from me.

pcraig3:
Actually, there's a huge difference on firefox somehow.

Basically, all of our links not in body text seem to get smaller.

on-firefox

Not sure how we feel about this.

pcraig3:
Worth noting that it's already made it into the buyer app, so maybe we should just suck it up.

PR #496, opened 2016-06-17
Upgrade toolkit and remove superfluous "home" class from header/footer links

pcraig3:
Upgrading the toolkit brings in the latest govuk frontend and slightly mitigates the problem where [the pricing field layout breaks if you have a scrollbar enabled on your browser](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/266).  

Also got rid of the `home` class which was being applied to a couple of header links and a footer link.
Very well researched.  Read the commit message.

I didn't notice any layout changes applying this pull request, so it should be totally invisible to users.


gemmaleigh:
Nice research!

TheDoubleK:
Let's do it :+1:  

:wave: :house: :wave:

PR #497, opened 2016-06-20
Fix analytics wrapper for window.location.href

tombye:
`window.location.href` is wrapped in an interface to allow easier testing.

When it was implemented, it was done so with this error.


TheDoubleK:
:+1:

PR #498, opened 2016-06-21
Send page pathname rather than whole URL

tombye:
We hypothesise that Google Analytics needs to be sent the `pathname` of the current URL for a pageview rather than the full URL so this enables that change.


pcraig3:
:+1: go for it.

PR #499, opened 2016-06-21
Update text about how suppliers will be notified of G8 result

TheDoubleK:
We are no longer going to give letters to all suppliers, so the text here needs to be changed.


idavidmcdonald:
👍 as soon as the CI tests pass this should be good to go

PR #500, opened 2016-06-24
Update toolkit to 16.0.0

pcraig3:
**Note: needs https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/269 to be merged first.**

Includes a fix for the changed `instruction-list` pattern, which is only used on the framework agreement page.  No other breaking changes.


TheDoubleK:
Looks like this is the only `instruction-list` so all good :+1:

PR #502, opened 2016-07-04
[#121479497] View opportunity response

TheDoubleK:
Completes this story: https://www.pivotaltracker.com/story/show/121479497

New results page looks like this (**NOTE brand new updated screenshot** with 2/3 columns and padding):
result-page-with-padding-on-wide-cells

This introduces quite a lot of DOS-specific logic into the frontend app, which is pretty horrible.  I can't see any way around it for this page though, and it's all in the one "results page" template.

New flow (always to result page) analytics change (`&result=success/fail` parameter) has been discussed with Ash and he just needs to know when this gets released.

Words on the page have been iterated with @ralph-hawkins and @superroz quite a lot, so try not to find anything else.


TheDoubleK:
You could have mentioned it last week when I sent all the screenshots :)

I'll see if I can work it out now.

ralph-hawkins:
You’re right, sorry I missed it. It was in the original pivotal story.

app/templates/briefs/_brief_response_data.html
```diff
@@ -0,0 +1,29 @@
+{% import "toolkit/summary-table.html" as summary %}
+{% for section in content.summary(brief_response) %}
+    {% if not (section.id == 'your-nice-to-have-skills-and-experience' and not section.get_question('niceToHaveRequirements').boolean_list_questions) %}
```
allait: Is there any way to avoid this?
Why do we need both conditions? I think I don't really understand what we're checking.
Is there a way to remove the section from display manifest?

TheDoubleK: Exactly - if there are no questions without this we get just the heading displayed.

TheDoubleK: I know it's ugly but I couldn't find any other good way round it... it's tricky because it's not an empty section - it still has the `niceToHave` question, but the question happens to not have any `boolean_list_questions` set in it.

allait:
I'm a bit concerned about how tied this page is to our current question options etc (like a special case for 'an' before 'Interview'). And it feels like we need to think about making boolean questions less scary.

But it looks like we need this for labs tomorrow and I don't see any obvious quick ways of improving this, so 👍

TheDoubleK:
I agree, it's not really ideal at all.... sorry.

PR #503, opened 2016-07-05
Don't assume there is an evaluation type

TheDoubleK:
The default evaluation type isn't passed in (as I thought it would be), so for briefs without and additional evaluation types specified the page breaks with 500.  This prevents errors, but still needs more work to show the default evaluation type in the list.


allait:
Oh yeah. The mandatory option is added by `.summary`, I don't know what would be a good way to get it into briefs.

Let's get this in for testing tomorrow, but only to staging 👍

TheDoubleK:
Cool, yeah it looks like a bit of a headache to get it in, will have to call `summary` on the brief though I think.

PR #504, opened 2016-07-06
Display the default evaluation type for brief responded to

TheDoubleK:
Should hopefully be the final bugfix for: https://www.pivotaltracker.com/story/show/121479497

The default evaluation types defined in the frameworks content are not stored in the brief object JSON at all, and are only added to the brief when `.summary` is called in the content loader.
So in order to be able to display all evaluation methods that will be used, including the defaults, we need to load the `edit_brief` manifest and call `.summary` with the current brief.

Here you can see the default "a work history" is displayed, despite not being stored in the brief data:
result-page-with-default-evaluation-type-shown


allait:
👍

PR #506, opened 2016-07-14
[#121577345] Sign and return your framework agreement

idavidmcdonald:
Built multipage flow to handle signing and returning your framework agreement

## Contract start
- Contains table of which lots you have passed and which you did not apply for
- Instructions for how to return your framework agreement
- The link to download your signature page is to a unique file for each company (TO BE COMPLETED)

screen shot 2016-07-14 at 16 27 30

## Signer details
- Form to get `signerName` and `signerRole`
- Form is validated by `form.validate_on_submit` after which values are sent to the API (which also does its own validation)
- If `signerName` and `signerRole` already exist in `agreementDetails` then these are prefilled in the form
- If filename exists in session, when the form is submitted you are taken straight to the contract review page as you have likely been editing your signer details

screen shot 2016-07-14 at 16 27 42

## Signature upload
- Upload PDF, JPG or PNG file
- Validation checks that file exists, is of correct format, is not empty and is smaller than 5MB
- File is uploaded to s3 and a new file is uploaded every time. Older files are archived.
- The filename of the uploaded file is stored in the session to be shown on the contract review page
- If you have previously uploaded a file in this session and revisit this page then you will see the filename and upload time for the previously uploaded file (include picture)
- If you have previously uploaded a file but your session has since expired and revisit this page, you will only see the upload time for the previously uploaded file (include picture)
- If you have previously uploaded a file and revisit this page and do not wish to reupload a new one then clicking `Save and continue` will not require you to have chosen a file

screen shot 2016-07-14 at 16 29 32

***
### 1. File uploaded, session exists
screen shot 2016-07-14 at 16 30 39
***

### 2. File uploaded, session has expired
image
***

## Contract review
- If you do not have `signerName`, `signerRole` and a file in s3 when loading this page, you will be aborted with a 400
- User is required to confirm they have the authority to return the framework agreement
- On submit, if form is valid and API call to mark `agreementReturned` as true succeeds then we send a confirmation email. Emails are sent to the current logged in user and to the primary contact on the declaration. If these people are the same (indicated by the same email address) then only one email is sent.

### 1. File uploaded, session exists

screen shot 2016-07-14 at 16 30 46
***
### 2. File uploaded, session has expired

image

***

## Contract submitted (replaces framework dashboard)
- Shows details of the returned agreement (who signed it, who it was submitted by and when it was submitted)
- A link is provided to download the previously uploaded framework agreement signature page
- A link is provided to read the full standard framework agreement

screen shot 2016-07-14 at 16 30 53

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -577,6 +577,7 @@ def upload_framework_agreement(framework_slug):
     return redirect(url_for('.framework_agreement', framework_slug=framework_slug))
 
 @main.route('/frameworks/<framework_slug>/signer-details', methods=['GET'])
+@login_required
 def signer_details(framework_slug):
     framework = get_framework(data_api_client, framework_slug)
     return_supplier_framework_info_if_on_framework_or_abort(data_api_client, framework_slug)
@@ -597,6 +598,7 @@ def signer_details(framework_slug):
 
 
 @main.route('/frameworks/<framework_slug>/signer-details', methods=['POST'])
+@login_required
 def submit_signer_details(framework_slug):
     framework = get_framework(data_api_client, framework_slug)
     return_supplier_framework_info_if_on_framework_or_abort(data_api_client, framework_slug)
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -605,7 +605,9 @@ def signer_details(framework_slug):
                 extra={
                     'signerName': form['signerName'].data,
                     'signerRole': form['signerRole'].data,
-                    'errors': ",".join(chain.from_iterable(form.errors.values()))})
+                    'errors': ",".join(chain.from_iterable(form.errors.values()))
+                }
+            )
 
             error_keys_in_order = [key for key in question_keys if key in form.errors.keys()]
             form_errors = [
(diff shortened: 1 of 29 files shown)


app/main/views/frameworks.py
```diff
@@ -85,17 +85,46 @@ def framework_dashboard(framework_slug):
...
+            key_list, framework_slug, "communications/updates/"
+        )
+    }
+
+    # if framework_agreement_version and application_made, show new framework dashboard template
+    if supplier_is_on_framework and framework['frameworkAgreementVersion'] and supplier_framework_info['agreementReturned']:  # noqa
+        agreements_bucket = s3.S3(current_app.config['DM_AGREEMENTS_BUCKET'])
+        signature_page = get_most_recently_uploaded_agreement_file_or_none(agreements_bucket, framework_slug)
+
+        # The default view (download_agreement_file) uses the DM_SUBMISSIONS_BUCKET, which doesn't work
```
risicle: Probably still better to do this formatting in `config.py` than have the s3 host hard coded in here.

app/main/views/frameworks.py
```diff
@@ -85,17 +85,46 @@ def framework_dashboard(framework_slug):
...
+    last_modified = {
+        'supplier_pack': get_last_modified_from_first_matching_file(
+            key_list, framework_slug, "communications/{}".format(supplier_pack_filename)
+        ),
+        'supplier_updates': get_last_modified_from_first_matching_file(
+            key_list, framework_slug, "communications/updates/"
+        )
+    }
+
+    # if framework_agreement_version and application_made, show new framework dashboard template
```
risicle: Where is `application_made` used here to determine this?

app/main/views/frameworks.py
```diff
@@ -85,17 +85,46 @@ def framework_dashboard(framework_slug):
...
+    if supplier_is_on_framework and framework['frameworkAgreementVersion'] and supplier_framework_info['agreementReturned']:  # noqa
+        agreements_bucket = s3.S3(current_app.config['DM_AGREEMENTS_BUCKET'])
+        signature_page = get_most_recently_uploaded_agreement_file_or_none(agreements_bucket, framework_slug)
+
+        # The default view (download_agreement_file) uses the DM_SUBMISSIONS_BUCKET, which doesn't work
+        signature_page_url = get_signed_url(agreements_bucket, signature_page['path'], 'https://{}.s3-eu-west-1.amazonaws.com'.format(current_app.config['DM_AGREEMENTS_BUCKET']))  # noqa
```
allait: We shouldn't hardcode the S3 URL here. It should be `current_app.config['DM_ASSETS_URL']` to go through nginx in live environments.

app/main/views/frameworks.py
```diff
@@ -566,3 +595,219 @@ def upload_framework_agreement(framework_slug):
...
+            )
+
+            # If they have already uploaded a file then let them go to straight to the contract review
+            # page as they are likely editing their signer details
+            if session.get('signature_page'):
+                return redirect(url_for(".contract_review", framework_slug=framework_slug))
+
+            return redirect(url_for(".signature_upload", framework_slug=framework_slug))
+        else:
+            current_app.logger.warning(
```
allait: I don't think we need to log this. We display form errors to the user anyway, and we generally don't log form data from unsuccessful requests.

app/main/views/frameworks.py
```diff
@@ -566,3 +595,219 @@ def upload_framework_agreement(framework_slug):
...
+                current_user.supplier_id,
+                '{}{}'.format(SIGNED_AGREEMENT_PREFIX, get_extension(request.files['signature_page'].filename))
+            )
+            agreements_bucket.save(
+                upload_path,
+                request.files['signature_page'],
+                acl='private'
+            )
+
+            session['signature_page'] = request.files['signature_page'].filename
```
allait: I thought we decided not to store the filename and show the view link instead

idavidmcdonald: Decision reversed after watching user testing (aware Cath said otherwise in show and tell yesterday but she had not been updated on that decision when presenting)

app/main/views/frameworks.py
```diff
@@ -566,3 +595,219 @@ def upload_framework_agreement(framework_slug):
...
+            ]
+            if supplier_framework['declaration']['primaryContactEmail'] != current_user.email_address:
+                email_recipients.append(
+                    {
+                        'name': current_user.name,
+                        'to_address': current_user.email_address
+                    }
+                )
+
+            for recipient in email_recipients:
```
allait: We can send Mandrill emails to multiple recipients with one API call, so we could get rid of the loop and simplify this a bit if we replace `Dear {{ name }}` with `Dear supplier`, which we've already used in a few existing email templates.

app/main/views/frameworks.py
```diff
@@ -566,3 +595,219 @@ def upload_framework_agreement(framework_slug):
...
+
+            flash(
+                'Your framework agreement has been returned to the Crown Commercial Service to be countersigned.',
+                'success'
+            )
+
+            return redirect(url_for(".framework_dashboard", framework_slug=framework_slug))
+
+        else:
+            current_app.logger.warning(
```
allait: Again, since this is form validation errors, I don't think we need to log it

app/main/views/frameworks.py
pcraig3: Comment was incorrect, not the code.  Sorry for the confusion.

app/main/views/frameworks.py
pcraig3: Killed it 🔪

app/main/views/frameworks.py
```diff
@@ -85,17 +86,43 @@ def framework_dashboard(framework_slug):
     if countersigned_framework_agreement_exists_in_bucket(framework_slug, current_app.config['DM_AGREEMENTS_BUCKET']):
         countersigned_agreement_file = COUNTERSIGNED_AGREEMENT_FILENAME
 
+    application_made = supplier_is_on_framework or (len(complete_drafts) > 0 and declaration_status == 'complete')
```
allait: It looks like we've done a similar check before. @TheDoubleK do you remember if there's a reason to keep the draft checks here? Drafts could be removed in the future and I thought we're settings `onFramework` flag for all suppliers now.

app/main/views/frameworks.py
```diff
@@ -566,3 +593,178 @@ def upload_framework_agreement(framework_slug):
...
+    signature_page = get_most_recently_uploaded_agreement_file_or_none(agreements_bucket, framework_slug)
+
+    # if supplier_framework doesn't have a name or a role or the agreement file, then redirect to first page
+    if not (
+        supplier_framework['agreementDetails'] and
+        supplier_framework['agreementDetails'].get('signerName') and
+        supplier_framework['agreementDetails'].get('signerRole') and
+        signature_page
+    ):
+        return render_template("errors/500.html"), 400
```
allait: Let's make this a 404.

app/templates/frameworks/contract_review.html
```diff
@@ -0,0 +1,117 @@
...
+  {% endwith %}
+{% endblock %}
+
+{% block main_content %}
+
+<div class="grid-row">
+  <div class="column-two-thirds">
+    {%
+      with
+      heading = "Check the details you’ve given before returning the signature page for {}".format(supplier_framework.declaration.nameOfOrganisation),
```
allait: Should be `heading = "... {} ..." | format(supplier_framework.declaration.nameOfOrganisation)` to avoid unicode issues

app/templates/frameworks/signer_details.html
```diff
@@ -0,0 +1,73 @@
...
+    {% include "toolkit/breadcrumb.html" %}
+  {% endwith %}
+{% endblock %}
+
+{% block main_content %}
+<div class="grid-row">
+  <div class="column-two-thirds">
+    {%
+    with
+      heading = "Details of the person who is signing on behalf of {}".format(supplier_framework.declaration.nameOfOrganisation),
```
allait: `| format`

tests/app/main/test_frameworks.py
```diff
@@ -1642,6 +1709,580 @@ def test_dos_drafts_list_with_closed_framework(self, count_unanswered, data_api_
...
+                "/suppliers/frameworks/g-cloud-8/contract-review",
+                data={}
+            )
+            assert res.status_code == 400
+            page = res.get_data(as_text=True)
+            assert not send_email.called
+            assert "You must confirm you have the authority to return the agreement" in page
+
+    @mock.patch('dmutils.s3.S3')
+    @mock.patch('app.main.frameworks.send_email')
```
allait: Wrong path

tests/app/main/test_frameworks.py
```diff
@@ -1642,6 +1709,580 @@ def test_dos_drafts_list_with_closed_framework(self, count_unanswered, data_api_
...
+
+            res = self.client.post(
+                "/suppliers/frameworks/g-cloud-8/contract-review",
+                data={
+                    'authorisation': 'I have the authority to return this agreement on behalf of company name'
+                }
+            )
+
+            # Delcaration primaryContactEmail and current_user.email_address are different so expect two recipients
+            send_email.assert_called_once(
```
allait: Should be `assert_called_once_with`

tests/app/main/test_frameworks.py
```diff
@@ -1642,6 +1709,580 @@ def test_dos_drafts_list_with_closed_framework(self, count_unanswered, data_api_
...
+            }]
+
+            res = self.client.post(
+                "/suppliers/frameworks/g-cloud-8/contract-review",
+                data={
+                    'authorisation': 'I have the authority to return this agreement on behalf of company name'
+                }
+            )
+
+            send_email.assert_called_once(
```
allait: `assert_called_once_with`

tests/app/main/test_frameworks.py
```diff
@@ -1642,6 +1709,580 @@ def test_dos_drafts_list_with_closed_framework(self, count_unanswered, data_api_
...
+                data={
+                    'authorisation': 'I have the authority to return this agreement on behalf of company name'
+                }
+            )
+
+            assert res.status_code == 500
+            assert not send_email.called
+
+    @mock.patch('dmutils.s3.S3')
+    @mock.patch('app.main.frameworks.get_supplier_framework_info')
```
allait: `app.main.frameworks` doesn't exist

app/main/views/frameworks.py
```diff
@@ -566,3 +593,178 @@ def upload_framework_agreement(framework_slug):
...
+    form_errors = None
+
+    if request.method == 'POST':
+        if form.validate_on_submit():
+            data_api_client.register_framework_agreement_returned(
+                current_user.supplier_id, framework_slug, current_user.email_address, current_user.id
+            )
+
+            email_recipients = [supplier_framework['declaration']['primaryContactEmail']]
+            if supplier_framework['declaration']['primaryContactEmail'] != current_user.email_address:
```
risicle: considering de-case-sensitivizing this?

in theory the _pre-domain part_ of an email can be case sensitive, but in practise almost no mail services treat differently-cased mailboxes differently.

allait:
👍

tests/app/main/test_frameworks.py
```diff
@@ -609,6 +624,58 @@ def test_upload_message_if_agreement_is_not_returned(self, data_api_client):
...
+                ('Specialist Cloud Services', 'Pass'),
+            ]
+
+            res = self.client.get("/suppliers/frameworks/g-cloud-8/agreement")
+            doc = html.fromstring(res.get_data(as_text=True))
+
+            assert res.status_code == 200
+
+            lots_and_statuses = []
+            lot_table_rows = doc.xpath('//*[@id="content"]//table/tbody/tr')
```
risicle: Incidental sidenote: did you know about xpath's `normalize-space()`? https://developer.mozilla.org/en-US/docs/Web/XPath/Functions/normalize-space

You can see how I've used it in `test_briefs.py`

Can be used to write robustify your tests and quicker to work with than external `strip()` monkeying...

PR #508, opened 2016-07-18
framework dashboard guidance links: minor fix for html

risicle:
was previously outputting an `ul` directly within another `ul` in the live/standstill state. instead put both possible blocks in a `li.browse-list-item`.

spotted this while digging in the wrong place for another story. all this patch does is move the `li.browse-list-item` outside the `{% if %}` block.

### DIFF
--- a/app/templates/frameworks/_guidance_links.html
+++ b/app/templates/frameworks/_guidance_links.html
@@ -1,48 +1,47 @@
-{% if framework.status in ['standstill', 'live'] and countersigned_agreement_file %}
-<ul class="list-no-bullet">
-  <li>
-    <a href="{{ url_for('.download_agreement_file', framework_slug=framework.slug, document_name=countersigned_agreement_file) }}">
-      <span>Download your countersigned framework agreement (.pdf)</span>
-    </a>
-  </li>
-
-  <li>
-    <a href="{{ url_for('.download_agreement_file', framework_slug=framework.slug, document_name=result_letter_filename) }}">
-      <span>Download your application result letter (.pdf)</span>
-    </a>
-  </li>
-
-  <li>
-    <a href="{{ url_for('.download_supplier_file', filepath=supplier_pack_filename, framework_slug=framework.slug) }}">
-      <span>Download guidance and legal documentation (.zip)</span>
-    </a>
-    {% if last_modified.supplier_pack %}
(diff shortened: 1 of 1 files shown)


pcraig3:
So I've had a look and you're basically right, but I think we want to add a title as well.

### Now
Currently, our bad HTML lets us get away without a title because it doesn't look like our guidance links are in a separate section.

screen shot 2016-07-18 at 15 09 21

### Yours
Yours fixes the HTML, but now they are definitely in their own section (and all of our other sections have titles).

screen shot 2016-07-18 at 15 16 45

### What I think we need
Add an `h2` that says "Documents, updates and questions", and I think we're in the clear.
If you're feeling really punctilious, you could even ask Roz or Ralph or Laurence about if that's the right title.
screen shot 2016-07-18 at 15 17 22

risicle:
Ah yes nice digging.

I might just try and invoke them through the medium of github transcendence @superroz @ralph-hawkins  :ghost:

superroz:
New heading looks OK. Let's do it. 'Documents' is a bit non-specific but any more detail would end up looking ridiculous.

This may be another story but it's weird that under 'You submitted:', it says: '2 Infrastructure as a Service service'. Is there a way of making 'service' plural when there's more than 1. '2 Infrastructure as a Service services' would be better. Still weird because of repetition of 'service' but better.

pcraig3:
@superroz, that's a good point.  [There's already code there that appears to do this](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/templates/frameworks/_submitted_services.html#L11), but it somehow not being triggered.  

Probably a new story/bug if there's a time-consuming reason it's not working, but it's possible that it's a really easy fix.

risicle:
Re the "services" bug, I've located it and it's a mega easy fix...

But first I guess I've got to create a story for it, get the story triaged, approved..............

risicle:
Or maybe I'll just fix the damn thing.

superroz:
If it's super easy, I vote fix it :)

sarahonetui:
Yep, if it's super easy, do it.

risicle:
I'll get someone to thumbsup this PR first because the patches will collide and be annoying otherwise.

pcraig3:
👍

PR #509, opened 2016-07-18
[#121577345] QA fixes frontend and copy fixes

idavidmcdonald:
Fixes three issues thrown up in testing by @andrewchong 

- Add missing "." 
- Replace link to DOS with G-Cloud 8 link
- Remove `signature_page` from session after a framework agreement as a) no longer needed and b) to minimise the risk of old incorrect data being shown to a user

### DIFF
--- a/app/templates/frameworks/contract_submitted.html
+++ b/app/templates/frameworks/contract_submitted.html
@@ -82,7 +82,7 @@
 
       <div class="dmspeak">
         <div class="isolated-text">
-          <p><a href="https://www.gov.uk/government/publications/digital-outcomes-and-specialists-framework-agreement">Read the standard framework agreement</a></p>
+          <p><a href="https://www.gov.uk/government/publications/g-cloud-8-framework-agreement">Read the standard framework agreement</a></p>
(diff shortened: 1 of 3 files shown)


app/templates/frameworks/contract_submitted.html
```diff
@@ -82,7 +82,7 @@ <h2 class="page-subheading">Your framework agreement</h2>
 
       <div class="dmspeak">
         <div class="isolated-text">
-          <p><a href="https://www.gov.uk/government/publications/digital-outcomes-and-specialists-framework-agreement">Read the standard framework agreement</a></p>
+          <p><a href="https://www.gov.uk/government/publications/g-cloud-8-framework-agreement">Read the standard framework agreement</a></p>
```
TheDoubleK: It would be nice to pull all framework-specific things on this page from the `framework` object that's passed in, as this page is going to cater for multiple frameworks in the future.  But that would need extra stuff added to the database and API that we're not going to do today...

So :+1: as a temporary fix.

TheDoubleK:
Sweet - let's do it :+1:
If only we could hurry Travis somehow...

PR #510, opened 2016-07-18
Fix missing `supplier_pack_filename` variable from template

idavidmcdonald:
`supplier_pack_filename` had been forgotten from our variables passed through to the template, causing the 'Download guidance and legal documentation (.zip)' link to break.


pcraig3:
👍  once it's green.

PR #511, opened 2016-07-18
Add audit event for signature page uploads

pcraig3:
We want to be able to track whenever changes are made to signature pages.
Through the API, we can track all changes to things we store in agreementDetails, but we don't know anything about file uploads.
This audit event tells us when files are uploaded, who did the uploading, and what their original filenames were.


allait:
👍

PR #512, opened 2016-07-21
Require login to download responses to a brief

TheDoubleK:
Users must be logged in to download brief responses.

Previously if no user was logged in then this route gave a 500 error as no supplier ID could be found for the anonymous user.


pcraig3:
👍  oh duh :eyeroll:

PR #514, opened 2016-07-22
framework dashboard: fix pluralization of lot units

risicle:
by rehashing the way we pass lot data to templates. this code seems to have missed out on the addition of `unitSingular` and `unitPlural` to the lots table, so use that. rather than doing laborious cherry-picking of lot dict items we may as well just pass copies of the api-supplied lot dicts themselves with our added data annotated on.

this fell out of #508


pcraig3:
Yeah man, everything still works. 

Go for it.

👍

pcraig3:
Look at me doing the wrong thing 😈

PR #515, opened 2016-08-01
Always use https:// protocol for bower dependencies

tombye:
git:// is unauthenticated (and we don't pin with commit SHA) so https:// should be preferred.

Since we're downloading from GitHub we need to add `.git` to access the Git endpoint over HTTPS.

This is a duplicate of the following pull requests:
- [#367 on the buyer app](https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/367/commits)
- [#199 on the admin app](https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/199)


TheDoubleK:
:+1:

PR #517, opened 2016-08-03
[#127062871] Track account creation using GA virtual page views

idavidmcdonald:
Currently, we are not able to tell using GA how many user accounts have been created. We could try to use a page view of the URL after an account is created to indicate this, however the URL is also accessible from other user flows and will not therefore be an accurate count of how many user accounts have been created. 

Therefore, as part of [this pivotal story](https://www.pivotaltracker.com/story/show/127062871), we wish to send a GA virtual page view every time a supplier account is successfully created. By embedding

`data-analytics="trackPageView" data-url="/suppliers/vpv/?account-created=true"`

into our template, we will send the required virtual page view on the success page(functionality [already exists](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/assets/javascripts/analytics/_virtualPageViews.js) to do this automatically when the required `data` attributes are provided).

This tracking tag will only be embedded if they have come through the account creation process (indicated by setting a flash message).

Note, a similar process [has been created](https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/374) for the creation of buyer accounts.

### DIFF
--- a/app/main/views/login.py
+++ b/app/main/views/login.py
@@ -90,6 +90,7 @@ def submit_create_user(encoded_token):
                 "auth/create_user_error.html",
                 token=None), 400
 
+        flash('account-created', 'flag')
         return redirect(url_for('.dashboard'))
 
 
--- a/app/templates/suppliers/dashboard.html
+++ b/app/templates/suppliers/dashboard.html
@@ -15,8 +15,7 @@
 {% endblock %}
 
 {% block main_content %}
-
-  {% with messages = get_flashed_messages(with_categories=True) %}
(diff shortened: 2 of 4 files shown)


pcraig3:
Cool, looks good and the existing JS tests already cover the new thing we've added here.

I'm going to hold off on thumbsing this one up until we've got a more definitive answer about sending double pageviews, [which I described in more detail in a comment in the buyer app](https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/374#issuecomment-237620947).

It seems a bit hypocritical of me because I already gave the other one a thumbs up, but I kind of think they'll both be fine.

pcraig3:
Also 👍

PR #519, opened 2016-08-11
Analytics to tell if a supplier is eligible to apply for a brief

idavidmcdonald:
### Overview
[Pivotal story](https://www.pivotaltracker.com/story/show/126116679)

When a supplier clicks to apply for a brief they can be one of 4 states:
1. are eligible and can apply
2. are not eligible as they are not on the framework
3. are not eligible as they are not on the lot
4. are not eligible as they can not provide the specialist role

After clicking the button to apply for a brief, in all of the above cases, they are taken to the same URL. In the success case (eligible) they are shown the form to apply. In the failure case (not eligible) they are shown the relevant message for why they can not apply to this brief. We wish to add analytics to this page so we can track the 4 cases. This is done by setting a Google Analytics custom dimension with a value related to one of the 4 cases.

### Approach
- For the non eligible case, the `not_is_supplier_eligible_for_brief_error` template is used. We add a `data-reason` attribute to the page that stores the reason for ineligibility such as "supplier-not-on-dos". 

- For the success case (where the supplier is eligible to apply) we add nothing to the template.

- When visiting the route to apply for a brief, we check the html content to see if a `data-reason` is present. If so, we send that value. If not then we can assume that they are viewing the page to apply and are eligible, so we send "supplier-able-to-apply". Note, the javascript acts as the gateway by checking the page url to make sure we only send this custom dimension on the required route (as our templates may be shared by several pages, we may have `data-reason`s present on different routes).

### DIFF
--- a/app/assets/javascripts/analytics/_pageViews.js
+++ b/app/assets/javascripts/analytics/_pageViews.js
@@ -1,7 +1,23 @@
 (function (GOVUK) {
   GOVUK.GDM.analytics.pageViews = {
     'init': function () {
+      this.setCustomDimensions()
       GOVUK.analytics.trackPageview();
+    },
+
+    'setCustomDimensions': function() {
+      var eligibility;
+      
+      if (/^\/suppliers\/opportunities\/(\d+)\/responses\/create$/.test(GOVUK.GDM.analytics.location.pathname())) {
+        eligibility = $('[data-reason]').first().data('reason');
+
+        // If we do not find a reason for ineligibility then we will assume that they are able to apply
+        if (!eligibility) {
+          eligibility = 'supplier-able-to-apply';
+        }
+        
+        GOVUK.analytics.setDimension(25, eligibility);
+      }
     }
   };
 })(window.GOVUK);
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -208,7 +208,12 @@ def _render_not_eligible_for_brief_error_page(brief, clarification_question=Fals
     has_framework_lot_service = has_framework_service and bool(data_api_client.find_services(
         **dict(common_kwargs, lot=brief['lotSlug'])
     )["services"])
-    # if has_framework_lot_service is true, we can deduce that the problem is that the roles don't match.
+
+    if has_framework_service:
+        # if has_framework_lot_service is true, we can deduce that the problem is that the roles don't match.
+        reason = "supplier-not-on-role" if has_framework_lot_service else "supplier-not-on-lot"
+    else:
+        reason = "supplier-not-on-{}".format(brief['frameworkFramework'])
 
     return render_template(
         "briefs/not_is_supplier_eligible_for_brief_error.html",
@@ -217,5 +222,6 @@ def _render_not_eligible_for_brief_error_page(brief, clarification_question=Fals
         has_framework_lot_service=has_framework_lot_service,
         framework_name=brief['frameworkName'],
         lot=brief['lotSlug'],
+        reason=reason,
         **dict(main.config['BASE_TEMPLATE_DATA'])
     ), 400
--- a/spec/javascripts/unit/AnalyticsSpec.js
+++ b/spec/javascripts/unit/AnalyticsSpec.js
@@ -213,4 +213,52 @@ describe("GOVUK.Analytics", function () {
     });
 
   });
+
+  describe("Supplier eligible to apply for brief", function() {
+    var $notOnFrameworkString = $('<div class="grid-row" data-reason="supplier-not-on-dos">');
+    
+    afterEach(function () {
+      $notOnFrameworkString.remove();
+    });
+    
+    it('should send custom dimension if data-reason', function() {
+      spyOn(GOVUK.GDM.analytics.location, "pathname")
+          .and
+          .callFake(function () {
+            return "/suppliers/opportunities/1/responses/create";
+        });
+
+      $(document.body).append($notOnFrameworkString);
+
+      window.GOVUK.GDM.analytics.pageViews.init();
+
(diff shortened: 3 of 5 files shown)


app/main/views/briefs.py
```diff
@@ -199,7 +199,7 @@ def view_response_result(brief_id):
     )
 
 
-def _render_not_eligible_for_brief_error_page(brief, clarification_question=False):
+def _render_not_eligible_for_brief_error_page(brief, clarification_question=False, custom_dimension=False):
```
tombye: I suggest changing this extra parameter from the boolean into a string containing the reason the supplier isn't eligible, ie: `ineligibility_reason=None|'supplier-not-on-role'|'supplier-not-on-lot'|'supplier-not-on-framework'`.

I think doing this would mean the information passed through isn't only relevant to analytics and we could remove the block that sets the `dimension` variable in the template (line 7 in `app/templates/briefs/not_is_supplier_eligible_for_brief_error.html`).

app/assets/javascripts/analytics/_pageViews.js
```diff
@@ -1,7 +1,33 @@
...
-    }
+    },
+
+    'setCustomDimensions': function() {
+      var customDimensionData,
+          dimensionValToBeSet,
+          h1Val;
+      
+      if (this.createOpportunityResponseRegex.test(GOVUK.GDM.analytics.location.pathname())) {
+        customDimensionData = $('[data-analytics=customDimension]').first().data('custom-dimension');
```
tombye: It feels like we should be setting another attribute for the dimension index here.

I'm basing this on the assumption that we're following the precedent of the `data-analytics` attribute marking the type of analytics information we set with `data-analytics="trackPageView"`. If that's the case, we need to follow that convention of all the information needed for the analytics event being available in the data attributes on the node. At the moment, this assumes a custom dimension on this page will always be set on index 25.

Alternatively, this could be rewritten to be more specific to our case, for example:

```
<div class="grid-row" data-analytics="supplierNotEligible" data-reason="{{ ineligibility_reason }}">
```

app/assets/javascripts/analytics/_pageViews.js
```diff
@@ -1,7 +1,33 @@
...
+        h1Val = $('h1').first().text();
+
+        // If we find a custom dimension data attribute then we will send that as our custom dimension
+        if (customDimensionData) {
+          dimensionValToBeSet = customDimensionData;
+        // Else check if the page title text indicates the supplier is indeed able to apply
+        } else if (h1Val.indexOf("You can’t {{ inhibited_action }} this opportunity") == -1 && h1Val.indexOf("Apply for ") >= 0) {
```
tombye: It doesn't seem like this condition is needed. From my understanding of the different page states possible, those without the custom dimension can be assumed to be those where a supplier is eligible.

If that's the case, we also don't need to cache the dimension in `dimensionValToBeSet` as we can return the function as soon as a dimension is set.

spec/javascripts/unit/AnalyticsSpec.js
```diff
@@ -213,4 +213,94 @@ describe("GOVUK.Analytics", function () {
...
+          .callFake(function () {
+            return "/suppliers/opportunities/1/responses";
+        });
+
+      window.GOVUK.GDM.analytics.pageViews.setCustomDimensions();
+
+      expect(window.ga.calls.any()).toEqual(false);
+    });
+
+    it('should only match regex for create response route', function() {
```
tombye: These are valuable things to test but could we do this by faking calls to `GOVUK.analytics.location.pathname` instead? I think there's some value in not testing the internals of our analytics code if possible (mainly informed by the talk we saw recently on this - https://www.youtube.com/watch?v=URSWYvyc42M).

app/assets/javascripts/analytics/_pageViews.js
```diff
@@ -1,7 +1,23 @@
...
   GOVUK.GDM.analytics.pageViews = {
     'init': function () {
+      this.setCustomDimensions()
       GOVUK.analytics.trackPageview();
+    },
+
+    'setCustomDimensions': function() {
+      var eligibility;
+      
+      if (/^\/suppliers\/opportunities\/(\d+)\/responses\/create$/.test(GOVUK.GDM.analytics.location.pathname())) {
```
pcraig3: Hey @tombye, how do we feel about doing hardcoded URL strings in our Javascript?  
This was [something that @ris brought up before](https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/380#discussion_r74052889) when you were away, and I don't really know how we feel about it.

app/main/views/briefs.py
```diff
@@ -208,7 +208,12 @@ def _render_not_eligible_for_brief_error_page(brief, clarification_question=Fals
     has_framework_lot_service = has_framework_service and bool(data_api_client.find_services(
```
pcraig3: I know it's not technically to do with this story, but it looks to me like this check which calls `data_api_client.find_services()` twice could be more efficient.  Seems like we could just get the services once and then see if any of them are in the right lot.

app/assets/javascripts/analytics/_pageViews.js
tombye: It would be better to have one single reference for the Python and JavaScript to use (perhaps a config file) but at the moment it's all kept inside the app so we can't do this.

@risicle is right and I've done projects before where the JavaScript and server-side language both point at a single data source (JSON usually) for shared information.

Might be worth looking at in future as we can output all the paths registered to the app so it could be a build task which produces a JSON file the JavaScript could reference.

tombye:
Looks good to me, nice work 👍

PR #522, opened 2016-08-19
[#127807161] Display contract variation page

TheDoubleK:
For this story: https://www.pivotaltracker.com/story/show/127807161

Depends on API changes here: 
 - [x] https://github.com/alphagov/digitalmarketplace-api/pull/435

~~There are minor content updates to come on Monday morning, but worth getting this up now for people to take a look.~~

Content updated to latest from document on 21/8/2016, screenshots here updated.

UPDATE: New validation message screenshot updated.

## Variation not yet signed
g-cloud 8 unsigned contract variation

## Error message if you don't check the box
screen shot 2016-08-22 at 13 38 38

## Success message if you check the box and submit
screen shot 2016-08-21 at 10 35 01

## Email sent, which looks like this in G-Mail:
screen shot 2016-08-21 at 10 39 30

# Once agreed the page looks like this
g-cloud 8 signed contract variation

### DIFF
--- a/tests/app/main/test_frameworks.py
+++ b/tests/app/main/test_frameworks.py
@@ -2337,7 +2337,7 @@ class TestContractVariation(BaseApplicationTest):
             status='live',
             framework_agreement_version='3.1'
         )
-        self.g8_framework['frameworks']['variations'] = {1: {"createdAt": "2018-08-16"}}
+        self.g8_framework['frameworks']['variations'] = {"1": {"createdAt": "2018-08-16"}}
 
         with self.app.test_client():
             self.login()
@@ -2352,16 +2352,6 @@ class TestContractVariation(BaseApplicationTest):
         assert_equal(
             len(doc.xpath('//h1[contains(text(), "G-Cloud 8: proposed contract variation")]')), 1)
 
-    def test_get_page_renders_with_default_variation_number(self, data_api_client):
-        data_api_client.get_framework.return_value = self.g8_framework
-        data_api_client.get_supplier_framework_info.return_value = self.good_supplier_framework
-        res = self.client.get("/suppliers/frameworks/g-cloud-8/contract-variation")
-
-        assert_equal(res.status_code, 200)
-        doc = html.fromstring(res.get_data(as_text=True))
-        assert_equal(
-            len(doc.xpath('//h1[contains(text(), "G-Cloud 8: proposed contract variation")]')), 1)
-
     def test_supplier_must_be_on_framework(self, data_api_client):
         supplier_not_on_framework = self.good_supplier_framework.copy()
         supplier_not_on_framework['frameworkInterest']['onFramework'] = False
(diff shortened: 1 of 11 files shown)


tests/app/main/test_frameworks.py
```diff
@@ -2316,3 +2316,198 @@ def test_framework_dashboard_shows_returned_agreement_details(
...
+        data_api_client.get_framework.return_value = self.g8_framework
+        data_api_client.get_supplier_framework_info.return_value = already_agreed
+
+        res = self.client.get("/suppliers/frameworks/g-cloud-8/contract-variation/1")
+        page_text = res.get_data(as_text=True)
+        doc = html.fromstring(page_text)
+
+        assert res.status_code == 200
+        assert len(doc.xpath('//h2[contains(text(), "Contract variation status")]')) == 1
+        assert "<span>William Drayton<br />agreed@email.com<br />Friday 19 August 2016 at 16:47</span>" in page_text
```
risicle: Yeeeeeea boiiiiii.

Ahem. Sorry.

tests/app/main/test_frameworks.py
```diff
@@ -2316,3 +2316,198 @@ def test_framework_dashboard_shows_returned_agreement_details(
             assert '<tdclass="summary-item-field"><span><p>signername</p><p>signerrole</p></span></td>' in page_without_whitespace  # noqa
             assert '<tdclass="summary-item-field"><span><p>User</p><p>email@email.com</p><p>Sunday10July2016at22:20</p></span></td>' in page_without_whitespace  # noqa
             assert '<tdclass="summary-item-field-first"><span>WaitingforCCStocountersign</span></td>' in page_without_whitespace  # noqa
+
+
+@mock.patch('app.main.views.frameworks.data_api_client', autospec=True)
+class TestContractVariation(BaseApplicationTest):
```
risicle: Is there a test for the case where the api returns a 400 for an already-signed variation?

tests/app/main/test_frameworks.py
TheDoubleK: :alarm_clock:

tests/app/main/test_frameworks.py
TheDoubleK: There isn't.  The app will error and 500 in that case, which I think is OK because the API call should never be made if the variation is already agreed - it would be a genuine application error if it did.

risicle: What about over-enthusiastic double-clickers?

The bane of a web developer's existence.

TheDoubleK: I've tyried pretty hard and can't provoke an error in the frontend by double clicking.

laurenceberry:
We should drop the current page from the breadcrumbs for now. Other than that looks good to me.

Wynndow:
I can't see anything wrong! 👍 : when tests pass!

PR #523, opened 2016-08-22
Redirect to contract variation after returning framework agreement

Wynndow:
For [this](https://www.pivotaltracker.com/story/show/127807417) story on Pivotal.
Will need rebasing on top of [this](https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/522) once it's merged before tests can go green.

If a user returns their framework agreement but has not agreed
to the framework variation then redirect them to the variation.

If a user returns their framework agreement and has agreed to the
framework variation then redirect them to the framework dashboard.

If there are no framework agreement variations for a framework then
the user should redirect to the framework dashboard.

Functionality is built under the assumption that a framework will
have no more than one variation. Removing this assumption would
add a reasonable amount of complexity, such as which variation
should be redirected too. At the moment, we have no foresight that
a framework having more than one variation is coming any time soon
so should not be worth investing in.

### DIFF
--- a/tests/app/main/test_frameworks.py
+++ b/tests/app/main/test_frameworks.py
@@ -2138,7 +2138,7 @@ class TestReturnSignedAgreement(BaseApplicationTest):
 
     @mock.patch('dmutils.s3.S3')
     @mock.patch('app.main.views.frameworks.send_email')
-    def test_valid_framework_agreement_returned_sends_confirmation_emails_and_redirects_to_framework_dashboard(
+    def test_valid_framework_agreement_returned_sends_confirmation_emails_and_unsets_session(
         self, send_email, s3, return_supplier_framework, data_api_client
     ):
         with self.app.test_client():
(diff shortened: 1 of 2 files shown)


tests/app/main/test_frameworks.py
```diff
@@ -2280,6 +2274,150 @@ def test_email_not_sent_if_api_call_fails(
             assert not send_email.called
 
     @mock.patch('dmutils.s3.S3')
+    @mock.patch('app.main.views.frameworks.send_email')
+    def test_framework_agreement_returned_having_signed_contract_variation_redirects_to_framework_dashboard(
```
TheDoubleK: Technically this can't happen - suppliers can only agree to a variation *after* returning their framework agreement (the contract variation page isn't shown if they haven't returned framework agreement yet).

But I don't think it hurts to keep this test as a weird just-in-case?

TheDoubleK:
This looks good to me.  Let's wait until the page PR is merged and get it rebased before a thumbs-up.

TheDoubleK:
:+1:

PR #524, opened 2016-08-22
Add link to contract variation page to framework dashboard

TheDoubleK:
If a supplier has returned their framework agreement then their documents page will show a link to any proposed contract variations.

Words agreed with Cath and Laurence.

Looks like this:
screen shot 2016-08-22 at 15 11 48


allait:
👍

PR #526, opened 2016-08-23
Enable Travis cache for pip, npm and bower

allait:
We've seen travis builds fail sometimes during npm install, so maybe [caching the dependency paths][1] will help to avoid this.

This might introduce other issues though, with stale cache causing the build to run outdated code, so keep in mind you might need to clear the travis cache if the build fails unexpectedly.

[1]: https://docs.travis-ci.com/user/caching

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,9 @@
 language: python
+cache:
+  pip: true
+  directories:
+    - node_modules
+    - bower_components
 python:
   - "2.7"
   - "3.4"


pcraig3:
Seems to contradict the [best practices recommended by MBS](http://www.brunton-spall.co.uk/post/2016/08/24/build-pipelines/) (Principle 3 in particular), but it totally seems better than failing unexpectedly with some regularity.

pcraig3:
👍

PR #527, opened 2016-08-31
[#129454097]Brief eligibility error messages fix

risicle:
Story https://www.pivotaltracker.com/story/show/129454097

We weren't filtering a supplier's services by those `published` when determining the reason for their non-eligibility, hence might make false deductions. So instead now filter all these queries for `published` services.

First created regression tests that fail with existing code, then added fix.

Also @idavidmcdonald I merged most of your separated `data-reason`-checking tests into their corresponding standard tests - if you can get away with testing for something by adding an assert to an existing test, there's no reason not to do it that way. Here I needed to create two new variants of the standard tests - if the asserts for the `data-reason` are included, when I copy the tests to create the new variants I get the `data-reason`-checking "for free", as opposed to having to remember to create _yet another_ two tests to cover it properly...

### DIFF
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -203,25 +203,25 @@ def _render_not_eligible_for_brief_error_page(brief, clarification_question=Fals
     common_kwargs = {
         "supplier_id": current_user.supplier_id,
         "framework": brief['frameworkSlug'],
+        "status": "published",
     }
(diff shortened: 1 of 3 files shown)


idavidmcdonald:
Good work 👍

risicle:
Ooh do I get a gold star?

risicle:
:star:

PR #528, opened 2016-09-05
[#119736241] Update utils to include 5xx ERROR logging

allait:
Makes the app log 5xx responses with ERROR level, which is useful for frontend apps since their responses can fail when the API responds with 5xx and no exceptions get logged.

This is the only frontend app that doesn't have the up-to-date utils version yet.


TheDoubleK:
:+1:

PR #529, opened 2016-09-14
[#130297677] Unicode bugfix plus supplier editing pre-live

TheDoubleK:
Two for the price of one...
This bugfix: https://www.pivotaltracker.com/story/show/130297677

Plus turning on existing supplier service editing in Pre-Live environments, so we can see what still needs building.

### DIFF
--- a/app/templates/frameworks/contract_variation.html
+++ b/app/templates/frameworks/contract_variation.html
@@ -143,7 +143,7 @@
             {{ summary.heading("Contract variation status", id="contract_variation_status") }}
             {% call(item) summary.list_table(
               [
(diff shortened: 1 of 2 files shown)


idavidmcdonald:
Do we want to add/modify a test for the unicode behaviour?

TheDoubleK:
In an ideal world, yes.

TheDoubleK:
Curses - looks like I've flushed out a whole bunch of potential unicode errors I need to fix now.

idavidmcdonald:
Spoke to Kev. Think we need this to go now so 👍. He's going to fix the other unicode errors he found separately.

PR #530, opened 2016-09-14
Adjust tests to run with non-ascii chars in supplier users

TheDoubleK:
We recently ran into a production bug caused by non-ascii characters in a supplier user's name.

This adjusts the tests to login as users with non-ascii characters in their name and supplier name, so hopefully there will be less chance of a similar bug making it through to production in future.


idavidmcdonald:
👍

PR #532, opened 2016-09-27
Save agreement path on upload

TheDoubleK:
This ensures that the supplier frontend saves the uploaded agreement file path into our database on upload.

These paths are not yet used by the frontend for rendering things, but they will be soon.


idavidmcdonald:
👍

PR #534, opened 2016-10-04
Remove "proposed" from page title if countersigned by CCS

TheDoubleK:
Once the contract variation is agreed we don't want the title to say "proposed".

### DIFF
--- a/app/templates/frameworks/contract_variation.html
+++ b/app/templates/frameworks/contract_variation.html
@@ -52,7 +52,7 @@
     <div class="grid-row">
         <div class="column-two-thirds">
             {% with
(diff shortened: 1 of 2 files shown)


app/templates/frameworks/contract_variation.html
```diff
@@ -52,7 +52,7 @@
     <div class="grid-row">
         <div class="column-two-thirds">
             {% with
-              heading = framework.name + ": proposed contract variation",
+              heading = "{}:{}contract variation".format(framework.name, " " if variation_details.get('countersignedAt') else " proposed "),
```
TheDoubleK: I tried various combinations but liked this the best in the end.

PR #535, opened 2016-10-04
Use file paths from FrameworkAgreements to find agreement files

TheDoubleK:
This updates the app to use the new FrameworkAgreement data to ascertain whether an agreement has been returned and/or countersigned, rather than checking for the existence of a file in S3.

It also saves uploaded agreements with a unique timestamped filename, which gets saved into the FrameworkAgreement table.  This will allow us to support multiple FrameworkAgreement records and thus fix the bug where a user could update a new file but not hit the green button, making the file out of step with our data.

We also set a new `download_filename` for signature pages - indicating that it's a signature page, not the whole signed agreement (a download filename was set for pre-G8 frameworks, but was missing from the new G8 flow).

The logic for showing things is as follows:
 * If `agreementReturnedAt` is set then we know the signed agreement (or signature page)  has been returned, and we can use the `agreementPath` to create signed links to download it.
 * If `countersignedPath` is set then we know the countersigned agreement has been returned and uploaded to S3 so suppliers can download it. (We don't look at `countersignedAt` because that only indicates that admin has approved it, not that the document exists).
 * During the G-8 signing flow the `agreementPath` is set first, when the file is uploaded, but no `agreementReturnedAt` is set until the green button is pressed.  This allows us to show the uploaded-at timestamp of the document prior to the green button being pressed, by getting metadata about the file from the S3 key.

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -15,8 +15,8 @@ from dmutils.formats import datetimeformat
 from dmutils import s3
 from dmutils.documents import (
     RESULT_LETTER_FILENAME, AGREEMENT_FILENAME, SIGNED_AGREEMENT_PREFIX,
-    SIGNATURE_PAGE_FILENAME, get_agreement_document_path, get_signed_url, get_extension, file_is_less_than_5mb,
-    file_is_empty, file_is_image, file_is_pdf, sanitise_supplier_name
+    SIGNATURE_PAGE_FILENAME, get_document_path, generate_timestamped_document_upload_path, get_signed_url,
+    get_extension, file_is_less_than_5mb, file_is_empty, file_is_image, file_is_pdf, sanitise_supplier_name
 )
 
 from ... import data_api_client, flask_featureflags
@@ -331,7 +331,7 @@ def download_agreement_file(framework_slug, document_name):
     elif document_name == 'countersigned-agreement':
         path = supplier_framework_info['countersignedPath']
     else:
-        path = get_agreement_document_path(framework_slug, current_user.supplier_id, document_name)
+        path = get_document_path(framework_slug, current_user.supplier_id, 'agreements', document_name)
 
     if not path:
         abort(404)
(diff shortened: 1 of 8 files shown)


TheDoubleK:
Note:  We'll need to get the corresponding Admin app changes (forthcoming) in before this, otherwise it won't be able to find any newly uploaded framework agreements.

app/main/views/frameworks.py
```diff
@@ -332,7 +324,17 @@ def download_agreement_file(framework_slug, document_name):
         abort(404)
 
     agreements_bucket = s3.S3(current_app.config['DM_AGREEMENTS_BUCKET'])
-    path = get_agreement_document_path(framework_slug, current_user.supplier_id, document_name)
+
+    if document_name == 'signed-agreement':
```
TheDoubleK: The point of this is that we specifically do *not* want to use `get_document_path` for signed and countersigned documents, because the full path is now stored in the database and we want to use that. But we still need the old behaviour for result letters and signed agreement files.

~~So I guess the alternative to doing it this way, with specific keys for signed and countersigned, would be to make separate routes for signed and countersigned that look at supplier framework...~~

Scratch that, yes maybe a better way would be to have the key from the `supplier_framework` passed as the `document_name` parameter, look in the supplier framework for the key, and if it isn't found fall back to `get_document_path`.

This would introduce some camelCase to our document URLs though. Can we live with that?

TheDoubleK: The point of this is that we specifically do *not* want to use `get_document_path` for signed and countersigned documents, because the full path is now stored in the database and we want to use that. But we still need the old behaviour for result letters and signed agreement files.

~~So I guess the alternative to doing it this way, with specific keys for signed and countersigned, would be to make separate routes for signed and countersigned that look at supplier framework...~~

Scratch that, yes maybe a better way would be to have the key from the `supplier_framework` passed as the `document_name` parameter, look in the supplier framework for the key, and if it isn't found fall back to `get_document_path`.

This would introduce some camelCase to our document URLs though. Can we live with that?

TheDoubleK: Sorry for duplication - obviously I meant that we need to retain the old behaviour for *unsigned* agreement downloads.

app/main/views/frameworks.py
```diff
@@ -332,7 +324,17 @@ def download_agreement_file(framework_slug, document_name):
...
+        path = supplier_framework_info['agreementPath']
+    elif document_name == 'countersigned-agreement':
+        path = supplier_framework_info['countersignedPath']
+    else:
+        path = get_document_path(framework_slug, current_user.supplier_id, 'agreements', document_name)
+
+    if not path:
```
TheDoubleK: Yeah this was belt-and-braces, I think we can probably do without this check.

app/main/views/frameworks.py
idavidmcdonald: Have implemented your suggestion @allait.

allait:
👍

PR #539, opened 2016-10-10
[#131615547]Service sumbission "continue" flow

risicle:
Did a little test tidying at the same time.


risicle:
Ok so I've pushed some more changes, the topmost one refactors/merges `edit_service_submission` and `update_section_submission` into a single view, but only after covering the `GET` path with some new tests so I could make sure they still pass (I'm still not particularly happy with how much these tests actually assert, but they're better than the `0` tests there were before)

TheDoubleK:
Tested locally for both DOS and G-Cloud flows and all seems well, code looks good.
:+1:

PR #542, opened 2016-10-25
[#130358101] Update FrameworkAgreements directly 

TheDoubleK:
For this story: https://www.pivotaltracker.com/story/show/130358101

** UPDATED WITH SMALLER SPACE ABOVE BUTTON **
Now depends on this being merged before tests will pass: 
 - [x] https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/282

These changes make it so that FrameworkAgreements are updated directly rather than via a SupplierFramework.

For pre-G-Cloud-8 frameworks this means making three API calls in a single route (i.e. to  create, add signer details and "sign" the agreement).

For G-Cloud-8 and beyond the new signing agreement flow now creates a FrameworkAgreement when the new "Return your signed signature page" button is first clicked (it's currently a link not a button):

screen shot 2016-10-26 at 16 58 40

and the ID of the FrameworkAgreement thus created becomes part of the  URL for subsequent pages, allowing the FrameworkAgreement to be updated  directly with first the signer details, then the agreement path (when the  file is uploaded) and finally the "signed_at" timestamp is added when the  final "confirm" button is clicked.

This is the penultimate piece of the puzzle to fix the bug that it was previously  possible for someone to come along and update the signer details for a signed  agreement after it had been countersigned, and without uploading a new document or hitting the "sign" button.

The risk is mitigated by this new flow (a new draft FrameworkAgreement is created  every time the green button is pressed, and existing ones not updated), but  is still susceptible to users navigating directly to the URL for a specific FrameworkAgreement and updating it.  The final piece of the puzzle to block this altogether will come when this API `TODO` is done (coming v. soon!): https://github.com/alphagov/digitalmarketplace-api/blob/master/app/main/views/agreements.py#L94

Note, any users in the middle of the "return a framework agreement" flow when this is deployed will hit a 404 error page and have to start again.  We discussed with @allait and decided that this is fine as it is highly unlikely anyone will be doing that (as nearly all DOS and G-8 agreements are returned already) and if they are then starting over is no great hardship.

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -680,10 +680,7 @@ def signature_upload(framework_slug, agreement_id):
     framework = get_framework(data_api_client, framework_slug, allowed_statuses=['standstill', 'live'])
     supplier_framework = return_supplier_framework_info_if_on_framework_or_abort(data_api_client, framework_slug)
     agreement = data_api_client.get_framework_agreement(agreement_id)['agreement']
-
-    # Need to write a test for this and move this functionality into a helper?
-    if agreement['frameworkSlug'] != framework_slug or agreement['supplierId'] != supplier_framework['supplierId']:
-        abort(404)
+    check_agreement_is_related_to_supplier_framework_or_abort(agreement, supplier_framework)
 
     agreements_bucket = s3.S3(current_app.config['DM_AGREEMENTS_BUCKET'])
     signature_page = agreements_bucket.get_key(agreement.get('signedAgreementPath'))
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -566,19 +566,15 @@ def upload_framework_agreement(framework_slug):
         )
     )
 
-    updated_supplier_framework = data_api_client.register_framework_agreement_returned(
-        current_user.supplier_id, framework_slug, current_user.email_address)
-    agreement_id = updated_supplier_framework.get("frameworkInterest", {}).get("agreementId")
-    try:
-        # If setting the agreement path fails then rollback the retuned agreement and raise error
-        data_api_client.update_framework_agreement(
-            agreement_id, {"signedAgreementPath": path}, current_user.email_address
-        )
-    except Exception as e:
-        data_api_client.unset_framework_agreement_returned(
-            current_user.supplier_id, framework_slug, current_user.email_address
-        )
-        raise e
+    agreement_id = data_api_client.create_framework_agreement(
+        current_user.supplier_id, framework_slug, current_user.email_address
+    )['agreement']['id']
+    data_api_client.update_framework_agreement(
+        agreement_id, {"signedAgreementPath": path}, current_user.email_address
+    )
+    data_api_client.sign_framework_agreement(
+        agreement_id, current_user.email_address, {"uploaderUserId": current_user.id}
+    )
 
     try:
         email_body = render_template(
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -605,6 +605,19 @@ def upload_framework_agreement(framework_slug):
     return redirect(url_for('.framework_agreement', framework_slug=framework_slug))
 
 
+@main.route('/frameworks/<framework_slug>/create-agreement', methods=['POST'])
+@login_required
+def create_framework_agreement(framework_slug):
+    framework = get_framework(data_api_client, framework_slug)
+    return_supplier_framework_info_if_on_framework_or_abort(data_api_client, framework_slug)
+
+    agreement_id = data_api_client.create_framework_agreement(
+        current_user.supplier_id, framework_slug, current_user.email_address
+    )["agreement"]["id"]
+
+    return redirect(url_for('.signer_details', framework_slug=framework_slug, agreement_id=agreement_id))
+
+
 @main.route('/frameworks/<framework_slug>/signer-details', methods=['GET', 'POST'])
 @login_required
 def signer_details(framework_slug):
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -740,25 +740,28 @@ def signature_upload(framework_slug, agreement_id):
 def contract_review(framework_slug, agreement_id):
     framework = get_framework(data_api_client, framework_slug, allowed_statuses=['standstill', 'live'])
     supplier_framework = return_supplier_framework_info_if_on_framework_or_abort(data_api_client, framework_slug)
-    agreements_bucket = s3.S3(current_app.config['DM_AGREEMENTS_BUCKET'])
-    signature_page = agreements_bucket.get_key(supplier_framework['agreementPath'])
+    agreement = data_api_client.get_framework_agreement(agreement_id)['agreement']
+    check_agreement_is_related_to_supplier_framework_or_abort(agreement, supplier_framework)
 
-    # if supplier_framework doesn't have a name or a role or the agreement file, then 404
+    # if framework agreement doesn't have a name or a role or the agreement file, then 404
     if not (
-        supplier_framework['agreementDetails'] and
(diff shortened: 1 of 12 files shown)


idavidmcdonald:
Top margin on that "return your signed signature page" button?

TheDoubleK:
I pinged @laurenceberry but haven't heard back yet.  Did you have any specific guidance on how it should look?

idavidmcdonald:
Sorry, I believe it was that it should not have so much top margin (however definitely worth checking directly). I think there is some new stuff that just got released in the toolkit for button links having advice above them that could come in use here if so.

Wynndow:
Yeah, I just added the [link button with advice](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/link-button.html)

tests/app/main/test_frameworks.py
```diff
@@ -1778,29 +1937,117 @@ def test_dos_drafts_list_with_closed_framework(self, count_unanswered, data_api_
...
+                on_framework=False)
+
+            res = self.client.post("/suppliers/frameworks/g-cloud-8/create-agreement")
+            assert res.status_code == 404
+
+    def test_404_if_framework_in_wrong_state(self, data_api_client):
+        with self.app.test_client():
+            self.login()
+            # Suppliers can only sign agreements in 'standstill' and 'live' lifecycle statuses
+            data_api_client.get_framework.return_value = self.framework(status='pending')
```
Wynndow: Is it worth testing other framework statuses that shouldn't be valid?

tests/app/main/test_frameworks.py
```diff
@@ -1778,29 +1937,117 @@ def test_dos_drafts_list_with_closed_framework(self, count_unanswered, data_api_
...
+        self, check_agreement_is_related_to_supplier_framework_or_abort, return_supplier_framework, data_api_client
+    ):
+        self.login()
+        data_api_client.get_framework.return_value = get_g_cloud_8()
+        data_api_client.get_framework_agreement.return_value = self.framework_agreement(supplier_id=2345)
+        supplier_framework = self.supplier_framework(framework_slug='g-cloud-8', on_framework=True)['frameworkInterest']
+        return_supplier_framework.return_value = supplier_framework
+
+        res = self.client.get("/suppliers/frameworks/g-cloud-8/234/signer-details")
+        check_agreement_is_related_to_supplier_framework_or_abort.assert_called_with(
```
Wynndow: Does this actually test that it's aborted? It looks like it just checks if the method was called.

tests/app/main/test_frameworks.py
```diff
@@ -1886,42 +2138,167 @@ def test_provide_signer_details_form_with_valid_input_redirects_to_upload_page(
...
+    ):
+        self.login()
+        data_api_client.get_framework.return_value = get_g_cloud_8()
+        data_api_client.get_framework_agreement.return_value = self.framework_agreement(supplier_id=2345)
+        supplier_framework = self.supplier_framework(framework_slug='g-cloud-8', on_framework=True)['frameworkInterest']
+        return_supplier_framework.return_value = supplier_framework
+        s3.return_value.get_key.return_value = None
+
+        res = self.client.get("/suppliers/frameworks/g-cloud-8/234/signature-upload")
+        check_agreement_is_related_to_supplier_framework_or_abort.assert_called_with(
```
Wynndow: Again, is this checking that we abort?

tests/app/main/test_frameworks.py
```diff
@@ -2188,24 +2557,45 @@ def test_contract_review_page_aborts_if_visited_when_information_required_to_ret
...
+    ):
+        self.login()
+        data_api_client.get_framework.return_value = get_g_cloud_8()
+        data_api_client.get_framework_agreement.return_value = self.framework_agreement(supplier_id=2345)
+        supplier_framework = self.supplier_framework(framework_slug='g-cloud-8', on_framework=True)['frameworkInterest']
+        return_supplier_framework.return_value = supplier_framework
+        s3.return_value.get_key.return_value = None
+
+        res = self.client.get("/suppliers/frameworks/g-cloud-8/234/contract-review")
+        check_agreement_is_related_to_supplier_framework_or_abort.assert_called_with(
```
Wynndow: Same again, does this check it actually aborts.

TheDoubleK: Good spot Chris!  I'll add some extra assertions in the morning.

TheDoubleK: Actually Chris, looking at these not on my phone and with more context than 2 lines of code these tests mock out the `check_agreement_is_related_to_supplier_framework_or_abort` in order to check that the request fails for the "right" reason - i.e. the one specified in the test - rather than some other reason.  I'll add some comments to the tests to make it clear that this is what's happening.

The rationale is that the abort method is separately unit tested to be doing the right thing, so asserting that it was called with arguments that will provoke an abort is sufficient for these tests.

TheDoubleK: There isn't any point in making the mock actually abort and checking the response is 404 because that's just testing that the mocking framework does what it's supposed to, and wouldn't test any of our code.

risicle:
I'm beginning to read this now

risicle:
Sorry something I don't understand is why it has to work with pre g8 contracts at all anymore...?

TheDoubleK:
Because there are still DOS and G7 suppliers who haven't returned their agreements and there is a slim chance that they might do in future.  For those frameworks we don't collect any signer details or anything like that.  We probably *could* push everyone through the new flow and just ignore the extra data they have to enter and call it some sort of penalty for being so late.  But a decision was taken at some point that the flow should remain the same as it always was for older frameworks.

app/main/views/frameworks.py
```diff
@@ -610,11 +607,26 @@ def upload_framework_agreement(framework_slug):
     return redirect(url_for('.framework_agreement', framework_slug=framework_slug))
 
 
-@main.route('/frameworks/<framework_slug>/signer-details', methods=['GET', 'POST'])
+@main.route('/frameworks/<framework_slug>/create-agreement', methods=['POST'])
 @login_required
-def signer_details(framework_slug):
-    framework = get_framework(data_api_client, framework_slug)
+def create_framework_agreement(framework_slug):
+    framework = get_framework(data_api_client, framework_slug, allowed_statuses=['standstill', 'live'])
```
risicle: Also there's this odd situation where the logic that the behaviour is switched by `frameworkAgreementVersion` is spread across both the frontend and the api. Let's hope they always agree.

risicle:
More broadly (probably not for this PR) I feel at some point we should be moving a lot of these views to `agreements.py`. There seems to have been an attempt at some point to base the file organization on url paths, but by now that seems to be screwed up six ways from sunday, so I don't know.

app/main/views/frameworks.py
```diff
@@ -700,56 +719,49 @@ def signature_upload(framework_slug):
...
+    # if framework agreement doesn't have a name or a role or the agreement file, then 404
     if not (
-        supplier_framework['agreementDetails'] and
-        supplier_framework['agreementDetails'].get('signerName') and
-        supplier_framework['agreementDetails'].get('signerRole') and
-        supplier_framework['agreementPath']
+        agreement.get('signedAgreementDetails') and
+        agreement['signedAgreementDetails'].get('signerName') and
+        agreement['signedAgreementDetails'].get('signerRole') and
+        agreement.get('signedAgreementPath')
```
TheDoubleK: This doesn't work if the `signedAgreementDetails` key exists with a `None` value though, (e.g. `{'signedAgreementDetails': None, 'signedAgreementPath': None, 'frameworkSlug': 'g-cloud-8', 'id': 234, 'supplierId': 2345}`) so I prefer the version as it.

TheDoubleK:
Concurrent user issue discussed in real life; I've aded the `frameworkAgreementVersion` checks and some trailing commas.

Good to go now?

PR #543, opened 2016-10-25
[#131156187] Update DOS2 welcome email content

TheDoubleK:
For this story: https://www.pivotaltracker.com/story/show/131156187

Updated from the document referenced in the story.
Email looks like this when delivered (gmail):
screen shot 2016-10-25 at 16 06 49

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -51,7 +51,7 @@ def framework_dashboard(framework_slug):
                 [user['emailAddress'] for user in supplier_users['users'] if user['active']],
                 email_body,
                 current_app.config['DM_MANDRILL_API_KEY'],
-                'You have started your {} application'.format(framework['name']),
+                'You started a {} application'.format(framework['name']),
                 current_app.config['CLARIFICATION_EMAIL_FROM'],
                 current_app.config['CLARIFICATION_EMAIL_NAME'],
                 ['{}-application-started'.format(framework_slug)]
--- a/app/templates/emails/digital-outcomes-and-specialists-2_application_started.html
+++ b/app/templates/emails/digital-outcomes-and-specialists-2_application_started.html
@@ -6,12 +6,30 @@
 <body>
     Dear supplier,
     <br /><br />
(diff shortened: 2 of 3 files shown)


app/main/views/frameworks.py
```diff
@@ -51,7 +51,7 @@ def framework_dashboard(framework_slug):
                 [user['emailAddress'] for user in supplier_users['users'] if user['active']],
                 email_body,
                 current_app.config['DM_MANDRILL_API_KEY'],
-                'You have started your {} application'.format(framework['name']),
+                'You started a {} application'.format(framework['name']),
```
risicle: Only problem with this language is it will require a "a"/"an" switch if we come up with a framework beginning with a vowel.

app/templates/emails/digital-outcomes-and-specialists-2_application_started.html
```diff
@@ -6,12 +6,30 @@
...
     Dear supplier,
     <br /><br />
-    You, or one of your colleagues, has started an application for the Digital Outcomes and Specialists 2 framework. All contributors on your Digital Marketplace account will now receive email updates.
+    You, or one of your team, started an application to become a Digital Outcomes and Specialists 2 supplier.
+    All contributors on your Digital Marketplace account will now receive email updates.
+    The deadline for completing your application is 6 December 2016.
     <br /><br />
-    You need to confirm that you’re eligible to offer services under Digital Outcomes and Specialists 2 as part of the application. You don’t need to describe your services now. You’ll be asked to provide more details when you respond to a buyer’s specific brief.
+    You must ask any clarification questions you have through your Digital Marketplace account before
+    5pm GMT, 22 November 2016.
```
risicle: Ultimately these dates should probably be pulled from the content loader I guess.

app/main/views/frameworks.py
TheDoubleK: Good point.  I'm not sure what the motivation for changing it was... I'll check, would be nice not to set ourselves that trap for the future, but also nice to not build in logic we don't need right now.

PR #545, opened 2016-10-27
Brief responses call submit route on api

idavidmcdonald:
Will set the brief response to be submitted for our existing one page flow. This will allow us to then do a data migration and set all brief responses as submitted.

Send page questions when creating a brief response. This is a preemptive change as the API will shortly be validating content when creating a brief response based on page questions rather than all questions in the schema as it does at the moment. This is similar to when creating a brief.

Tests will fail until [API client pull request](https://github.com/alphagov/digitalmarketplace-apiclient/pull/55) is merged.

### DIFF
--- a/tests/app/main/test_briefs.py
+++ b/tests/app/main/test_briefs.py
@@ -575,11 +575,18 @@ class TestRespondToBrief(BaseApplicationTest):
         assert res.status_code == 302
         assert res.location == "http://localhost/suppliers/opportunities/1234/responses/result?result=success"
         data_api_client.create_brief_response.assert_called_once_with(
-            1234, 1234, processed_brief_submission, 'email@email.com')
+            1234,
+            1234,
+            processed_brief_submission,
+            'email@email.com',
+            page_questions=[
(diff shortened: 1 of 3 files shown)


tests/app/main/test_briefs.py
```diff
@@ -575,13 +575,22 @@ def test_create_new_brief_response(self, data_api_client):
...
+            processed_brief_submission,
+            'email@email.com',
+            page_questions=[
+                'respondToEmailAddress', 'essentialRequirements', 'niceToHaveRequirements', 'availability', 'dayRate'
+            ]
+        )
+        data_api_client.submit_brief_response.assert_called_once_with(
+            6, 'email@email.com')
 
     def test_create_new_brief_response_shows_result_page_for_not_all_essentials(self, data_api_client):
```
tombye: This results in a redirect so it's a little confusing the test name indicates it should result in a page type.

idavidmcdonald: Have changed.

idavidmcdonald:
👍 self thumbs up as tom is in an interview and has said happy when his comment has been addressed.

PR #546, opened 2016-10-27
Always call .filter before displaying section data

allait:
Since section fields can go through additional processing with Jinja we need to ensure we're always calling .filter before displaying them to the user so that we don't accidentally display template source.

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -261,7 +261,7 @@ def framework_submission_services(framework_slug, lot_slug):
 def framework_supplier_declaration(framework_slug, section_id=None):
     framework = get_framework(data_api_client, framework_slug, allowed_statuses=['open'])
 
-    content = content_loader.get_manifest(framework_slug, 'declaration')
+    content = content_loader.get_manifest(framework_slug, 'declaration').filter({})
     status_code = 200
 
     if section_id is None:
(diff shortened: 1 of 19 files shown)


app/main/views/frameworks.py
```diff
@@ -911,14 +911,19 @@ def view_contract_variation(framework_slug, variation_slug):
                 {'question': form['accept_changes'].label.text, 'input_name': 'accept_changes'}
             ]
 
+    supplier_name = supplier_framework['declaration']['nameOfOrganisation']
+    variation_content = content_loader.get_message(framework_slug, variation_content_name).filter(
+        {'supplier_name': supplier_name}
+    )
```
pcraig3: While quite a lot better than the old way of finding the supplier name, this still seems a bit ad-hoc.  
How are we supposed to think about it? Do we just pass in whatever we need as needed?

tests/app/main/test_frameworks.py
```diff
@@ -3172,7 +3172,7 @@ def test_success_message_is_displayed_on_success(self, data_api_client):
         assert res.status_code == 200
         assert len(
             doc.xpath('//p[@class="banner-message"][contains(text(), "You have accepted the proposed changes.")]')
-        ) == 1
+        ) == 1, res.get_data(as_text=True)
```
allait: It's useful, otherwise that error message just says `0 != 1`

pcraig3:
If the [pull request in utils](https://github.com/alphagov/digitalmarketplace-utils/pull/286) is approved, you could potentially pull that in as well.

TheDoubleK:
If I rebase this now and add content 4.0.2 will it be good to go?

pcraig3:
Hey Kev!  Hold on a sec before merging this.

PR #549, opened 2016-11-07
[#131617119]framework_dashboard view: heavily refactor

risicle:
Story https://www.pivotaltracker.com/story/show/131617119

Superficially splitting "supplier packs" into individual files, however this required changing a lot organizationally and improving a lot of the tests.

Apologies this is one great big changeset, it wasn't clear how much I would have to change when I started. Depends on frameworks 3.5.0 which isn't merged yet.


risicle:
Will follow with some screenshots, but there are so many different states, I can't show all of them.

TheDoubleK:
I'm getting three test failures in `test_frameworks.py`, as is Travis.

TheDoubleK:
The page isn't working right for DOS (supplier on framework with a countersigned agreement) - DOS has no "framework_agreement_details" and so should be an "old-style" page with a link to "Download your countersigned framework agreement (.pdf)", not "Download your ‘counterpart’ framework agreement signature page":
screen shot 2016-11-07 at 14 13 12

risicle:
Will look at test failures - screenshots follow:

Framework @ `open`, unsigned agreement, open questions:
20161107_fw_db_unsigned_cqo

Framework @ `standstill`, unsigned agreement:
20161107_fw_db_standstill_unsigned

Framework @ `standstill`, signed agreement:
20161108_fw_db_standstill_signed

risicle:
I think the test failures will be things introduced by the updated frameworks finally getting pulled in... investigating.

risicle:
Ah yes it's because I'd had to remove its expectation for the `framework_agreement_url` while I was getting the frameworks repo in place.

risicle:
This does make me realize that none of my tests cover the `-final-agreement.pdf` functionality.

TheDoubleK:
Just looking at your "standstill-with signed agreement" screenshot, and also G8 live-with-signed-agreement locally (see below) - having the "Sign and return your framework agreement" link so prominent at the top of the page when they have already returned something feels wrong.

The prototype page for that state (http://dm-prototype.herokuapp.com/structure/documents/services/?framework_slug=dos2&framework_state=signed) doesn't have the link shown at all, but I think that is probably wrong too as suppliers definitely still need to be able to re-upload signed agreement files, e.g. if CCS ask them to.  But perhaps we could hide the link and if they need to re-upload an agreement then they have to navigate directly to the URL to re-upload?  One to ask @cathrooney about I think.

screen shot 2016-11-07 at 16 05 44

risicle:
That's interesting kev, was just talking to @cathrooney about a related situation, will throw this issue into the mix...

risicle:
Good catch @TheDoubleK - @cathrooney says the link shouldn't be shown at all in this case.

app/templates/frameworks/_agreement_returned_legal.html
```diff
@@ -0,0 +1,106 @@
...
+        </a>
+      </p>
+      {% endfor %}
+    {% endif %}
+    {% if communications_files.call_off.last_modified %}
+      <p>
+        <a href="{{ url_for('.download_supplier_file', filepath=communications_files.call_off.filename, framework_slug=framework.slug) }}">Download the call-off contract template</a>
+      </p>
+    {% endif %}
+    {% if not supplier_framework.countersigned %}
```
TheDoubleK: This doesn't seem like the right condition for this piece of text?  One to check with @cathrooney but I think we can say "you can start selling your services" once they have returned a signed agreement (i.e. `if supplier_framework.agreementReturned`), whether or not it has been countersigned is irrelevant.

app/templates/frameworks/_guidance_links.html
```diff
@@ -1,87 +1,104 @@
...
+        {% if supplier_framework.agreementReturned %}
+          sell your services
+        {% else %}
+          apply
         {% endif %}
-    </li>
+      </a>
+    </p>
+  </div>
+{% if framework.status not in ['standstill', 'live'] and not supplier_framework.agreementReturned %}
```
TheDoubleK: I don't think we need the `and` part of this clause because `supplier_framework.agreementReturned` can only be set while the framework is `in ['standstill', 'live']`, so if framework isn't standstill or live then there's no way the agreement can be returned.

app/templates/frameworks/_guidance_links.html
```diff
@@ -1,87 +1,104 @@
...
+      </span>
+    </p>
+    {% endif %}
+  </div>
+{% endif %}
+  <div>
+    <h2 class="heading-xmedium" class="heading-xmedium">Communications</h2>
 
-    <li>
+    {% if framework.status not in ['standstill', 'live'] %}
```
TheDoubleK: This feels like it should be shown for `framework.status == 'open'` only.

app/templates/frameworks/_guidance_links.html
```diff
@@ -1,87 +1,104 @@
...
       <a href="{{ url_for('.framework_updates', framework_slug=framework.slug) }}">
-        <span>
-          {% if framework.clarificationQuestionsOpen %}
-            Read updates and ask clarification questions
-          {% else %}
-            Read updates and clarification question responses
-          {% endif %}
-        </span>
+        View communications and
+        {% if framework.status not in ['standstill', 'live'] %}ask{% endif %}
```
TheDoubleK: Again, you can only ask questions while it's `open`.  This would show "ask" during `pending` too, which we don't want.

app/templates/frameworks/dashboard.html
```diff
@@ -1,6 +1,16 @@
 {% extends "_base_page.html" %}
 
-{% block page_title %}Apply to {{ framework.name }} – Digital Marketplace{% endblock %}
+{% block page_title %}
```
TheDoubleK: Nice one - this definitely needed sorting out!

app/templates/frameworks/dashboard.html
```diff
@@ -23,38 +33,43 @@
 
   {% with messages = get_flashed_messages(with_categories=True) %}
     {% for category, message in messages %}
+    {% endfor %}
+  {% endwith %}
```
TheDoubleK: This `with` block doesn't do anything with the flashed messages - looks like a partial duplicate of the block below and this top one can be deleted.

app/templates/frameworks/_agreement_returned_legal.html
```diff
@@ -0,0 +1,106 @@
+{% if supplier_is_on_framework and framework.status in ['standstill', 'live'] and supplier_framework.agreementReturned %}
+  <div>
+    <h2 class="page-subheading">Legal documents</h2>
+    <div class="dmspeak">
+    {% if supplier_framework.countersigned %}
```
risicle: Balls. OK.

app/templates/frameworks/_agreement_returned_legal.html
```diff
@@ -0,0 +1,106 @@
...
+      <p>
+        Your framework agreement signature page has been sent to
+        the Crown Commercial Service (<abbr title="Crown Commercial Service">CCS</abbr>).
+        They’ll check and countersign the agreement, or contact you if there’s a
+        problem. You don’t need to wait for <abbr title="Crown Commercial Service">CCS</abbr> to
+        countersign your agreement before you start selling services.
+      <p/>
+    {% else %}
+      <p>
+        Your original and counterpart signature pages, and the standard framework
```
risicle: So just nothing in their place then?

app/templates/frameworks/_agreement_returned_legal.html
```diff
@@ -0,0 +1,106 @@
...
+        {{ summary.field_name('Submitted by') }}
+        {% call summary.field() %}
+          <p>{{ supplier_framework.agreementDetails.uploaderUserName }}</p>
+          <p>{{ supplier_framework.agreementDetails.uploaderUserEmail }}</p>
+          <p>{{ supplier_framework.agreementReturnedAt|datetimeformat }}</p>
+        {% endcall %}
+      {% endcall %}
+      {% call summary.row() %}
+        {{ summary.field_name('Countersignature') }}
+        {{ summary.field_name('Waiting for CCS to countersign') }}
```
risicle: Actually according to the prototype this information table should just be removed entirely when countersigned.

app/templates/frameworks/_agreement_returned_legal.html
risicle: I was just going on the fact that it's present in the pre-countersigned page and not in the countersigned page.

risicle: ...and @laurenceberry says this is the correct behaviour - they need a reassurance paragraph pre-countersigning but not otherwise.

app/templates/frameworks/_agreement_returned_legal.html
risicle: (@kevkeenoy says yes, the latter)

app/templates/frameworks/_guidance_links.html
risicle: Again.

risicle:
Thanks kev, this really needed an eye from someone with a better idea of the bigger picture - I'd been so close to it the various combinations were baffling me, but I think we've caught a good number of wrongnesses.

risicle:
Though I have noticed that none of my tests cover the `pending` framework status at all.

app/templates/frameworks/_agreement_returned_legal.html
```diff
@@ -0,0 +1,106 @@
...
+      </p>
+    {% endif %}
+      <p>
+        <a href="{{ url_for('.download_agreement_file', framework_slug=framework.slug, document_name=signed_agreement_document_name) }}" target="_blank">Download your ‘original’ framework agreement signature page</a>
+        <br/>
+        The agreement signed by your company
+      </p>
+    {% if countersigned_agreement_file %}
+      <p>
+        <a href="{{ url_for('.download_agreement_file', framework_slug=framework.slug, document_name=countersigned_agreement_file) }}" target="_blank">Download your ‘counterpart’ framework agreement signature page</a>
```
risicle: Maybe I should locate larry to give me some new copy here...

risicle: /me summons @andymcaleer instead

TheDoubleK:
The page for "old" frameworks when an agreement has ben returned but not countersigned looks a bit funky, but as we should have no actual users in that state maybe we can live with it I think.

screen shot 2016-11-09 at 11 06 38

TheDoubleK:
* (but we actually do have some I think)

TheDoubleK:
Hmm, the text is wrong for old frameworks with countersigned agreements too though - it still says "Download your ‘original’ framework agreement signature page", but should be "Download your countersigned framework agreement".

# OLD
screen shot 2016-11-09 at 11 22 26

# CURRENT PR
screen shot 2016-11-09 at 11 21 40

risicle:
:\

risicle:
Oh do 'old' frameworks not have an "original" (non countersigned) agreement?

TheDoubleK:
Actually the old version didn't have an option to download the original uploaded file...

app/main/views/frameworks.py
```diff
@@ -74,52 +74,75 @@ def framework_dashboard(framework_slug):
...
+            "path": "communications/",
+            "filename": "{}-final-agreement.pdf".format(framework_slug),
+        },
+        "call_off": {
+            "path": "communications/",
+            "filename": "{}-call-off.pdf".format(framework_slug),
+        },
+        "reporting_template": {
+            "path": "communications/",
+            "filename": "{}-reporting-template.xlsx".format(framework_slug),
```
TheDoubleK: I think this is going to be `.xls` rather than `xlsx`...

risicle: mmmok here's hoping office suites are clever enough to auto-detect whichever filetype it is.

TheDoubleK:
:+1:

PR #551, opened 2016-11-08
Replace in-app beta banner with new toolkit version

pcraig3:
Replaces the in-app beta banner with the new toolkit version.  
Means we can delete a couple of files.

Largely the same as I'm doing in the [buyer frontend](https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/399).


allait:
👍

PR #552, opened 2016-11-09
#130018567 Start page for supplier applications

idavidmcdonald:
For [this pivotal story](https://www.pivotaltracker.com/story/show/130018567)

- Adds the start page for applying to a brief response. Content is specific to the lot.
- For the moment, redirects to a temporary holder route but will in future go to the new flow for applying to a brief

- If no brief response exists, then the button says 'Start application' and clicking it will create a new draft brief response
image

- If a draft brief response exists, then the button says 'Continue application'. When we create the new flow (as part of [this pivotal story](https://www.pivotaltracker.com/story/show/129843611)) this button will need to link to their existing brief response application.
image

- If a brief response already exists then they are redirected to their already submitted brief response.

Will need rebasing after the [content loader filtering pull request](https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/546) has been merged

### DIFF
--- a/tests/app/main/test_briefs.py
+++ b/tests/app/main/test_briefs.py
@@ -947,6 +947,16 @@ class TestStartBriefResponseApplication(BaseApplicationTest):
 
         assert "provide the specialist's day rate" not in data
 
+    def test_start_page_is_hidden_by_feature_flag(self, data_api_client):
+        self.app.config['FEATURE_FLAGS_NEW_SUPPLIER_FLOW'] = False
+        data_api_client.get_brief.return_value = api_stubs.brief(status='live', lot_slug='digital-outcomes')
+        data_api_client.find_brief_responses.return_value = {
+            'briefResponses': []
+        }
+        res = self.client.get('/suppliers/opportunities/1234/responses/start')
+        assert res.status_code == 404
+
+
 @mock.patch("app.main.views.briefs.data_api_client")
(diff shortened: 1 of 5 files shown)


tests/app/main/test_briefs.py
```diff
@@ -818,6 +818,190 @@ def test_create_new_brief_response_redirects_to_login_for_buyer(self, data_api_c
...
+            'briefResponses': []
+        }
+        res = self.client.get('/suppliers/opportunities/1234/responses/start')
+        assert res.status_code == 200
+
+        doc = html.fromstring(res.get_data(as_text=True))
+        assert doc.xpath('//h1')[0].text.strip() == "Apply for ‘I need a thing to do a thing’"
+        assert doc.xpath("//input[@class='button-save']/@value")[0] == 'Start application'
+
+    def test_start_page_is_viewable_and_has_start_button_if_no_existing_brief_response(
```
tombye: The assertion in this test has already been checked by the `test_start_application_contains_brief_title` test. Do we need this test or can we just rename `test_start_application_contains_brief_title` to indicate the assertions covered?

idavidmcdonald: Well spotted! I will remove the assertion for the button content from the first test as it doesn't belong in there and keep the second test (as this gives us two tests to mirror the two cases for when the button says 'start application' and where it says 'continue application')

PR #553, opened 2016-11-09
Fix HTML descriptions in supplier registration forms

allait:
Supplier registration questions are manually using toolkit form templates, so they were affected by the removal of safe filters from the toolkit.


TheDoubleK:
:+1: gonna merge this and get it on to Preview now.

PR #554, opened 2016-11-10
Update frontend toolkit version

TheDoubleK:
Preview seems to have old content loaded even though version is listed as 4.0.2.

Bumping toolkit version mainly as a means to forcing a reload on preview :/


TheDoubleK:
:+1: do it, Keenoy.

PR #555, opened 2016-11-11
Add safe tag to "ask a brief question" description

pcraig3:
This was rendering with the verbatim html tags ever since the `markdown` filter was removed from the toolkit testbox template.

Works now though.

***

### Fixes this: 

image


TheDoubleK:
:+1:

PR #556, opened 2016-11-11
Update guidance file links

TheDoubleK:
Addresses this bug: https://www.pivotaltracker.com/story/show/134161817

Call-off contracts have a "proposed" and "final" version, which we were previously missing.
Also, we shouldn't just have "agreement" files - they should always be "framework-agreement" for clarity.

**NOTE**: I have already created all the required files in S3 for all environments (i.e. duplicated the current files with new filenames), so links should work fine once this is released without any further S3 wrangling required (obviously it will need testing in Preview/Staging to be sure though).

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -108,15 +108,19 @@ def framework_dashboard(framework_slug):
         },
         "proposed_agreement": {
             "path": "communications/",
-            "filename": "{}-proposed-agreement.pdf".format(framework_slug),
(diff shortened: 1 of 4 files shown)


risicle:
:+1: Cool - I just chose the filenames off the top of my head as placeholders, surprised they got released.

TheDoubleK:
Yeah we should watch that... just you and me reviewing stuff for our mini-team's stories.

risicle:
Could be worse, at my last place I left in some placeholder text in an obscure tooltip that said "Trousers falling down". Was there for two years.

PR #557, opened 2016-11-14
Fix initial project setup

danielknell:
- cleared out unused code from bootstrap.sh
- added npm_install to Makefile in line with other projects
- added npm_install to run_all directive in Makefile
- made run_app use the virtualenv python
- used DM_ENVIRONMENT to infer a GULP_ENVIRONMENT in order to build development assets on development environments and production assets in all other environments

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ endif
 run_all: requirements frontend_build run_app
 
 run_app: show_environment virtualenv
-	python application.py runserver
+	${VIRTUALENV_ROOT}/bin/python application.py runserver
 
 virtualenv:
 	[ -z $$VIRTUAL_ENV ] && [ ! -d venv ] && virtualenv venv || true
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -33,4 +43,4 @@ show_environment:
 	@echo "Environment variables in use:"
 	@env | grep DM_ || true
 
-.PHONY: run_all run_app virtualenv requirements requirements_for_test frontend_build test test_pep8 test_python test_javascript show_environment
+.PHONY: run_all run_app virtualenv requirements requirements_for_test npm_build frontend_build test test_pep8 test_python test_javascript show_environment
```
idavidmcdonald: Sorry, my knowledge is not great about make files. Where is this `npm_build` defined? I see `npm_install`, although that is not listed in the `.PHONY`, how come?

idavidmcdonald: My guess is that `npm_build` should instead be `npm_install`?

danielknell: fixed

idavidmcdonald:
👍 Looks good when the tests are green.

PR #560, opened 2016-11-16
Update utils version

pcraig3:
This is technically a breaking change because it removes the `|markdown` filter from jinja, which means we can't use it in our templates in this app anymore.

Fortunately, we've removed all uses of `|markdown` in this app, so there aren't any changes needed.


TheDoubleK:
:+1: after a rebase.

PR #561, opened 2016-11-17
Always show legal docs

TheDoubleK:
Fixes this bug: https://www.pivotaltracker.com/story/show/134350957

(I also picked up another issue that the "result letter" link wasn't appearing for FAILed G-Cloud 8 suppliers.  All failed suppliers get a letter, even on the "new" style frameworks.)

### DIFF
--- a/app/templates/frameworks/_framework_actions.html
+++ b/app/templates/frameworks/_framework_actions.html
@@ -1,5 +1,5 @@
 {% if framework.status in ['standstill', 'live'] and application_made and not countersigned_agreement_file %}
-  {% if not framework.frameworkAgreementVersion %}
+  {% if not (framework.frameworkAgreementVersion and supplier_is_on_framework) %}
   <li class="browse-list-item">
     <a class="browse-list-item-link" href="{{ url_for('.download_agreement_file', framework_slug=framework.slug, document_name='result-letter.pdf') }}" download>
       <span>Download your application {% if supplier_is_on_framework %}award{% else %}result{% endif %} letter (.pdf)</span>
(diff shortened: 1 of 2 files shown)


app/templates/frameworks/_guidance_links.html
```diff
@@ -21,7 +21,13 @@ <h2 class="heading-xmedium">Guidance</h2>
       </a>
     </p>
   </div>
-{% if framework.status not in ['standstill', 'live'] and not supplier_framework.agreementReturned %}
+{#
+  For framework states prior to 'pending' 'supplier_is_on_framework' will be false for everyone so links will be shown.
+  Passes and fails are set some time during 'pending' but we want to continue to show the links to *all* suppliers
+  up until the start of standstill.
+  After that they are hidden from successful suppliers and replaced by '_agreement_returned_legal.html' content.
```
risicle: :cake: for this explanation

app/templates/frameworks/_framework_actions.html
```diff
@@ -1,5 +1,5 @@
 {% if framework.status in ['standstill', 'live'] and application_made and not countersigned_agreement_file %}
-  {% if not framework.frameworkAgreementVersion %}
+  {% if not framework.frameworkAgreementVersion or not supplier_is_on_framework %}
```
risicle: Um, couldn't we have `if not (framework.frameworkAgreementVersion and supplier_is_on_framework)` plz? Good old De Morgan, removes any order of precedence ambiguity.

TheDoubleK: Yep makes sense.  By the time I worked out what the condition should actually be my mind was pretty much melted so I just went with it.

TheDoubleK:
Updated the conditional.  I was also surprised at the test happiness.  This has only changed the guidance links for standstill period though, and there aren't many tests around standstill statuses.  I was being lazy and should probably add some.

PR #562, opened 2016-11-17
add tests for xsses on brief clarification question page

leohemsted:
* ensure question contents are escaped in the emails
* ensure brief name is escaped in the header


allait:
👍

PR #563, opened 2016-11-18
Supplier frontend on PaaS

allait:
### Add gunicorn to the requirements.txt

We're running gunicorn when deploying apps to PaaS, so it has to be listed in the requirements file.

### Add .cfignore file for PaaS
Lists folders ignored by `cf push`. Significantly speeds up PaaS deployments. Contents copied from the buyer frontend.

### DIFF
--- a/.cfignore
+++ b/.cfignore
@@ -1,0 +1,5 @@
+node_modules/
+bower_components/
+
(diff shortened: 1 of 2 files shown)


galund:
👍

PR #566, opened 2016-11-25
Apply for a brief multipage flow

idavidmcdonald:
We will be running a new multipage process for suppliers applying for a brief. This process will need to run in parallel to the current one for a period of two weeks. After this period we will be able to remove the current process (one page flow).

br_flow

### Prototype
https://dm-prototype.cloudapps.digital/structure/opportunity/dos-opportunities?opportunity=0

### Waiting on the following pull requests to be merged

- [x] https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/290

- [ ] https://github.com/alphagov/digitalmarketplace-frameworks/pull/334

### Pivotal stories
Covers the following stories:
- https://www.pivotaltracker.com/story/show/129843611
- https://www.pivotaltracker.com/story/show/129839965
- https://www.pivotaltracker.com/story/show/131617615

### Half way house
We are aiming to build the prototype but this pull request only has half of the work done. This is so we can get something small(er) code reviewed/QA/merged etc rather than a massive block of work. Note, because of this, the flow will not work and validation will fail. This is more to show the general premise and check the questions which have been done are appearing correctly and are correct for design and copy.

### Validation
We have not generated a new schema for the API for this new brief response flow (because we will have to maintain two different schemas at the same time and hand off to the appropriate one). Therefore:
- our new flow is attempting to validate against the old schema and some questions will fail if they are different data formats to the original.
- The final submit application page will fail due to validation.
- Essential requirements met page will not validate. To progress past this question you must not pick any answer.
- Validation messages have not yet been reviewed

### Questions to match the prototype
We still need to build questions for evidence for essential requirements, progressive disclosure question for nice to haves with evidence.

### DIFF
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -92,7 +92,7 @@ def start_brief_response(brief_id):
             current_user.email_address,
         )['briefResponses']
         brief_response_id = brief_response['id']
-        return redirect(url_for('.edit_brief_response', brief_response_id=brief_response_id))
+        return redirect(url_for('.edit_brief_response', brief_id=brief_id, brief_response_id=brief_response_id))
 
     brief_response = data_api_client.find_brief_responses(
         brief_id=brief_id,
@@ -105,7 +105,7 @@ def start_brief_response(brief_id):
             flash('already_applied', 'error')
             return redirect(url_for(".view_response_result", brief_id=brief_id))
         if brief_response[0].get('status') == 'draft':
-            existing_draft_response = True
+            existing_draft_response = brief_response[0]
     else:
         existing_draft_response = False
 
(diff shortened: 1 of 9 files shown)


app/main/views/briefs.py
```diff
@@ -116,10 +116,107 @@ def start_brief_response(brief_id):
...
+    if role:
+        brief_service = data_api_client.find_services(
+            supplier_id=current_user.supplier_id,
+            framework=brief['frameworkSlug'],
+            status="published",
+            lot=brief["lotSlug"],
+        )["services"][0]
+        max_day_rate = brief_service.get(role + "PriceMax")
+
+    content = content_loader.get_manifest(
```
Wynndow: @tombye `max_day_rate` is needed to get the correct content. `filter` is called on the content object on line 156 and `max_day_rate` is passed in. It's to do with the way the new jinja stuff works. Any variable you want to be available to jinja get's passed in here. It gets consumed [here](https://github.com/alphagov/digitalmarketplace-frameworks/blob/ess-req-met/frameworks/digital-outcomes-and-specialists/questions/brief-responses/dayRate.yml#L3).

tombye:
All the things I commented on are either not a problem or resolved so 👍

PR #567, opened 2016-11-25
[#134252361] avoid S3 connection when no files being uploaded

galund:
We only want to connect to S3 if we are actually going to use that connection to upload 1 or more files.


TheDoubleK:
Looks good :+1:

PR #569, opened 2016-11-29
encrypt email tokens

leohemsted:
stops personally identifiable information being committed to logs. also clean up file so we're not calling `.get` to get values out of dicts that we always need to be there. also don't pass in `role` to decode_token anymore

bump dmutils to 23.0.0 for this

- [x] https://github.com/alphagov/digitalmarketplace-utils/pull/288

### DIFF
--- a/app/main/views/login.py
+++ b/app/main/views/login.py
@@ -19,7 +19,7 @@ from ... import data_api_client
 def create_user(encoded_token):
     form = CreateUserForm()
 
-    token = decode_invitation_token(encoded_token, role='supplier')
+    token = decode_invitation_token(encoded_token)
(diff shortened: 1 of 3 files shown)


pcraig3:
Should be 👍  once tests are green.

leohemsted:
@pcraig3 https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/404 will need to merged first, since that's got where the decryption happens, and that needs to be out in the wild before we start encrypting tokens

pcraig3:
Right, we'll I've just rebased that one as well, so they can both go out today, hopefully.
Good point though.

PR #570, opened 2016-12-01
framework_dashboard: heed framework.clarificationQuestionsOpen

risicle:
to determine whether to show prompts for clarification questions instead of just framework.state being "open".

Add a test that covers this.

Would be quite good to get this out quickly as it's currently showing misleading text on production.


TheDoubleK:
Has this content been signed off?  The old version had "clarification question responses", which might be clearer?

risicle:
I'm going by "what's minimally different from what's already deployed". The content was checked, I'm just changing the flag that triggers it.

PR #571, opened 2016-12-02
Turn off new supplier flow feature flag on staging

idavidmcdonald:
This is by request of Andrew so staging is the same as production and we can use it for testing the existing supplier flow.


tombye:
👍

PR #572, opened 2016-12-02
Add new watcher for frameworks to make framework dev nicer

galund:
New watcher for frameworks to make framework dev nicer


tombye:
Could we use this as an opportunity to remove the 'FAILED' in the readme after the bit about the watch task? Can't find a reason it's there but it's no adding anything to the understanding of this.

pcraig3:
Yeah, it looks like a mistake to me.

pcraig3:
But  👍  otherwise.

PR #574, opened 2016-12-06
Allow end to end flow for new brief responses

Wynndow:
Part of this story on Pivotal [https://www.pivotaltracker.com/story/show/129843611](https://www.pivotaltracker.com/story/show/129843611).

This PR is a small update to the supplier frontend which allows the new style brief-responses to work end to end. Without it, new style brief responses would cause the app to fall over after submission.

There is a check made when viewing the response result that all of the `essentialRequirements` are True. The new style brief responses do not have this key, instead it they have `essentialRequirementsMet` which is just a boolean. This PR makes the app check if this new key exists and if it's True, before checking the original `essentialRequirements` key.

### DIFF
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -86,7 +86,7 @@ def start_brief_response(brief_id):
 
     if not (datetime.strptime(current_app.config['FEATURE_FLAGS_NEW_SUPPLIER_FLOW'], "%Y-%m-%d")
             <= datetime.strptime(brief['publishedAt'], DATETIME_FORMAT)):
-        return redirect(url_for('.brief_response', brief_id=brief['id']))
+        abort(404)
 
     if not is_supplier_eligible_for_brief(data_api_client, current_user.supplier_id, brief):
         return _render_not_eligible_for_brief_error_page(brief)
@@ -233,7 +233,7 @@ def brief_response(brief_id):
 
     if (datetime.strptime(current_app.config['FEATURE_FLAGS_NEW_SUPPLIER_FLOW'], "%Y-%m-%d")
             <= datetime.strptime(brief['publishedAt'], DATETIME_FORMAT)):
-        return redirect(url_for('.start_brief_response', brief_id=brief['id']))
+        abort(404)
 
     if not is_supplier_eligible_for_brief(data_api_client, current_user.supplier_id, brief):
         return _render_not_eligible_for_brief_error_page(brief)
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -3,7 +3,9 @@ from __future__ import unicode_literals
 
 import re
 
-from flask import abort, flash, redirect, render_template, request, url_for
+from datetime import datetime
+
+from flask import abort, flash, redirect, render_template, request, url_for, current_app
 from flask_login import current_user
 import flask_featureflags as feature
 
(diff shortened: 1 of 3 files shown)


app/main/views/briefs.py
```diff
@@ -81,6 +83,10 @@ def ask_brief_clarification_question(brief_id):
 def start_brief_response(brief_id):
     brief = get_brief(data_api_client, brief_id, allowed_statuses=['live'])
 
+    if not (datetime.strptime(current_app.config['FEATURE_FLAGS_NEW_SUPPLIER_FLOW'], "%Y-%m-%d")
+            <= datetime.strptime(brief['publishedAt'][0:10], "%Y-%m-%d")):
```
tombye: Cutting a date_string seems like a slightly brittle way of zero'ing the rest of it beyond year, month and day.

I say that not knowing how solid the format is and knowing little about best practice in Python so take this with a pinch of salt.

Wynndow: I found another way! Have imported the constant from dmutils that is used to convert the `published_at` date of the brief to string when serializing it in the api. It can be used to do the reverse and create a date object in the supplier frontend.

idavidmcdonald:
If a user visits the incorrect flow should we redirect or should we 404? I'd personally lean towards 404 but could go either way. I don't know if we have any precedent of this in our apps?

Wynndow:
Yeah, it's a good point. Part of me thinks that it'd be more helpful to redirect, but on the other hand a user would need to manually change the URL before they experience the 'help', so perhaps a 404 is better.

@tombye, what are your thoughts?

tombye:
I think it's a 404 too. The URLs we'd redirect have/will have never existed for those briefs so my feeling is we shouldn't enable their use or at least support it with a redirect.

Wynndow:
Cool. I've chucked another commit in which 404's instead of redirecting.

idavidmcdonald:
Might need to do similar how we have with the API and check if a feature flag is off then the redirect code does what we want it to

Wynndow:
Good shout. If the flag was set to False/not set it was causing a type error when using the `/create` route. Have updated and added a couple of tests.

app/main/views/briefs.py
```diff
@@ -81,6 +84,10 @@ def ask_brief_clarification_question(brief_id):
 def start_brief_response(brief_id):
     brief = get_brief(data_api_client, brief_id, allowed_statuses=['live'])
 
+    if not (datetime.strptime(current_app.config['FEATURE_FLAGS_NEW_SUPPLIER_FLOW'], "%Y-%m-%d")
```
idavidmcdonald: It looks like there is a test for it that passes but if I ran this just in my terminal I think it would break?
image

Wynndow: That view is behind the `@feature.is_active_feature('NEW_SUPPLIER_FLOW')` decorator. If the flag is false, then the route will 404 straight off the bat without going anywhere near this code, so I think it's fine?

idavidmcdonald: Nice, sorry I didn't spot that! All good.

tests/app/main/test_briefs.py
```diff
@@ -651,6 +651,7 @@ def setup(self):
         self.brief = api_stubs.brief(status='live', lot_slug='digital-specialists')
         self.brief['briefs']['essentialRequirements'] = ['Essential one', 'Essential two', 'Essential three']
         self.brief['briefs']['niceToHaveRequirements'] = ['Nice one', 'Top one', 'Get sorted']
+        self.brief['briefs']['publishedAt'] = '2016-10-25T12:00:00.000000Z'
```
Wynndow: @idavidmcdonald Cool, just added one.

idavidmcdonald:
Should be good after a rebase and tests pass 👍

PR #575, opened 2016-12-07
[#134427911] Add tests for markdown and html escaping brief content

idavidmcdonald:
For [this pivotal story](https://www.pivotaltracker.com/story/show/134427911).

Our IT health check flagged issues with XSS vulnerabilities for brief data. These have been fixed in the content loader however we also wish to double check they have indeed been fixed on the front end (as there is the possibility to override the content loader and put unsafe content into pages). 

Whilst we are doing so, we also check that markdown is escaped (again the content loader should be doing this but could be overwritten).

There are several different brief response fields playing back information (essential requirements, nice to have requirements, start date, day rate). The pivotal story lists exactly what pages need testing and what for (i.e. xss/markdown/both).

### DIFF
--- a/tests/app/main/test_briefs.py
+++ b/tests/app/main/test_briefs.py
@@ -1426,6 +1426,41 @@ class TestResponseResultPage(BaseApplicationTest):
         with self.app.test_client():
             self.login()
 
+    def test_view_response_result_page_escapes_essential_and_nice_to_have_requirements(self, data_api_client):
+        self.brief['briefs']['essentialRequirements'] = [
+            '<h1>Essential one with xss</h1>',
+            '**Essential two with markdown**'
+        ]
+        self.brief['briefs']['niceToHaveRequirements'] = [
+            '<h1>n2h one with xss</h1>',
+            '**n2h two with markdown**'
+        ]
+
+        data_api_client.get_brief.return_value = self.brief
+        data_api_client.get_framework.return_value = self.framework
+        data_api_client.is_supplier_eligible_for_brief.return_value = True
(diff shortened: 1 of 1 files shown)


tests/app/main/test_briefs.py
```diff
@@ -532,6 +532,23 @@ def test_essential_requirements_met_question_replays_all_brief_requirements(self
...
+        self.data_api_client.get_brief.return_value = self.brief
+
+        res = self.client.get(
+            '/suppliers/opportunities/1234/responses/5/do-you-have-all-the-essential-skills-and-experience'
+        )
+        assert res.status_code == 200
+
+        doc = html.fromstring(res.get_data(as_text=True))
+        list_items = doc.xpath("//*[@id='input-essentialRequirementsMet-question-advice']/ul/li/text()")
+        assert list_items[0] == '<h1>Essential one with xss</h1>'
```
TheDoubleK: Not sure if it would make the test clearer to assert the escaped version is in the page - i.e. `&lt;h1&gt;Essential one with xss&lt;/h1&gt;`.
Although it's nicer to render the HTML and have the xpath look in the right place, so maybe how it is is fine.

TheDoubleK:
:+1: for the fourth wall.

idavidmcdonald:
Cheers Kev. Discussed this in person with Tom and we are adding some comments just to explain something that might not be immediately obvious. Then will be good to go.

tombye:
Yeah, I was blocked with openssl problems but now pushed changes to make it clearer the escaping has happened.

PR #576, opened 2016-12-07
Fix crash if framework's "lot question" is inconsistent with lots in DB.

galund:
Probably only causes issues when we are prototyping a new framework, but we'll definitely need to do this again in future, and obscure crashes aren't good.

Possibly a better version of this fix would assume the database to be correct, i.e. provide defaults for values (like 'lot description') that come from the framework/yaml question. Getting this committed for now so that we can support the PaaS testing ASAP.


allait:
Closing this since this only is useful for the prototyping environment

PR #577, opened 2016-12-14
[Proposal] Upgrade pip every time to kill the cryptography bug

pcraig3:
We've had a problem installing cryptography on local machines,
one solution to which is to make sure your is updated.

This pull request would add a step to the Makefile to always
upgrade pip before installing the rest of the requirements.

If we want to do this, I can port it to the other apps.

[Issue discussion on github](https://github.com/pyca/cryptography/issues/2350#issuecomment-179321979)

We could also just put something into the README, if that was a lower-overhead thing people would me more comfortable with.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,10 +16,13 @@ run_app: show_environment virtualenv
 virtualenv:
 	[ -z $$VIRTUAL_ENV ] && [ ! -d venv ] && virtualenv venv || true
 
-requirements: virtualenv requirements.txt
+upgrade_pip: virtualenv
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -16,10 +16,13 @@ run_app: show_environment virtualenv
 virtualenv:
 	[ -z $$VIRTUAL_ENV ] && [ ! -d venv ] && virtualenv venv || true
 
-requirements: virtualenv requirements.txt
+upgrade_pip:
```
allait: This depends on virtualenv

allait:
Yeah, seems reasonable 👍

PR #578, opened 2016-12-15
	Show one question per page for brief response rather than section

idavidmcdonald:
We used to show one section per page. Now we will show one question per page instead. We have added new template `_base_edit_question_page.html` which follows a very similar structure to a template with the same purpose in the buyer frontend. This will enable us to render multiquestions using this template as well as single questions (something which was not supported by the previous `base_edit_section_page` template).

This change will change the url structure from section names to question id's. Apart from that, the flow should remain exactly the same.

Tests will fail till [this frameworks pull request](https://github.com/alphagov/digitalmarketplace-frameworks/pull/345) has been merged.

image


app/main/views/briefs.py
```diff
@@ -162,21 +162,27 @@ def edit_brief_response(brief_id, brief_response_id, section_id=None):
         brief['frameworkSlug'], 'edit_brief_response'
     ).filter({'lot': lot['slug'], 'brief': brief, 'max_day_rate': max_day_rate})
 
-    if section_id is None:
+    section = content.get_section(content.get_next_editable_section_id())
+    if section is None or not section.editable:
```
Wynndow: If a section isn't editable do we definitely want to abort? What if in the future there was a brief type that could have a none editable section in it (for information maybe as you went through the flow)?
I might be talking rubbish and that may never happen though...this is just speculation...

app/templates/briefs/edit_brief_response_question.html
```diff
@@ -0,0 +1,33 @@
...
+    ]
+  %}
+    {% include "toolkit/breadcrumb.html" %}
+  {% endwith %}
+{% endblock %}
+
+{% block save_button %}
+  {%
+    with
+    label="Continue" if not is_last_page else "Submit application",
```
Wynndow: Maybe `"Submit application" if is_last_page else "Continue"` to get rid of the negative

idavidmcdonald: Makes sense. Have changed.

app/main/views/briefs.py
Wynndow: Cool

Wynndow:
👍 
When tests go green

PR #581, opened 2016-12-23
Essential requirements evidence

idavidmcdonald:
[Pivotal story](https://www.pivotaltracker.com/story/show/129841565)

- Pulls in latest frameworks repo giving us a new page for the new `essentialRequirements` dynamic list question
- Corresponding tests for the new page
- Uses the content loader `unformat_data` to transform brief response data into form data (see nice example of unformatting [here](https://github.com/alphagov/digitalmarketplace-content-loader/blob/master/dmcontent/questions.py#L315) in the content loader)

screen shot 2017-01-04 at 12 26 14

### DIFF
--- a/tests/app/main/test_briefs.py
+++ b/tests/app/main/test_briefs.py
@@ -369,7 +369,10 @@ class TestApplyToBrief(BaseApplicationTest):
         self.brief['briefs']['essentialRequirements'] = ['Essential one', 'Essential two', 'Essential three']
         self.brief['briefs']['niceToHaveRequirements'] = ['Nice one', 'Top one', 'Get sorted']
 
-        lots = [api_stubs.lot(slug="digital-specialists", allows_brief=True)]
+        lots = [
+            api_stubs.lot(slug="digital-specialists", allows_brief=True),
+            api_stubs.lot(slug="digital-outcomes", allows_brief=True)
+        ]
         self.framework = api_stubs.framework(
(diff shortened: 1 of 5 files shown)


tests/app/main/test_briefs.py
```diff
@@ -676,11 +676,9 @@ def test_essential_requirements_evidence_escapes_brief_data(self):
         doc = html.fromstring(res.get_data(as_text=True))
         questions = doc.xpath(
             "//*[@class='question-heading']")
-        # if item is excaped correctly it will appear as text rather than an element
+
         assert questions[0].find("h1") is None
-        assert questions[0].text.strip() == '&lt;h1&gt;Essential one with xss&lt;/h1&gt;'
```
galund: Hang on, if they are appearing as literal '<h1>' then they *aren't* being escaped, this sounds like the test is finding a bug?

Wynndow: Is my understanding of 'escaped' off? Do html tags need to be converted to special characters?

idavidmcdonald: But does the assertion above confirm there is no h1 element (which would imply xss)?

idavidmcdonald: image
Have checked manually and they are not rendered as html elements but rendered as a string which means we have not been xss'd?

tombye:
👍

PR #582, opened 2017-01-09
Add nice to have question to new supplier flow

Wynndow:
Part of this story on Pivotal [https://www.pivotaltracker.com/story/show/129841633](https://www.pivotaltracker.com/story/show/129841633).

This PR updates the supplier frontend to deal with the new nice-to-have-requirements question

The nice-to-have-requirements question uses the progressive disclosure pattern to collect the suppliers evidence. This requires an updated version of the govuk frontend toolkit.

The new version of the content loader fixes a bug with the unpacking of brief_response data for dynamic lists questions.

Finally, a small update to the `update_brief_response` view to only replay the form data from a POST request if the request fails validation. Previously the form date was being merged with the service data already pulled from the api, and this was causing issues with data being displayed when it shouldn't have been.

Also tests.

### DIFF
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -203,7 +203,7 @@ def edit_brief_response(brief_id, brief_response_id, question_id=None):
         except HTTPError as e:
             errors = question.get_error_messages(e.message)
             status_code = 400
-            service_data.update(question.unformat_data(question.get_data(request.form)))
+            service_data = question.unformat_data(question.get_data(request.form))
 
         else:
             if next_question_id:
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -170,12 +170,12 @@ def edit_brief_response(brief_id, brief_response_id, question_id=None):
     # data. The question will be skipped in the brief response flow (see below). If a user attempts to access the
     # question by directly visiting the url, this check will return a 404. It has been created specifically for nice to
     # have requirements, and works because briefs and responses share the same key for this question/response.
-    if question_id in list(brief.keys()) and not brief[question_id]:
+    if question_id in brief.keys() and not brief[question_id]:
         abort(404)
 
     # If a question is to be skipped in the normal flow (due to the reason above), we update the next_question_id.
     next_question_id = section.get_next_question_id(question_id)
-    if next_question_id in list(brief.keys()) and not brief[next_question_id]:
+    if next_question_id in brief.keys() and not brief[next_question_id]:
         next_question_id = section.get_next_question_id(next_question_id)
 
     def redirect_to_next_page():
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -166,10 +166,19 @@ def edit_brief_response(brief_id, brief_response_id, question_id=None):
     if section is None or not section.editable:
         abort(404)
 
-    next_question_id = section.get_next_question_id(question_id)
+    # If a question in a brief is optional and is unanswered by the buyer, the brief will have the key but will have no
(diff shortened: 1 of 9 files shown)


tombye:
Going through the new flow while applying to a brief with no nice-to-have questions I noticed the page for this is still shown. We shouldn't show this question in this case.

Wynndow:
@tombye Yep, that was a great shout. We've added a couple of bits to skip an optional question if it doesn't need answering, and to return 404 if someone tries to access said question directly.

app/main/views/briefs.py
```diff
@@ -166,10 +166,19 @@ def edit_brief_response(brief_id, brief_response_id, question_id=None):
...
+    # If a question in a brief is optional and is unanswered by the buyer, the brief will have the key but will have no
+    # data. The question will be skipped in the brief response flow (see below). If a user attempts to access the
+    # question by directly visiting the url, this check will return a 404. It has been created specifically for nice to
+    # have requirements, and works because briefs and responses share the same key for this question/response.
+    if question_id in list(brief.keys()) and not brief[question_id]:
+        abort(404)
+
+    # If a question is to be skipped in the normal flow (due to the reason above), we update the next_question_id.
     next_question_id = section.get_next_question_id(question_id)
+    if next_question_id in list(brief.keys()) and not brief[next_question_id]:
```
benvand: Looks like you could use:
`if not breif.get(question_id, True) or True`
`if not breif.get(next_question_id, True) or True`
or
`if not breif.get(question_id, '') or True`
`if not breif.get(next_question_id, '') or True`
either of which are _probably_ slightly more efficient but I like the readability of your solution. No need to change unless a microscopic performance boost is a concern here.

Wynndow: @benvand By having the `or True`, isn't the conditional going to evaluate to `True` regardless of if it finds the key in brief or not?

app/main/views/briefs.py
```diff
@@ -166,10 +166,19 @@ def edit_brief_response(brief_id, brief_response_id, question_id=None):
...
+        next_question_id = section.get_next_question_id(next_question_id)
 
-    # If no question_id in url then redirect to first question
-    if question_id is None:
+    def redirect_to_next_page():
```
benvand: 💯 Like it

app/main/views/briefs.py
```diff
@@ -166,10 +166,19 @@ def edit_brief_response(brief_id, brief_response_id, question_id=None):
     if section is None or not section.editable:
         abort(404)
 
+    # If a question in a brief is optional and is unanswered by the buyer, the brief will have the key but will have no
+    # data. The question will be skipped in the brief response flow (see below). If a user attempts to access the
+    # question by directly visiting the url, this check will return a 404. It has been created specifically for nice to
+    # have requirements, and works because briefs and responses share the same key for this question/response.
+    if question_id in brief.keys() and not brief[question_id]:
```
benvand: Could store `brief.keys()` as `questions`. Would that make sense?
```
if question_id in questions and not brief[question_id]:
    # if question was asked and not answered 
```
Not a dealbreaker

app/main/views/briefs.py
benvand: Ugh sorry I was in a rush and thinking of something else. Meant:
```
if brief.get(question_id, None) is '':
```

tombye:
Looks good to me 👍

app/main/views/briefs.py
Wynndow: Problem is the question this is written to target is nice to have requirements, which would be an empty list, not an empty string.
By specifying the type it potentially restricts it being used by other question types.

PR #583, opened 2017-01-10
Don't call filter() unnecessarily

TheDoubleK:
Supplier declarations don't have any `depends` dependencies in content, so the call to filter here doesn't achieve anything.

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -265,7 +265,7 @@ def framework_submission_services(framework_slug, lot_slug):
 def framework_supplier_declaration(framework_slug, section_id=None):
     framework = get_framework(data_api_client, framework_slug, allowed_statuses=['open'])
 
-    content = content_loader.get_manifest(framework_slug, 'declaration').filter({})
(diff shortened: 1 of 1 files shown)


allait:
Filter also runs questions through Jinja/markdown now, so in order to avoid any errors if the questions change we try to call it everywhere now. This line might be something @pcraig3 and I added when we were working on the content-loader templating.

TheDoubleK:
Ah right, OK.  This was prompted by finding a bunch of tests in the buyer frontend (service presenters) that called `filter({})` on service listings and then ended up testing nothing because all service questions have a `depends on lot...` and so all get removed by an empty filter.

allait:
@TheDoubleK yeah, that sounds like our fault when we were adding `.filter` calls everywhere. Did the tests get fixed?

TheDoubleK:
Yep the test fixes are merged already.  I just went back afterwards to remove any other apparently unnecessary ones from code (there were only two uses of a `filter({})` - one here and one in the buyer frontend).

PR #584, opened 2017-01-10
[#136029251]service editing: add "service updated" flash message & tests

risicle:
Story https://www.pivotaltracker.com/story/show/136029251

(tl;dr we're adding a flash message that says "You’ve edited your service. The changes are now live on the Digital Marketplace.")

20170110_service_editing_flash_message

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -167,6 +167,8 @@ def update_section(service_id, section_id):
             errors=errors
         )
 
+    flash(u"You\u2019ve edited your service. The changes are now live on the Digital Marketplace.", 'message')
+
     return redirect(url_for(".edit_service", service_id=service_id))
 
 
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -167,7 +167,7 @@ def update_section(service_id, section_id):
             errors=errors
         )
 
-    flash(u"You\u2019ve edited your service. The changes are now live on the Digital Marketplace.", 'message')
+    flash({"service_name": posted_data.get("serviceName") or service.get("serviceName")}, 'service_updated')
 
     return redirect(url_for(".edit_service", service_id=service_id))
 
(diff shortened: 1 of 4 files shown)


app/main/views/services.py
```diff
@@ -150,6 +150,8 @@ def update_section(service_id, section_id):
             errors=errors
         )
 
+    flash(u"You\u2019ve edited your service. The changes are now live on the Digital Marketplace.", 'message')
```
TheDoubleK: We generally flash a short message key from the view and have the long message text itself in the template file.  We've had extensive discussion about this in the past, and the reason behind doing it this way is that if the content of messages needs to be updated then the template is the natural place to look to do that, rather than in the backend view code.

tests/app/main/test_services.py
```diff
@@ -253,12 +253,69 @@ def test_should_view_public_service_with_correct_message(
...
+
+        # service removed message should not have been triggered
+        self.assert_not_in_strip_whitespace(
+            'Service name 123 has been removed.',
+            res.get_data(as_text=True)
+        )
+
+        # service removed notification banner shouldn't be there
+        self.assert_not_in_strip_whitespace(
+            '<h2>This service was removed on Monday 23 March 2015</h2>',
```
TheDoubleK: Where does the date here come from? I guess it must be in the fixture somehow but I can't see where?  

Still, maybe just asserting not `<h2>This service was removed on` would cover more possible failure cases?

app/main/views/services.py
risicle: Only reason I did it this way is it's currently the only way we have of delivering this in a non-destination-page-specific (and hence robust) way. And there's no way it can accidentally be misinterpreted by the wrong template and dumped to the screen as a raw dictionary as is possible in some cases.

tests/app/main/test_services.py
risicle: Good point - it was just copied from L257

TheDoubleK:
:+1: after a rebase

PR #587, opened 2017-01-19
Moar framework agreement urls

TheDoubleK:
This will also fix the email template date for DOS2 emails.


idavidmcdonald:
👍 after tests

PR #588, opened 2017-01-20
Add previous page links

idavidmcdonald:
For [this pivotal story](https://www.pivotaltracker.com/story/show/137359355)

### DIFF
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -198,12 +198,8 @@ def edit_brief_response(brief_id, brief_response_id, question_id=None):
     # Unformat brief response into data for form
     service_data = question.unformat_data(brief_response)
 
-    # This line can soon be removed when we no longer show boolean list questions as part of the new flow
-    section.inject_brief_questions_into_boolean_list_question(brief)
-
     status_code = 200
     errors = {}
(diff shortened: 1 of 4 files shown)


pcraig3:
👍

app/main/views/briefs.py
```diff
@@ -198,12 +198,8 @@ def redirect_to_next_page():
     # Unformat brief response into data for form
     service_data = question.unformat_data(brief_response)
 
-    # This line can soon be removed when we no longer show boolean list questions as part of the new flow
-    section.inject_brief_questions_into_boolean_list_question(brief)
-
```
pcraig3: This looks good.

Any chance we can get rid of `inject_brief_questions_into_boolean_list_question` entirely, or do we need to keep the old flow alive for some reason?

idavidmcdonald: Great question. This can't be deprecated for atleast 2 weeks after the new flow is launched (by then no one will be applying using the old flow).

I'll have to double check with more specifics whether this can be removed in general. I've added this as a note to our remove old flow and tidy up pivotal story - https://www.pivotaltracker.com/story/show/130019195

PR #589, opened 2017-01-20
[#137344025]supplier service editing: straighten out framework-determined behaviours for edit & remove views

risicle:
Story https://www.pivotaltracker.com/story/show/137344025

The first thing done here is to significantly refactor the tests around the edit & remove views.

The views are then fixed to work properly cross-framework. Frameworks without an `edit_service` content manifest will not allow editing **or** removal by a supplier. Not this latter change in behaviour - previously the removal of a e.g. dos service was not enforced by the actual remove view. The behaviour of the edit page for a dos supplier was... undefined (probably a 500). These are now 404s.

### DIFF
--- a/tests/app/main/test_services.py
+++ b/tests/app/main/test_services.py
@@ -11,18 +11,37 @@ import pytest
 from lxml import html
 from freezegun import freeze_time
 
-from nose.tools import assert_equal, assert_true, assert_false, assert_in, assert_not_in
+from nose.tools import assert_equal, assert_true, assert_false, assert_in, assert_not_in, assert_is
 from tests.app.helpers import BaseApplicationTest, empty_g7_draft_service
 
 
-@pytest.fixture(params=["g-cloud-6", "g-cloud-7"])
+# this is mostly a workaround for pytest not being able to do parametrization with unittest-derived class methods
+# otherwise it doesn't get us much over function parametrization...
+@pytest.fixture(params=(
+    ("g-cloud-6", "G-Cloud 6",),
+    ("g-cloud-7", "G-Cloud 7",),
+))
 def fixture_framework_slug_and_name(request):
-    frameworks = {
-        'g-cloud-6': ('g-cloud-6', 'G-Cloud 6'),
-        'g-cloud-7': ('g-cloud-7', 'G-Cloud 7')
-    }
+    return request.param
 
-    return frameworks[request.param]
+
+# and these fixtures really are just a workaround to not being able to use plain function parametrization in unittest-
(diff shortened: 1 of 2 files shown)


tests/app/main/test_services.py
```diff
@@ -303,15 +330,48 @@ def test_should_view_disabled_service_with_removed_message(
...
+    def test_should_not_view_other_suppliers_services(
+            self, data_api_client, supplier_service_editing_fw_params
+    ):
+        framework_slug, framework_name, framework_editable_services = supplier_service_editing_fw_params
+        self.login()
+        self._setup_service(
+            data_api_client, framework_slug, framework_name, service_status='published', service_belongs_to_user=False)
 
+        res = self.client.get('/suppliers/services/123')
+        if not framework_editable_services:
```
TheDoubleK: Is this check necessary as we're expecting a 404 either way?

tests/app/main/test_services.py
```diff
@@ -303,15 +330,48 @@ def test_should_view_disabled_service_with_removed_message(
...
+        framework_slug, framework_name, framework_editable_services = supplier_service_editing_fw_params
         self.login()
-        self._get_service(data_api_client, fixture_framework_slug_and_name, service_status='published')
+        self._setup_service(data_api_client, framework_slug, framework_name, service_status='published')
 
+        # NOTE two http requests performed here
         res = self.client.post('/suppliers/services/123/remove', follow_redirects=True)
+        if not framework_editable_services:
+            assert_equal(res.status_code, 404)
+            assert_is(data_api_client.update_service_status.called, False)
```
TheDoubleK: Aren't we killing nose methods rather than introducing more?

risicle: The nose-removal patch sits on top of this one (*I think*) so the nose-removal happens after this - I guess I just wanted to keep it symmetrical with the rest of the code as shown?

tests/app/main/test_services.py
risicle: Ah yes indeed well spotted, I flipped a coin when I came to this one - did I want to add an unnecessary clause or make it less obvious that this test follows the same basic flow as the others. Scanning the code by eye would bring up a "hold on he's missed something here" moment. But looking at it now it does look kinda stupid.

Still, I don't want to have to do a chain-rebase... :hear_no_evil:

risicle: Hold on I need to rebase anyway. Will fix.

risicle: (I think I also had a bit of a :-\ should this really be a 403 instead? moment)

PR #590, opened 2017-01-23
Wrap email links in anchor tags

TheDoubleK:
For this bug: https://www.pivotaltracker.com/story/show/136103557

Some users have had issues with clicking the link in their invitation emails. This may help alleviate the problem.


allait:
:+1:

PR #592, opened 2017-01-24
Old test framework purge: supplier frontend edition

risicle:
~~This sits on top of https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/589 because the alternative would have caused muchos merge conflicts with it, so will have to wait till that's in.~~

Most significant thing here is the removal of nose, but some uses of mock's `.*called*` assertions have been converted to less error-prone forms.

### DIFF
--- a/tests/app/main/test_services.py
+++ b/tests/app/main/test_services.py
@@ -534,8 +534,8 @@ class TestSupplierEditUpdateServiceSection(BaseApplicationTest):
         }
     }
 
-    def setup(self):
-        super(TestSupplierEditUpdateServiceSection, self).setup()
+    def setup_method(self, method):
+        super(TestSupplierEditUpdateServiceSection, self).setup_method(method)
         with self.app.test_client():
             self.login()
 
@@ -662,8 +662,8 @@ class TestSupplierEditUpdateServiceSection(BaseApplicationTest):
 
 @mock.patch('app.main.views.services.data_api_client', autospec=True)
 class TestCreateDraftService(BaseApplicationTest):
-    def setup(self):
-        super(TestCreateDraftService, self).setup()
+    def setup_method(self, method):
+        super(TestCreateDraftService, self).setup_method(method)
         self._answer_required = 'Answer is required'
         self._validation_error = 'There was a problem with your answer to:'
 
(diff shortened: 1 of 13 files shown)


TheDoubleK:
Poke me to review this once #589 gets merged.

tests/app/main/test_services.py
```diff
@@ -301,14 +291,12 @@ def test_should_view_public_service_with_update_message(
...
+            assert res.status_code == 404
             return
 
-        assert_equal(res.status_code, 200)
+        assert res.status_code == 200
 
-        assert_true(
-            'Service name 123' in res.get_data(as_text=True)
-        )
+        assert'Service name 123' in res.get_data(as_text=True)
```
TheDoubleK: I'm amazed PEP8 doesn't choke on this, but could do with a space after `assert` here.

tests/app/main/test_services.py
```diff
@@ -349,13 +337,11 @@ def test_should_view_private_service_with_correct_message(
...
-            assert_equal(res.status_code, 404)
+            assert res.status_code == 404
             return
 
-        assert_equal(res.status_code, 200)
-        assert_true(
-            'Service name 123' in res.get_data(as_text=True)
-        )
+        assert res.status_code == 200
+        assert'Service name 123' in res.get_data(as_text=True)
```
TheDoubleK: Also here.

tests/app/main/test_services.py
risicle: Blimey yeah, surprised it's even syntactically valid

tests/app/main/test_services.py
```diff
@@ -1451,8 +1422,8 @@ def test_unanswered_questions_count(self, count_unanswered, data_api_client):
         count_unanswered.return_value = 1, 2
         res = self.client.get('/suppliers/frameworks/g-cloud-7/submissions/scs/1')
 
-        assert_true(u'3 unanswered questions' in res.get_data(as_text=True),
-                    "'3 unanswered questions' not found in html")
+        assert u'3 unanswered questions' in res.get_data(as_text=True), \
+            "'3 unanswered questions' not found in html"
```
TheDoubleK: Weird how this is about the only place we set a custom failure message on an assert, but it's maintaining current behaviour so fine.

risicle: Yeah maybe the original author thought it was a particularly hard to track down eroor message by default

TheDoubleK:
I appear to have internalised a Python linter, which is slightly troubling, but :clap: :clap: - nice to clear all this out.
:+1:

risicle:
yays

PR #593, opened 2017-01-24
Make brief response results page work for new flow

tombye:
More specifically, makes the brief response results page work for both formats of data. The tests will not pass until [#363](https://github.com/alphagov/digitalmarketplace-frameworks/pull/363) on the content loader is merged.

new_flow_results_page

This almost doubles the amount of tests for the page so is best read commit-by-commit.

### DIFF
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -358,7 +358,12 @@ def view_response_result(brief_id):
 
     if len(brief_response) == 0:
         return redirect(url_for(".brief_response", brief_id=brief_id))
-    elif brief_response[0].get('essentialRequirementsMet') or all(brief_response[0]['essentialRequirements']):
+
+    # The legacy brief response data format has a booleans for each essential requirement
+    # rather than a single flag
+    if ('essentialRequirementsMet' not in brief_response[0]
+        and all(brief_response[0]['essentialRequirements'])
+            or brief_response[0].get('essentialRequirementsMet')):
         result_state = 'submitted_ok'
     else:
         result_state = 'submitted_unsuccessful'
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -358,12 +358,7 @@ def view_response_result(brief_id):
 
     if len(brief_response) == 0:
         return redirect(url_for(".brief_response", brief_id=brief_id))
-
-    # The legacy brief response data format has a booleans for each essential requirement
-    # rather than a single flag
-    if ('essentialRequirementsMet' not in brief_response[0]
-        and all(brief_response[0]['essentialRequirements'])
-            or brief_response[0].get('essentialRequirementsMet')):
+    elif brief_response[0].get('essentialRequirementsMet') or all(brief_response[0]['essentialRequirements']):
         result_state = 'submitted_ok'
     else:
         result_state = 'submitted_unsuccessful'
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -356,8 +356,17 @@ def view_response_result(brief_id):
         supplier_id=current_user.supplier_id
     )['briefResponses']
 
+    legacy_brief = True
+    if current_app.config['FEATURE_FLAGS_NEW_SUPPLIER_FLOW']:
+        legacy_brief = (
+            datetime.strptime(current_app.config['FEATURE_FLAGS_NEW_SUPPLIER_FLOW'], "%Y-%m-%d")
+            > datetime.strptime(brief['publishedAt'], DATETIME_FORMAT))
+
     if len(brief_response) == 0:
-        return redirect(url_for(".brief_response", brief_id=brief_id))
+        if legacy_brief:
+            return redirect(url_for(".brief_response", brief_id=brief_id))
+        else:
+            return redirect(url_for(".start_brief_response", brief_id=brief_id))
     elif brief_response[0].get('essentialRequirementsMet') or all(brief_response[0]['essentialRequirements']):
         result_state = 'submitted_ok'
         flash('submitted_ok', 'success')
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -367,8 +367,13 @@ def view_response_result(brief_id):
     framework, lot = get_framework_and_lot(
         data_api_client, brief['frameworkSlug'], brief['lotSlug'], allowed_statuses=['live'])
 
+    if brief_response.get('essentialRequirementsMet'):
+        brief_response_display_manifest = 'display_brief_response'
+    else:
+        brief_response_display_manifest = 'legacy_display_brief_response'
+
     response_content = content_loader.get_manifest(
-        framework['slug'], 'legacy_display_brief_response').filter({'lot': lot['slug'], 'brief': brief})
+        framework['slug'], brief_response_display_manifest).filter({'lot': lot['slug'], 'brief': brief})
     for section in response_content:
         section.inject_brief_questions_into_boolean_list_question(brief)
 
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -372,6 +372,7 @@ def view_response_result(brief_id):
         flash('submitted_ok', 'success')
     else:
         result_state = 'submitted_unsuccessful'
+        flash('submitted_unsuccessful', 'error')
 
     brief_response = brief_response[0]
     framework, lot = get_framework_and_lot(
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -368,7 +368,7 @@ def view_response_result(brief_id):
         data_api_client, brief['frameworkSlug'], brief['lotSlug'], allowed_statuses=['live'])
 
     response_content = content_loader.get_manifest(
-        framework['slug'], 'legacy_display_brief_response').filter({'lot': lot['slug']})
+        framework['slug'], 'legacy_display_brief_response').filter({'lot': lot['slug'], 'brief': brief})
     for section in response_content:
         section.inject_brief_questions_into_boolean_list_question(brief)
 
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -360,6 +360,7 @@ def view_response_result(brief_id):
         return redirect(url_for(".brief_response", brief_id=brief_id))
     elif brief_response[0].get('essentialRequirementsMet') or all(brief_response[0]['essentialRequirements']):
         result_state = 'submitted_ok'
+        flash('submitted_ok', 'success')
     else:
         result_state = 'submitted_unsuccessful'
 
--- a/app/templates/briefs/_brief_response_data.html
+++ b/app/templates/briefs/_brief_response_data.html
@@ -26,12 +26,22 @@
               {% endfor %}
             {% elif item.type == "dynamic_list" %}
               {% for question in item.questions %}
-                {% call summary.row() %}
-                  {% if question.type != "boolean" %}
-                    {{ summary.field_name(question.question) }}
-                    {{ summary.text(item.unformat_data(brief_response)[question.id]) }}
+                  {% if question.type == "boolean" %}
+                    {% if question._data['followup'] %}
+                      {% call summary.row() %}
+                        {{ summary.field_name(question.question) }}
+                        {{ summary.text(item.unformat_data(brief_response)[question._data['followup']]) }}
+                      {% endcall %}
+                    {% endif %}
+                  {% else %}
+                    {% set followup = item.questions[loop.index0 - 1]._data['followup'] %}
+                    {% if not followup or followup != question.id %}
+                      {% call summary.row() %}
+                        {{ summary.field_name(question.question) }}
+                        {{ summary.text(item.unformat_data(brief_response)[question.id]) }}
+                      {% endcall %}
+                    {% endif %}
                   {% endif %}
-                {% endcall %}
               {% endfor %}
             {% else %}
             {% call summary.row() %}
--- a/app/templates/briefs/_brief_response_data.html
+++ b/app/templates/briefs/_brief_response_data.html
@@ -1,21 +1,33 @@
 {% import "toolkit/summary-table.html" as summary %}
 {% for section in response_content.summary(brief_response) %}
+    {% if section.id in ("your-nice-to-have-skills-and-experience", "your-essential-skills-and-experience") and ('essentialRequirementsMet' in brief_response) %}
+      {% set field_headings = ["Requirement", "Evidence"] %}
+      {% set field_headings_visible = True %}
+    {% else %}
+      {% set field_headings = ["Opportunity attribute name", "Opportunity attribute value"] %}
+      {% set field_headings_visible = False %}
+    {% endif %}
+
+    {# Don't have a table for nice-to-have requirements if none have been given #}
     {% if not (section.id == 'your-nice-to-have-skills-and-experience' and not section.get_question('niceToHaveRequirements').boolean_list_questions) %}
         {{ summary.heading(section.name, id="opportunity-attributes-{}".format(loop.index)) }}
         {% call(item) summary.list_table(
           section.questions,
           caption=section.name,
-          field_headings=[
-            "Opportunity attribute name",
-            "Opportunity attribute value"
-          ],
-          field_headings_visible=False
+          field_headings=field_headings,
+          field_headings_visible=field_headings_visible
         ) %}
             {% if item.type == "boolean_list" %}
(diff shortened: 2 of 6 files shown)


idavidmcdonald:
At a second glance, the `display_brief_response` drives all of those questions using the legacy questions. So it's not just the dynamic list questions that need changing.

Because of this I think we should split the manifests. The reason is that the new flow is fundamentally different for user research participants and will determine what questions they get asked (old flow didn't ask start date, new flow does). Therefore we need to drive this page off the correct questions and can't rely on the fact that they have the same id's like my second suggested solution above.

app/templates/briefs/_brief_response_data.html
```diff
@@ -1,21 +1,33 @@
...
+          field_headings_visible=field_headings_visible
         ) %}
             {% if item.type == "boolean_list" %}
               {% for question in item.boolean_list_questions %}
                 {% call summary.row() %}
                   {{ summary.field_name(question, two_thirds=True) }}
-                  {{ summary['boolean'](item.value[loop.index0]) }}
+                  {% set answer = item.value[loop.index0] -%}
+                  {# Legacy boolean_list response answers are just a list of booleans #}
+                  {% if answer is not mapping -%}
```
tombye: The `-` character is part of jinja's whitespace controls (see http://jinja.pocoo.org/docs/2.9/templates/#whitespace-control).

Apart from that, I put the condition that way around to match the order set in the parent template of default first, then legacy afterwards. Happy to reverse it if that means it reads better.

tombye:
Agreed. There's a few funky workarounds I had to introduce to support the legacy questions (in particular the lack of the `dynamic_list` property). Doing so should make this code a lot better.

app/templates/briefs/_brief_response_data.html
idavidmcdonald: Oo thanks, I learnt something new.

Is the default (the new flow) not the case where it _is_ a mapping rather than not? Or maybe I'm confused

app/main/views/briefs.py
```diff
@@ -347,7 +347,12 @@ def view_response_result(brief_id):
 
     if len(brief_response) == 0:
         return redirect(url_for(".brief_response", brief_id=brief_id))
-    elif brief_response[0].get('essentialRequirementsMet') or all(brief_response[0]['essentialRequirements']):
+
+    # The legacy brief response data format has a booleans for each essential requirement
+    # rather than a single flag
+    if ('essentialRequirementsMet' not in brief_response[0]
+        and all(brief_response[0]['essentialRequirements'])
+            or brief_response[0].get('essentialRequirementsMet')):
```
tombye: That's the version I replaced(!) Turns out my replacing it was wrong. I did so to deal with a situation where you get to this page without `essentialRequirementsMet` being `False` which, it turns out isn't possible with the new flow.

I'll revert this to the older version which should solve it. Good spot.

tombye:
OK so I've added some changes to resolve any points in the comments, use a legacy manifest for the old flow and fix a few things brought up in the design and content crit from Friday.

I realise the commit history is quite long now so some guidance for readers:

1. anyone returning to this PR, the new commits start from `Revert "Rewrite essential requirements check"` 
2. this code needs version `5.10.0` of the frameworks repo which will be available once [#363](https://github.com/alphagov/digitalmarketplace-frameworks/pull/363) is merged
3. the new version of the page tells the user if they've succeeded or not through a notification banner at the top of the screen rather than in the h1

bower.json
```diff
@@ -7,6 +7,6 @@
     "jquery-details": "https://github.com/mathiasbynens/jquery-details/archive/v0.1.0.tar.gz",
     "digitalmarketplace-frontend-toolkit": "https://github.com/alphagov/digitalmarketplace-frontend-toolkit.git#v19.3.1",
     "govuk_template": "https://github.com/alphagov/govuk_template/releases/download/v0.19.2/jinja_govuk_template-0.19.2.tgz",
-    "digitalmarketplace-frameworks": "https://github.com/alphagov/digitalmarketplace-frameworks.git#5.9.1"
+    "digitalmarketplace-frameworks": "https://github.com/alphagov/digitalmarketplace-frameworks.git#5.10.0"
```
idavidmcdonald: Should this be 5.11.0 now?

app/main/views/briefs.py
```diff
@@ -360,15 +360,21 @@ def view_response_result(brief_id):
         return redirect(url_for(".brief_response", brief_id=brief_id))
     elif brief_response[0].get('essentialRequirementsMet') or all(brief_response[0]['essentialRequirements']):
         result_state = 'submitted_ok'
+        flash('submitted_ok', 'success')
```
idavidmcdonald: Does this mean we set a flash message for success everytime you visit the page and have submitted succesfully? Or should this just be shown the first time you complete the flow, i.e. the flash is set on the penultimate page?

tests/app/main/test_briefs.py
```diff
@@ -2086,3 +2092,203 @@ def test_evaluation_methods_shown_with_a_or_an(self, data_api_client):
...
         assert len(doc.xpath('//li[contains(normalize-space(text()), "an interview")]')) == 1
+
+
+@mock.patch("app.main.views.briefs.data_api_client")
+class TestResponseResultPage(ResponseResultPageBothFlows):
+
+    def setup_method(self, method):
+        super(TestResponseResultPage, self).setup_method(method)
+        self.brief['briefs']['niceToHaveRequirements'] = []
+        self.brief['briefs']['dayRate'] = []
```
tombye: Nope that's a mistake, good spot!

tests/app/main/test_briefs.py
```diff
@@ -2086,3 +2092,203 @@ def test_evaluation_methods_shown_with_a_or_an(self, data_api_client):
...
+        assert "&lt;h1&gt;Essential one with xss&lt;/h1&gt;" in data
+        assert "&lt;h1&gt;n2h one with xss&lt;/h1&gt;" in data
+        assert len(doc.xpath('//h1')) == 1
+
+        assert "**Essential two with markdown**" in data
+        assert "<strong>Essential two with markdown</strong>" not in data
+        assert "**n2h two with markdown**" in data
+        assert "<strong>n2h two with markdown</strong>" not in data
+
+    def test_view_response_result_submitted_ok(self, data_api_client):
```
tombye: Makes sense for both old and new flows to be variants of the same test. I'll make the changes.

tests/app/main/test_briefs.py
```diff
@@ -2086,3 +2092,203 @@ def test_evaluation_methods_shown_with_a_or_an(self, data_api_client):
...
+        data_api_client.is_supplier_eligible_for_brief.return_value = True
+        data_api_client.find_brief_responses.return_value = self.brief_responses
+        res = self.client.get('/suppliers/opportunities/1234/responses/result')
+
+        assert res.status_code == 200
+        doc = html.fromstring(res.get_data(as_text=True))
+        assert doc.xpath('//p[contains(@class, "banner-message")]')[0].text.strip() == \
+            "Your application for ‘I need a thing to do a thing’ has been submitted."
+
+    def test_essential_skills_shown_with_response(self, data_api_client):
```
tombye: Agreed. I'll add some tests to see if the actual data is coming through and, that working well, it'll make sense to combine both of those.

app/main/views/briefs.py
tombye: I've spoken to @ralph-hawkins and he says the flash message should be every time you visit the page after a successful submission. He also suggested rationalising how the banners (and h1s) work for both flows. Notes have been added to [the story](https://www.pivotaltracker.com:/story/show/130019135/comments/161850371) which list the changes needed for this.

tests/app/main/test_briefs.py
```diff
@@ -1945,15 +1960,6 @@ def test_view_response_result_submitted_unsuccessful(self, data_api_client):
...
-    def test_view_response_result_not_submitted_redirect_to_submit_page(self, data_api_client):
-        data_api_client.get_brief.return_value = self.brief
-        data_api_client.is_supplier_eligible_for_brief.return_value = True
-        data_api_client.find_brief_responses.return_value = {"briefResponses": []}
-        res = self.client.get('/suppliers/opportunities/1234/responses/result')
-
-        assert res.status_code == 302
-        assert res.location == 'http://localhost/suppliers/opportunities/1234/responses/create'
-
     def test_essential_skills_shown_with_response(self, data_api_client):
```
tombye: No and I think this is the same issue mentioned in https://www.pivotaltracker.com:/story/show/130019135/comments/161850371. I'm gonna take it as such so the fix for that should cover this one too.

tombye:
Right then, seeing as this is now horrible to read, here's a run-through of all the issues brought up in the last review and which commits fix them.

### Redirects from the start page in the new flow need to be tested against the /start route rather than /create

https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/593#discussion_r98444164

Added code to the controller based on the feature flag: https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/593/commits/45d2dfd1e1f01d5951dabe490a79222c3b6c36b1

### Requirements tests don’t test the actual data in the page

https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/593#discussion_r98456522
https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/593#discussion_r98454619

Now actually tests for it, not just it’s heading: https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/593/commits/03561e763397d09f7247c20c5bf1a6e213e16aae

### Day rate is an empty list

https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/593#discussion_r98445282

See: https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/593/commits/68fba3145d7aa2afbf888553fac7ee59c71235a5

### Naming of tests for live and closed briefs

https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/593#discussion_r98453401

Each set is now collapsed into one which is renamed: https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/593/commits/68fba3145d7aa2afbf888553fac7ee59c71235a5

### Repeated code

https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/593#discussion_r98449077

Now moved into the setup method, as requested: https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/593/commits/68fba3145d7aa2afbf888553fac7ee59c71235a5

### Contents of details table

https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/593#discussion_r98454926

Now added: https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/593/commits/2ca21eefb062ec0c5b4d33e5d8ce432f51d8b123

### Version of frameworks out of date

https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/593#discussion_r98458812

This version of the pull request is rebased over the right version now.

### Questions about the flash message

https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/593#discussion_r98464812

Now made the required changes requested by Ralph: https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/593/commits/ff9c2f406dd5d3b83376cfee2466323286667e6b

app/main/views/briefs.py
```diff
@@ -356,19 +356,34 @@ def view_response_result(brief_id):
         supplier_id=current_user.supplier_id
     )['briefResponses']
 
+    legacy_brief = True
```
idavidmcdonald: I was hoping to we wouldn't need to use the feature flag for this view but you are right that we do need to because we may have no brief response so we can't look at `essentialRequirementsMet`. Shame

tests/app/main/test_briefs.py
```diff
@@ -2013,9 +2073,8 @@ def test_nice_to_haves_heading_not_shown_when_there_are_none(self, data_api_clie
         ) == 0
 
     def test_supplier_details_shown(self, data_api_client):
```
idavidmcdonald: Is this test already covered by the shared tests class?

app/main/views/briefs.py
tombye: Yeah, I wasn't keen either.

tests/app/main/test_briefs.py
tombye: It is, good spot, sorry I missed that. I had a look and it turns out that test can't be shared because it relies on different frameworks content depending on the flow.

I put in a fix for this in the latest commit (https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/593/commits/3f4053fad03ae8337141a278976b957fe6aa1538)

tests/app/main/test_briefs.py
```diff
@@ -1900,110 +1958,289 @@ def test_view_response_result_page_escapes_essential_and_nice_to_have_requiremen
...
-            ]
-        }
+        data_api_client.find_brief_responses.return_value = self.brief_responses
         res = self.client.get('/suppliers/opportunities/1234/responses/result')
 
         assert res.status_code == 200
         doc = html.fromstring(res.get_data(as_text=True))
-        assert doc.xpath('//h1')[0].text.strip() == \
-            "Your response to ‘I need a thing to do a thing’ has been sent"
+        assert doc.xpath('//h1')[0].text.strip() == "Your application for ‘I need a thing to do a thing’"
```
idavidmcdonald: Is there nothing else that needs to be asserted in this case? Such as no 'Your application has been submitted' banner message or an alternative banner message?

tests/app/main/test_briefs.py
```diff
@@ -1900,110 +1958,289 @@ def test_view_response_result_page_escapes_essential_and_nice_to_have_requiremen
...
+        self.set_framework_and_eligibility_for_api_client(data_api_client)
+        data_api_client.get_brief.return_value = self.brief
+        data_api_client.find_brief_responses.return_value = self.brief_responses
+        res = self.client.get('/suppliers/opportunities/1234/responses/result')
+
+        assert res.status_code == 200
+        doc = html.fromstring(res.get_data(as_text=True))
+        assert len(doc.xpath('//li[contains(normalize-space(text()), "your day rate exceeds their budget")]')) == 0
+
+    def test_evaluation_methods_load_default_value(self, data_api_client):
```
idavidmcdonald: These two evaluation method tests appear to be exactly the same for both flows and I think can be moved into the shared tests.

tests/app/main/test_briefs.py
tombye: Good point. We've moved information about state into the banner now rather than the h1 so we should also be checking the banner.

tests/app/main/test_briefs.py
tombye: Agreed (and done).

app/templates/briefs/view_response_result.html
```diff
@@ -2,30 +2,60 @@
...
-  {% for category, message in messages %}
+  {% if ("error", "already_applied") in messages %}
     {%
     with
-    message = {
-      "already_applied": "You’ve already applied for ‘{}’ so you can’t apply again".format(brief.title) if result_state == "submitted_ok"
-                    else "You already applied for ‘{}’ but you didn’t meet the essential requirements".format(brief.title)
-    }[message] or message,
-    type = "destructive" if category == "error" else "success"
+    message = "You’ve already applied so you can’t apply again".format(brief.title) if result_state == "submitted_ok"
```
idavidmcdonald: Do we have tests that check these error flash messages?

tests/app/main/test_briefs.py
```diff
@@ -2297,3 +2332,20 @@ def test_supplier_details_shown(self, data_api_client):
         assert requirements_data.row(1).cell(1) == "02/02/2017"
         assert requirements_data.row(2).cell(0) == "Email address"
         assert requirements_data.row(2).cell(1) == "contact@big.com"
+
+    def test_visit_to_start_page_redirects_to_result_page(self, data_api_client):
```
idavidmcdonald: Your other option would be that instead of visiting the create page, you just mock the flash message and visit the result page and assert that the correct banner was shown. However, I think this will do for the moment

idavidmcdonald:
Woo, good work Tom 👍

tests/app/main/test_briefs.py
tombye: I think you're right about the flash message, it makes a nice distinction about the thing that's actually being tested which is more robust than the other approach.

I redid that last commit so it works like this now which should also cover the case of a visit to a question page for a completed response.

tombye:
I know this has been quite



...but thanks for all the feedback on it @idavidmcdonald

idavidmcdonald:
Nice, your change to that commit looks good too! Definitely good to go from my standpoint 👍

PR #594, opened 2017-01-24
Delete unused email template

TheDoubleK:
This is a relic from when the supplier app did password resets - it's all in the buyer app now.


pcraig3:
👍

PR #596, opened 2017-01-26
service update_section view: use `updated_service_name` dict key for flash message which makes more sense and is more consistent with remove_service…

risicle:
this has no real effect of course because the key is not consumed by the template - it's just a tiny sanity change

tiniest most insignificant change ever.


TheDoubleK:
:+1:

PR #597, opened 2017-01-26
Fix bug for previous page link when no nice to have requirements

idavidmcdonald:
We should skip the nice to have requirements evidence and page and go straight to the essentialRequirements page if the brief has no nice to have requirements.

Bug fix for https://www.pivotaltracker.com/story/show/137359355

### DIFF
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -226,6 +226,10 @@ def edit_brief_response(brief_id, brief_response_id, question_id=None):
                 return redirect(url_for('.view_response_result', brief_id=brief_id))
 
     previous_question_id = section.get_previous_question_id(question_id)
+    # Skip previous question if the brief has no nice to have requirements
+    if previous_question_id in brief.keys() and not brief[previous_question_id]:
+        previous_question_id = section.get_previous_question_id(previous_question_id)
+
     previous_question_url = None
     if previous_question_id:
         previous_question_url = url_for(
--- a/tests/app/main/test_briefs.py
+++ b/tests/app/main/test_briefs.py
@@ -527,6 +527,17 @@ class TestApplyToBrief(BaseApplicationTest):
         assert (doc.xpath("//a[text()='Back to previous page']/@href")[0] ==
                 '/suppliers/opportunities/1234/responses/5/availability')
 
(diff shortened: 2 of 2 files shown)


app/main/views/briefs.py
```diff
@@ -226,6 +226,10 @@ def redirect_to_next_page():
                 return redirect(url_for('.view_response_result', brief_id=brief_id))
 
     previous_question_id = section.get_previous_question_id(question_id)
+    # Skip previous question if the brief has no nice to have requirements
+    if previous_question_id in brief.keys() and not brief[previous_question_id]:
```
allait: I'm not sure I understand this condition. Why would `nicToHaveRequirements` be in the brief.keys if brief doesn't have any? Could we check for `if not brief.get(previous_question_id):` instead?

tests/app/main/test_briefs.py
```diff
@@ -527,6 +527,17 @@ def test_non_first_question_shows_previous_page_link(self):
         assert (doc.xpath("//a[text()='Back to previous page']/@href")[0] ==
                 '/suppliers/opportunities/1234/responses/5/availability')
 
+    def test_respond_to_email_previous_page_link_skips_nice_to_have_requirements_if_no_nice_to_haves_for_brief(self):
+        self.brief['briefs']['niceToHaveRequirements'] = []
```
allait: I have a suspicion that that empty list in the API might be a bug. The key really shouldn't exist if the buyer has never set any requirements.

idavidmcdonald: Have checked manually and it is currently being saved as an empty list when creating a brief. I'll add to the tech debt board

app/main/views/briefs.py
idavidmcdonald: Is this comment redundant due to the below mentioned API behaviour?

idavidmcdonald: Yep, have thought about it further and believe what currently stands is correct. Nice to have requirements is attributed to the brief and will come falsey if there are no nice to haves. Questions such as `availability` does not exist on the brief so would return `False` to `if not brief.get(previous_question_id)` and skip the question, which we do not want to do as they should not be skipped.

There might be more elegant ways to do this check (or the blunt way of hardcoding `niceToHaveRequirements` and deciding based on that) but I think I'll merge it as it stands.

PR #605, opened 2017-02-03
update hard coded reference to DOS framework

danielknell:
Backwards compatability as migration will soon land changing `framework.framework` from 'dos' to 'digital-outcomes-and-specialists'

### DIFF
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -417,7 +417,7 @@ def _render_not_eligible_for_brief_error_page(brief, clarification_question=Fals
             reason = data_reason_slug = "supplier-not-on-lot"
     else:
         reason = "supplier-not-on-framework"
-        data_reason_slug = "supplier-not-on-{}".format(brief['frameworkFramework'])
+        data_reason_slug = "supplier-not-on-{}".format(brief['frameworkSlug'])
 
     return render_template(
         "briefs/not_is_supplier_eligible_for_brief_error.html",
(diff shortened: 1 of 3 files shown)


app/main/views/briefs.py
```diff
@@ -417,7 +417,7 @@ def _render_not_eligible_for_brief_error_page(brief, clarification_question=Fals
             reason = data_reason_slug = "supplier-not-on-lot"
     else:
         reason = "supplier-not-on-framework"
-        data_reason_slug = "supplier-not-on-{}".format(brief['frameworkFramework'])
+        data_reason_slug = "supplier-not-on-{}".format(brief['frameworkSlug'])
```
TheDoubleK: Does this mean we also need to add a case to the `not_is_supplier_eligible_for_brief_error` template to match the case when this is DOS2? Or will it "just work"?

app/main/views/services.py
```diff
@@ -405,7 +405,9 @@ def edit_service_submission(framework_slug, lot_slug, service_id, section_id, qu
     """
     framework, lot = get_framework_and_lot(data_api_client, framework_slug, lot_slug, allowed_statuses=['open'])
 
-    force_return_to_summary = request.args.get('return_to_summary') or framework['framework'] == "dos"
+    force_return_to_summary = (request.args.get('return_to_summary') or
```
TheDoubleK: We need to fix this properly... but not right now.

app/main/views/briefs.py
idavidmcdonald: The template uses the `reason`, which remains unchanged, for logic.

The `data_reason_slug` has been changed but is just passed straight through and is then sent to GA as is. I have checked with Ash and this is completely fine.

TheDoubleK:
Good stuff :+1:

PR #607, opened 2017-02-07
Make it clear that this is the MIT licence

annashipman:
I was looking for a licence to send someone as an example and noticed that this one doesn't make it clear it's the MIT licence. This PR changes that, to make it easier for people to understand what licence is being used.

### DIFF
--- a/LICENCE
+++ b/LICENCE
@@ -1,3 +1,4 @@
+The MIT License (MIT)
 Copyright (C) 2015 Crown Copyright (Government Digital Service)
 
 Permission is hereby granted, free of charge, to any person obtaining a copy


annashipman:
As per, for example, https://github.com/alphagov/digitalmarketplace-aws/blob/master/LICENSE

allait:
Good call, thanks @annashipman!

We'll add them to some other ones that are missing. This also made me realise that between our repos we have 3 LICENSE files and 8 LICENCE files. We should probably choose one.

TheDoubleK:
I vote "LICENCE"!

allait:
Github search shows LICENSE is 20 times more popular though.

TheDoubleK:
Really?
¯\_(ツ)_/¯

annashipman:
LICENCE is the British spelling; LICENSE is the American spelling. So I also vote LICENCE.

...for the filename. However, for the licence itself, I'd use the exact wording, so `The MIT License` https://opensource.org/licenses/MIT

PR #609, opened 2017-02-08
[194] One question per page

samuelhwilliams:
One question per page flow for supplier services.

This changes the service specification flow for suppliers and moves towards a 'one question per page' standard. On the summary page, per-section editing is disabled and a link to add/edit answers is shown next to each question. As you work through the service spec, each question will follow on to the next question in the same section; when you reach the end of the section you will be returned to the service summary.

Included in this PR is a bugfix for multiquestions that will allow them to display any question type as children without individually specifying/checking the type each time.

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -175,7 +175,7 @@ def update_section(service_id, section_id):
 #  ####################  CREATING NEW DRAFT SERVICES ##########################
 
 
-@main.route('/frameworks/<framework_slug>/submissions/<lot_slug>/create', methods=['GET'])
+@main.route('/frameworks/<framework_slug>/submissions/<lot_slug>/create', methods=['GET', 'POST'])
 @login_required
 def start_new_draft_service(framework_slug, lot_slug):
     """Page to kick off creation of a new service."""
@@ -187,55 +187,51 @@ def start_new_draft_service(framework_slug, lot_slug):
     )
 
     section = content.get_section(content.get_next_editable_section_id())
+    if section is None:
+        section = content.get_section(content.get_next_edit_questions_section_id(None))
+        if section is None:
+            abort(404)
+
+        section = section.get_question_as_section(section.get_next_question_slug())
+
+    if request.method == 'POST':
+        update_data = section.get_data(request.form)
+
+        try:
+            draft_service = data_api_client.create_new_draft_service(
+                framework_slug, lot['slug'], current_user.supplier_id, update_data,
+                current_user.email_address, page_questions=section.get_field_names()
+            )['services']
+        except HTTPError as e:
+            update_data = section.unformat_data(update_data)
+            errors = section.get_error_messages(e.message)
+
+            return render_template(
+                "services/edit_submission_section.html",
+                framework=framework,
+                section=section,
+                service_data=update_data,
+                errors=errors
+            ), 400
+
+        return redirect(
+            url_for(
+                ".view_service_submission",
+                framework_slug=framework['slug'],
+                lot_slug=draft_service['lotSlug'],
+                service_id=draft_service['id'],
+            )
+        )
 
     return render_template(
         "services/edit_submission_section.html",
         framework=framework,
+        lot=lot,
+        one_service_limit=lot['oneServiceLimit'],
         service_data={},
-        section=section
+        section=section,
+        force_continue_button=True,
     ), 200
-
-
-@main.route('/frameworks/<framework_slug>/submissions/<lot_slug>/create', methods=['POST'])
-@login_required
-def create_new_draft_service(framework_slug, lot_slug):
-    """Hits up the data API to create a new draft service."""
-
-    framework, lot = get_framework_and_lot(data_api_client, framework_slug, lot_slug, allowed_statuses=['open'])
-
-    content = content_loader.get_manifest(framework_slug, 'edit_submission').filter(
-        {'lot': lot['slug']}
-    )
-
-    section = content.get_section(content.get_next_editable_section_id())
-
-    update_data = section.get_data(request.form)
-
-    try:
-        draft_service = data_api_client.create_new_draft_service(
-            framework_slug, lot['slug'], current_user.supplier_id, update_data,
-            current_user.email_address, page_questions=section.get_field_names()
-        )['services']
-    except HTTPError as e:
-        update_data = section.unformat_data(update_data)
-        errors = section.get_error_messages(e.message)
-
-        return render_template(
-            "services/edit_submission_section.html",
-            framework=framework,
-            section=section,
-            service_data=update_data,
-            errors=errors
-        ), 400
-
-    return redirect(
-        url_for(
-            ".view_service_submission",
-            framework_slug=framework['slug'],
-            lot_slug=draft_service['lotSlug'],
-            service_id=draft_service['id'],
-        )
-    )
 
 
 @main.route('/frameworks/<framework_slug>/submissions/<lot_slug>/<service_id>/copy', methods=['POST'])
@@ -379,6 +375,7 @@ def view_service_submission(framework_slug, lot_slug, service_id):
         "services/service_submission.html",
         framework=framework,
         lot=lot,
+        one_service_limit=lot['oneServiceLimit'],
         confirm_remove=request.args.get("confirm_remove", None),
         service_id=service_id,
         service_data=draft,
@@ -484,6 +481,7 @@ def edit_service_submission(framework_slug, lot_slug, service_id, section_id, qu
         "services/edit_submission_section.html",
         section=section,
         framework=framework,
+        lot=lot,
         next_question=next_question,
         service_data=service_data,
         service_id=service_id,
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -227,7 +227,6 @@ def start_new_draft_service(framework_slug, lot_slug):
         "services/edit_submission_section.html",
         framework=framework,
         lot=lot,
-        one_service_limit=lot['oneServiceLimit'],
         service_data={},
         section=section,
         force_continue_button=True,
@@ -375,7 +374,6 @@ def view_service_submission(framework_slug, lot_slug, service_id):
         "services/service_submission.html",
         framework=framework,
         lot=lot,
-        one_service_limit=lot['oneServiceLimit'],
         confirm_remove=request.args.get("confirm_remove", None),
         service_id=service_id,
         service_data=draft,
@@ -486,7 +484,6 @@ def edit_service_submission(framework_slug, lot_slug, service_id, section_id, qu
         service_data=service_data,
         service_id=service_id,
         force_return_to_summary=force_return_to_summary,
-        one_service_limit=lot['oneServiceLimit'],
         errors=errors,
     )
 
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -407,6 +407,7 @@ def edit_service_submission(framework_slug, lot_slug, service_id, section_id, qu
 
     force_return_to_summary = (request.args.get('return_to_summary') or
                                framework['framework'] == "digital-outcomes-and-specialists")
+    next_question = None
 
     try:
         draft = data_api_client.get_draft_service(service_id)['services']
@@ -422,6 +423,7 @@ def edit_service_submission(framework_slug, lot_slug, service_id, section_id, qu
     content = content_loader.get_manifest(framework_slug, 'edit_submission').filter(draft)
     section = content.get_section(section_id)
     if section and (question_slug is not None):
+        next_question = section.get_question_by_slug(section.get_next_question_slug(question_slug))
         section = section.get_question_as_section(question_slug)
 
     if section is None or not section.editable:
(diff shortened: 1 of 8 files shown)


TheDoubleK:
The create a new service seems broken - no question shown where you should be adding service name (e.g. `/suppliers/frameworks/g-cloud-9/submissions/cloud-hosting/create`):

screen shot 2017-02-09 at 16 49 04

samuelhwilliams:
Pushed an update to frameworks and updated this PR; update in a few minutes and /create should work again.

TheDoubleK:
Not a massive thing (although Ralph might disagree) but on the Service Name question the button saying "Save and return to service overview" is misleading as you've not yet ever seen the service overview page.

TheDoubleK:
Another weirdness - looks like the service question should all have a `depends_on: lot: ...` section.  `reselling.yml` is missing one, and for the Cloud Support lot is showing up in the overview page, not being regarded as required, even though it is not optional in the content:
screen shot 2017-02-10 at 10 56 45

And then on the question page itself it isn't rendering the sub-questions:
screen shot 2017-02-10 at 10 56 56

Even more weirdly, the question works just fine for Hosting and Software lots and the reseller sub-questions appear in their JSON schemas but not in the generated schema for Support :confused: 

So... changes for the content: I think it definitely needs a depends-on section, and also I'd change the `supplierType` key/question filename to be `resellerType` to reflect how tied it is to the reseller section/concept.

Also to note, the current JSON schemas in API master aren't quite up-to-date (the word count validations are still at 200 rather than 50), but as you're making more changes to the content now anyway there's no need to update that right away.

samuelhwilliams:
Re: save and return on /create, yeah, I noticed this and am thinking of how best to resolve it. The way I'm thinking now is pretty hacky.

Re: cloud-support, we haven't done any of the content for that yet so the process will not work and I haven't been testing it. This is 'known'.

app/templates/services/_base_service_page.html
```diff
@@ -59,9 +59,6 @@
       {% import "toolkit/summary-table.html" as summary %}
       {% for section in sections %}
         {{ summary.heading(section.name, id=section.slug) }}
-        {% if section.editable %}
```
TheDoubleK: Removing this section-level Edit link breaks all the old frameworks... (e.g. you can't edit any DOS2 services any more if you happen to have DOS2 open) it *might* be OK as they are closed for applications now anyway, but I'd be tempted to leave it in.  The trick then would be to get the `/create/` route to work OK with `edit_section: false`, i.e. get it to render as `/edit/service-name/service-name` rather than just `/edit/service-name` as it does at the moment.

TheDoubleK: e.g. it feels like the DOS-style per-lot questions we might want to show on a page together still:

## BEFORE
screen shot 2017-02-10 at 11 39 35

## NOW
screen shot 2017-02-10 at 11 39 26

app/templates/services/_base_service_page.html
```diff
@@ -78,30 +75,27 @@
...
+                {% if question.is_empty %}
+                  {{ summary.link('Answer question', url_for(".edit_service_submission", framework_slug=framework.slug, lot_slug=service_data.lot, service_id=service_id, section_id=section.id, question_slug=question.slug)) }}
                 {% else %}
-                  Not answered
+                  {{ summary.field("Not answered") }}
                 {% endif %}
-              {% endcall %}
-            {% elif question.is_empty %}
-              {% call summary.field() %}
-                <span class="summary-item-field-answer-required">{{ question.empty_message }}</span>
```
TheDoubleK: Are we not showing these empty messages at all any more?  They were introduced for DOS submissions, and I think Ralph was pretty strong on having them in the DOS summary tables.

It might be worth keeping this in, but ensure we don't show anything if the `empty_message` isn't set (it isn't for G-Cloud submissions).

TheDoubleK: At least, let's run it by @ralph-hawkins and see what he thinks.

app/templates/services/_base_service_page.html
```diff
@@ -78,30 +75,27 @@
...
-              {% endcall %}
+              {% endif %}
             {% else %}
               {{ summary[question.type](question.value, question.assurance) }}
             {% endif %}
-            {% if section.edit_questions and framework.status == 'open' %}
-              {% if not question.is_empty %}
-                {{ summary.remove_link('Remove', url_for(".remove_subsection", framework_slug=framework.slug, lot_slug=service_data.lot, service_id=service_id, section_id=section.id, question_slug=question.slug)) }}
```
TheDoubleK: Again, these "Remove" links are important for DOS-style frameworks, to add and remove specialist types, etc.

We don't want to break DOS submissions in making G-Cloud 9 work.

app/templates/services/_base_service_page.html
```diff
@@ -78,30 +75,27 @@
...
+            {% if framework.status == 'open' %}
+              {% if not question.is_empty and section.edit_questions %}
+                {{ summary.edit_link('Edit', url_for(".edit_service_submission", framework_slug=framework.slug, lot_slug=service_data.lot, service_id=service_id, section_id=section.id, question_slug=question.slug)) }}
               {% else %}
-                {{ summary.text("") }}
+                {% call summary.field(action=True) %}
+                  {{ 'Optional' if question.get('optional') else '' }}
```
TheDoubleK: This is another thing that looks odd for DOS... can @ralph-hawkins live with this?

screen shot 2017-02-10 at 12 17 58

app/templates/services/_base_service_page.html
TheDoubleK: I have a feeling we could make them contingent on `lot.one_service_limit` (it's something like that), which might mean we can have it both ways.

app/templates/services/edit_submission_section.html
```diff
@@ -35,36 +35,28 @@
 
 {% block save_button %}
 
-  {% if next_section_name and not force_return_to_summary %}
+  {% if not next_question or force_return_to_summary %}
     {%
       with
-      label="Save and continue",
-      name = "continue_to_next_section",
+      label="Save and return to service overview",
```
TheDoubleK: I think we should drop the word "service" from the button text, to make it more generic for whatever kind of thing you're adding - having the word "Service" on the button would be confusing during the flow for , e.g. labs or specialists.  (An alternative would be to pull the right noun from `lot.data.unitSingular`, but that feels like overkill to me.)

TheDoubleK:
Reviewed - left some comments.  I'm amazed that there aren't more tests that broke!

app/templates/services/_base_service_page.html
TheDoubleK: Tried and tested... `{% if one_service_limit %}` seems to work for these Remove links.

app/templates/services/edit_submission_section.html
samuelhwilliams: OK done.

app/templates/services/_base_service_page.html
samuelhwilliams: Sorted, thanks.

app/templates/services/_base_service_page.html
samuelhwilliams: Sorted, thanks.

app/templates/services/_base_service_page.html
samuelhwilliams: Sorted, thanks.

samuelhwilliams: Sorted.

app/templates/services/_base_service_page.html
samuelhwilliams: Sorted, thanks.

TheDoubleK:
I think the "Service name" section needs to be `editable:false` still.
screen shot 2017-02-13 at 09 56 39

samuelhwilliams:
Yep, that's in frameworks so doesn't affect this PR.

app/main/views/services.py
```diff
@@ -187,57 +187,56 @@ def start_new_draft_service(framework_slug, lot_slug):
     )
 
     section = content.get_section(content.get_next_editable_section_id())
+    if section is None:
+        section = content.get_section(content.get_next_section_id(None))
```
TheDoubleK: Ah, but I wonder if it will break the flow here if we make the service name section non-editable (to get rid of the extra link on the summary page).  I think it's worth getting the frameworks change in to see if it still works. (Or at least test the change locally.)

samuelhwilliams: I have tested it locally. It is fine.
screen shot 2017-02-13 at 10 38 56

app/main/views/services.py
```diff
@@ -187,57 +187,56 @@ def start_new_draft_service(framework_slug, lot_slug):
...
+                lot_slug=draft_service['lotSlug'],
+                service_id=draft_service['id'],
+            )
+        )
 
     return render_template(
         "services/edit_submission_section.html",
         framework=framework,
+        lot=lot,
+        one_service_limit=lot['oneServiceLimit'],
```
TheDoubleK: If the `lot` object is being passed to the template then I don't know if we also need the `one_service_limit` as a separate parameter - it's probably fine for the template to do `lot.one_service_limit`.

app/main/views/services.py
```diff
@@ -379,6 +378,7 @@ def view_service_submission(framework_slug, lot_slug, service_id):
         "services/service_submission.html",
         framework=framework,
         lot=lot,
+        one_service_limit=lot['oneServiceLimit'],
```
TheDoubleK: Same here - if the template has the `lot` anyway this seems superfluous.

app/main/views/services.py
samuelhwilliams: Thanks, removed.

app/main/views/services.py
TheDoubleK: Nice.

app/templates/services/edit_submission_section.html
```diff
@@ -35,32 +35,26 @@
...
     {% endwith %}
 
-    <p class="next-page-message">
-      Next: {{ next_section_name }}
-    </p>
-
-    {%
-    with
-      type = "secondary",
-      label = "Save and return to service overview"
```
TheDoubleK: We're definitely losing these grey buttons?  I still liked them. Ah well.

TheDoubleK:
OK I think this should be good to go once conflicts are resolved.

app/templates/services/edit_submission_section.html
samuelhwilliams: I'll double-check with @ralph-hawkins; they aren't on the prototype.

app/templates/services/_base_service_page.html
```diff
@@ -78,30 +78,40 @@
...
-          {% call summary.row(complete=not question.answer_required) %}
+          {% call summary.row() %}
             {{ summary.field_name(question.label) }}
-            {% if question.answer_required %}
-              {% call summary.field() %}
-                {% if framework and framework.status == 'open' %}
-                    <a href="{{ url_for(".edit_service_submission", framework_slug=framework.slug, lot_slug=service_data.lot, service_id=service_id, section_id=section.id, _anchor=question.id) }}">Answer required</a>
+            {% if question.is_empty %}
+              {% if framework and framework.status == 'open' and section.edit_questions %}
+                {% if lot['oneServiceLimit'] %}
```
allait: It doesn't matter either way, but we tend to use dot notation in templates (ie `lot.oneServiceLimit`), since it's equivalent in Jinja.

samuelhwilliams: Ah OK, thanks.

app/templates/services/edit_submission_section.html
TheDoubleK: Chatted to Cath - we are introducing these grey "Save and return" buttons for the declaration signing flow, and we think they should stay for the service flow here too.

app/main/views/services.py
```diff
@@ -187,57 +187,55 @@ def start_new_draft_service(framework_slug, lot_slug):
     )
 
     section = content.get_section(content.get_next_editable_section_id())
+    if section is None:
+        section = content.get_section(content.get_next_section_id(None))
+        while (section and not section.edit_questions):
```
samuelhwilliams: Have added a new flag to content-loader as per this.

app/templates/services/edit_submission_section.html
samuelhwilliams: Going to leave this as-is and will review with @ralph-hawkins and @cathrooney when it's on preview.

requirements.txt
```diff
@@ -5,7 +5,7 @@ werkzeug==0.10.4
 python-dateutil==2.4.2
 
 git+https://github.com/alphagov/digitalmarketplace-utils.git@24.0.4#egg=digitalmarketplace-utils==24.0.4
-git+https://github.com/alphagov/digitalmarketplace-content-loader.git@2.5.0#egg=digitalmarketplace-content-loader==2.5.0
+git+https://github.com/alphagov/digitalmarketplace-content-loader.git@2.6.0#egg=digitalmarketplace-content-loader==2.8.1
```
TheDoubleK: This needs to be 2.9.0 now.

TheDoubleK: Also, there are two numbers to update here.

TheDoubleK:
Al looks good, tested locally and all works as far as I can see :+1:

PR #611, opened 2017-02-13
138323285-declaration_reuse_view

benvand:


benvand:
Ugly commits, doing away with this branch

PR #612, opened 2017-02-14
138323285-declaration_reuse_view

benvand:
https://www.pivotaltracker.com/story/show/138323285

* Helper function for framework ordering
* Helper function to get a reusable declaration given a supplier
* New view reuse_framework_supplier_declaration to determine whether to ask if we should reuse
* New view reuse_framework_supplier_declaration_post to set prefill preference based on form
* New template app/templates/frameworks/reuse_declaration.html for reuse page
    (as per http://dm-prototype.herokuapp.com/structure/declaration-start-page/declaration?thisPage=declaration-reuse)
* Change start declaration 'next page' to new reuse page
* Bump apiclient dependency version
* Tests

Requires: https://github.com/alphagov/digitalmarketplace-api/pull/539

### DIFF
--- a/app/main/helpers/frameworks.py
+++ b/app/main/helpers/frameworks.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 import re
+from datetime import datetime
 
+from dmutils.formats import DATETIME_FORMAT
 from flask import abort
 from flask_login import current_user
 from dmapiclient import APIError
@@ -36,6 +38,18 @@ def get_framework_lot(framework, lot_slug):
         return next(lot for lot in framework['lots'] if lot['slug'] == lot_slug)
     except StopIteration:
         abort(404)
+
+
+def order_frameworks_for_reuse(frameworks):
+    """Sort frameworks by reuse suitability.
+
+    If a declaration has the reuse flag set and is the most recently closed framework then that's our framework.
+    """
+    return sorted(
+        filter(lambda i: i['allow_declaration_reuse'] and i['application_close_date'], frameworks),
+        key=lambda i: datetime.strptime(i['application_close_date'], DATETIME_FORMAT),
+        reverse=True
+    )
 
 
 def register_interest_in_framework(client, framework_slug):
@@ -79,6 +93,24 @@ def get_declaration_status(data_api_client, framework_slug):
         return 'unstarted'
     else:
         return declaration.get('status', 'unstarted')
+
+
+def get_reusable_declaration(supplier_id, client, exclude_framework_slugs=None):
+    """Given a list of declarations find the most suitable for reuse.
+
+     :param supplier_id: supplier whose declarations we are inspecting
+     :param client: data client if not frameworks
+     :param exclude_framework_slugs: list of framework slugs to exclude from results
+     :return: (declaration, framework)
+     """
+    exclude_framework_slugs = exclude_framework_slugs or []
+    declarations = {i['frameworkSlug']: i for i in client.find_supplier_declarations(supplier_id)['frameworkInterest']}
+    frameworks = client.find_frameworks()['frameworks']
+
+    for framework in order_frameworks_for_reuse(frameworks):
+        if framework['slug'] in declarations and framework['slug'] not in exclude_framework_slugs:
+            return framework, declarations[framework['slug']]
+    return None, None
 
 
 def get_supplier_framework_info(data_api_client, framework_slug):
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -3,6 +3,7 @@ from collections import OrderedDict
 from itertools import chain
 
 from dateutil.parser import parse as date_parse
+from dmapiclient import HTTPError
 from flask import render_template, request, abort, flash, redirect, url_for, current_app, session
 from flask_login import current_user
 import flask_featureflags as feature
@@ -30,7 +31,7 @@ from ..helpers.frameworks import (
     get_supplier_on_framework_from_info, get_declaration_status_from_info, get_supplier_framework_info,
     get_framework, get_framework_and_lot, count_drafts_by_lot, get_statuses_for_lot,
     return_supplier_framework_info_if_on_framework_or_abort, returned_agreement_email_recipients,
-    check_agreement_is_related_to_supplier_framework_or_abort
+    check_agreement_is_related_to_supplier_framework_or_abort, get_reusable_declaration
 )
 from ..helpers.validation import get_validator
 from ..helpers.services import (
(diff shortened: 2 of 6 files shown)


app/templates/frameworks/reuse_declaration.html
```diff
@@ -0,0 +1,73 @@
...
+      <form method="POST" action="{{ url_for(".reuse_framework_supplier_declaration_post", framework_slug=current_framework.slug, old_framework_slug=old_framework.slug) }}">
+        <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
+        {%
+          with
+          name = "reuse",
+          question = "Do you want to reuse the answers from your earlier declaration?",
+          inline = true,
+          type = "radio",
+          options = [
+            {"label": "Yes", "value": old_framework.name},
```
TheDoubleK: This is wrong - it should be the slug, not the name that gets passed in.  As it is the API is being hit with a name rather than a slug and throwing a 400 error.

app/main/views/frameworks.py
```diff
@@ -273,6 +273,89 @@ def framework_start_supplier_declaration(framework_slug):
                            framework_close_date=framework_close_date), 200
 
 
+@main.route('/frameworks/<framework_slug>/declaration/fdo', methods=['GET'])
```
TheDoubleK: I don't think we should have this temporary route at all - there's no way this should ever be released with this in. The POST that sets the reuse flag should, for now, redirect into the main supplier declaration signing flow.  From testing perspective it's easy enough to poke the API to make sure the right value has been set.

app/main/views/frameworks.py
```diff
@@ -273,6 +273,89 @@ def framework_start_supplier_declaration(framework_slug):
...
+@login_required
+def fdo_placeholder(framework_slug):
+    # TODO remove this method entirely and its template.
+    reuse = data_api_client.get_supplier_framework_info(
+        current_user.supplier_id, framework_slug
+    )['prefillDeclarationFromFrameworkSlug']
```
TheDoubleK: This dereference breaks - the API response looks like:
```
{u'frameworkInterest': {u'agreementId': None, u'agreementReturnedAt': None, u'agreementPath': None, u'supplierId': 700008, u'prefillDeclarationFromFrameworkSlug': u'digital-outcomes-and-specialists-2', u'countersignedPath': None, u'frameworkSlug': u'g-cloud-9', u'countersignedDetails': None, u'countersignedAt': None, u'agreementStatus': None, u'onFramework': None, u'agreementReturned': False, u'agreedVariations': {}, u'declaration': {}, u'countersigned': False, u'supplierName': u"Kev's Pies", u'agreementDetails': None}}
```
So it would need to be ['frameworkInterest']['prefillDeclarationFromFrameworkSlug']

TheDoubleK:
I'd much rather this new page got merged as part of a fully working flow, without any placeholder page or redirect.  We should only merge things that we believe could be released through to production as-is.

app/main/helpers/frameworks.py
```diff
@@ -38,6 +40,18 @@ def get_framework_lot(framework, lot_slug):
...
 
 
+def order_frameworks_for_reuse(frameworks):
+    """Sort frameworks by reuse suitability.
+
+    If a declaration has the reuse flag set and is the most recently closed framework then that's our framework.
+    """
+    return sorted(
+        filter(lambda i: i['allow_declaration_reuse'] and i['application_close_date'], frameworks),
+        key=lambda i: datetime.strptime(i['application_close_date'], DATETIME_FORMAT),
```
TheDoubleK: Why do we need to string format the date here?  Can't we just order by the datetime values themselves?

app/main/helpers/frameworks.py
```diff
@@ -81,6 +95,23 @@ def get_declaration_status(data_api_client, framework_slug):
         return declaration.get('status', 'unstarted')
 
 
+def get_reusable_declaration(declarations, client=None, exclude_framework_slugs=None):
+    """Given a list of declarations attempt to find one suitable for reuse.
+
+    :param declarations: list of supplier's previous declarations
+    :param frameworks: list viable frameworks
```
TheDoubleK: This docstring doesn't match the params for the function.

app/main/views/frameworks.py
```diff
@@ -273,6 +273,73 @@ def framework_start_supplier_declaration(framework_slug):
...
+        # Experimental feature to be used to force through reuse.
+        try:
+            # Get their old declaration.
+            declaration = data_api_client.get_supplier_declaration(supplier_id, reusable_declaration_framework_slug)
+            old_framework = data_api_client.get_framework(reusable_declaration_framework_slug)['frameworks']
+        except APIError:
+            abort(404)
+    else:
+        # Otherwise then attempt to determine if they have a declaration we can offer for reuse.
+        declarations = data_api_client.find_supplier_declarations(supplier_id)['frameworkInterest']
```
TheDoubleK: I'd be tempted to push this API call to get `declarations` into the `get_reusable_declaration` method itself - the client is passed in there anyway.  It feels a bit weird to get the list of declarations here and then have the logic to turn the list into a dict in the method - that would be nicer done all together.

app/main/views/frameworks.py
```diff
@@ -273,6 +273,73 @@ def framework_start_supplier_declaration(framework_slug):
...
+        old_framework, declaration = get_reusable_declaration(
+            declarations,
+            client=data_api_client,
+            exclude_framework_slugs=[framework_slug]
+        )
+        if not declaration:
+            # If not then set the prefill to none and redirect to the overview.
```
TheDoubleK: The default value is NULL, so I don't think there should be any need to make the API call here, just redirect.
Also, it feels weird to set data from a GET endpoint.

app/main/views/frameworks.py
```diff
@@ -273,6 +273,73 @@ def framework_start_supplier_declaration(framework_slug):
...
+    :query_param: reusable_declaration_framework_slug
+    :return: 404, redirect or reuse page (frameworks/reuse_declaration.html)
+    """
+    reusable_declaration_framework_slug = request.args.get('reusable_declaration_framework_slug', None)
+    supplier_id = current_user.supplier_id
+    # Get the current framework.
+    current_framework = data_api_client.get_framework(framework_slug)['frameworks']
+    if reusable_declaration_framework_slug:
+        # If they have supplied a declaration then force it. This is an 'admin' feature.
+        # Experimental feature to be used to force through reuse.
```
TheDoubleK: Shouldn't be "Experimental" - that implies it might not work :)

I'd go with something like 
```
# If a framework slug is supplied in the URL then use this for pre-filling if possible,
# overriding the default framework selection logic.
```

app/main/views/frameworks.py
```diff
@@ -273,6 +273,73 @@ def framework_start_supplier_declaration(framework_slug):
...
+        current_framework=current_framework,
+        old_framework=old_framework,
+        old_framework_application_close_date=date_parse(old_framework['application_close_date']).strftime('%B, %Y'),
+    ), 200
+
+
+@main.route('/frameworks/<framework_slug>/declaration/reuse', methods=['POST'])
+@login_required
+def reuse_framework_supplier_declaration_post(framework_slug):
+    try:
```
TheDoubleK: What condition is this `try... except` designed to catch?  Multiple slugs being posted?  Why would that happen?

app/main/views/frameworks.py
```diff
@@ -273,6 +273,73 @@ def framework_start_supplier_declaration(framework_slug):
...
+        old_framework_slug = bool(int(request.form.get('reuse', 0))) or None
+    except ValueError:
+        old_framework_slug = request.form.get('reuse')
+
+    supplier_id = current_user.supplier_id
```
TheDoubleK: I think we should add a check here that the supplier has a completed declaration for the specified framework and 404 if the declaration doesn't exist.  Would be weird to end up setting a value that says to reuse a declaration that we don't have.

TheDoubleK: Although if that's enforced by the API maybe we can do without it here?

tests/app/main/helpers/test_frameworks.py
```diff
@@ -364,3 +369,148 @@ def test_check_agreement_is_related_to_supplier_framework_or_abort_does_not_abor
...
     check_agreement_is_related_to_supplier_framework_or_abort(agreement, supplier_framework)
+
+
+def test_order_frameworks_for_reuse():
+    """Test happy path. Should return 2 frameworks, closest date first."""
+    t09 = '2009-03-03T01:01:01.000000Z'
+    t07 = '2007-03-03T01:01:01.000000Z'
+    t12 = '2012-03-03T01:01:01.000000Z'
+
+    fake_frameworks = [
```
TheDoubleK: These are already ordered in reverse chronological order as they get passed in - might be better to have them mixed up a bit to start with.

tests/app/main/helpers/test_frameworks.py
```diff
@@ -364,3 +369,148 @@ def test_check_agreement_is_related_to_supplier_framework_or_abort_does_not_abor
...
+        {'allow_declaration_reuse': True, 'application_close_date': t12, 'extraneous_field': 'foo'},
+        {'allow_declaration_reuse': False, 'application_close_date': t09, 'extraneous_field': 'foo'},
+        {'allow_declaration_reuse': True, 'application_close_date': t07, 'extraneous_field': 'foo'},
+    ]
+    ordered = order_frameworks_for_reuse(fake_frameworks)
+
+    assert len(ordered) == 2
```
TheDoubleK: It feels like the first test already covers this, and does more because it also checks that the "right" two frameworks are returned.

tests/app/main/helpers/test_frameworks.py
```diff
@@ -364,3 +369,148 @@ def test_check_agreement_is_related_to_supplier_framework_or_abort_does_not_abor
...
+        {'allow_declaration_reuse': True, 'application_close_date': t07, 'extraneous_field': 'foo'},
+        {'allow_declaration_reuse': True, 'application_close_date': t13, 'extraneous_field': 'foo'},
+        {'allow_declaration_reuse': True, 'application_close_date': t09, 'extraneous_field': 'foo'},
+        {'allow_declaration_reuse': True, 'application_close_date': t14, 'extraneous_field': 'foo'},
+        {'allow_declaration_reuse': True, 'application_close_date': t12, 'extraneous_field': 'foo'},
+        {'allow_declaration_reuse': True, 'application_close_date': t11, 'extraneous_field': 'foo'},
+    ]
+    ordered = order_frameworks_for_reuse(fake_frameworks)
+
+    assert ordered[0] == {'allow_declaration_reuse': True, 'application_close_date': t14, 'extraneous_field': 'foo'}
```
TheDoubleK: No harm in testing for the whole list here?

tests/app/main/test_frameworks.py
```diff
@@ -4124,3 +4125,211 @@ def test_error_if_box_not_ticked(self, data_api_client):
...
+        # Assert expected api calls.
+        data_api_client.get_supplier_declaration.assert_called_once_with(1234, 'g-cloud-8')
+
+    def test_404_when_specified_declaration_not_found(self, data_api_client):
+        """Fail on a 404 if declaration is specified but not found."""
+
+        # Modify the data client.
+        framework = {}
+        data_api_client.get_framework.return_value = {'frameworks': framework}
+        data_api_client.get_supplier_declaration.side_effect = APIError()
```
TheDoubleK: This could be `APIError(mock.Mock(status_code=404))` as I think 404 is what the API should return if it's working and doesn't find a declaration for that framework.  Raw `APIError()` on its own gives a 503.

tests/app/main/test_frameworks.py
```diff
@@ -4124,3 +4125,211 @@ def test_error_if_box_not_ticked(self, data_api_client):
...
+        # Do the get.
+        resp = self.client.get(
+            '/suppliers/frameworks/g-cloud-9/declaration/reuse?reusable_declaration_framework_slug=g-cloud-8'
+        )
+
+        # Assert the 404
+        assert resp.status_code == 404
+
+        # Assert expected api calls.
+        data_api_client.get_framework.assert_called_once_with('g-cloud-9')
```
TheDoubleK: Should this assertion also go into the "Expected API calls" section of the previous test?

app/main/views/frameworks.py
benvand: The post either sends the framework slug or `'0'` if the string can be cast to an int then `old_framework_slug` is None.

app/main/views/frameworks.py
benvand: > it feels weird to set data from a GET endpoint.
Agreed. I was told to explicitly set None if we _don't_ want to reuse but at this point we can rely on the default.

app/main/helpers/frameworks.py
benvand: Good catch

app/main/views/frameworks.py
benvand: Will change that

app/main/helpers/frameworks.py
benvand: That's parsing it back into a datetime so we can do the ordering

TheDoubleK: :man_facepalming:  sorry!

tests/app/main/helpers/test_frameworks.py
benvand: Yeah it's pretty verbose. I'll change this to separate assertions in the same test. I wanted to write a test for the filtering condition that indicates _what_ the function should be doing. A tests as documentation type thing.

TheDoubleK:
:+1: let's merge once tests pass.

PR #614, opened 2017-02-15
Use frameworkFramework when rendering service name

Wynndow:
For [this chore on Pivotal](https://www.pivotaltracker.com/story/show/139269583).

In the list view of services, if a service is a DOS service we don't
provide a link to edit that service. This was achieved by comparing the
services `frameworkName` to the string
`digital-outcomes-and-specialists`. This wouldn't work for DOS2.

We now compare using the recently update `frameworkFramework` which will
be `digital-outcomes-and-specialists` for all DOS services, regardless
of the framework iteration.

### DIFF
--- a/tests/app/main/test_services.py
+++ b/tests/app/main/test_services.py
@@ -155,7 +155,8 @@ class TestListServices(BaseApplicationTest):
                     'lotName': 'Special Lot Name',
                     'status': 'published',
                     'id': '123',
-                    'frameworkSlug': 'digital-outcomes-and-specialists'
+                    'frameworkSlug': 'digital-outcomes-and-specialists-2',
+                    'frameworkFramework': 'digital-outcomes-and-specialists',
                 }]
             }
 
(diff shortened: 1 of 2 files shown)


tests/app/main/test_services.py
```diff
@@ -155,7 +155,8 @@ def test_shows_dos_service_name_without_edit_link(self, data_api_client):
                     'lotName': 'Special Lot Name',
                     'status': 'published',
                     'id': '123',
-                    'frameworkSlug': 'digital-outcomes-and-specialists'
+                    'frameworkSlug': 'digital-outcomes-and-specialists',
```
idavidmcdonald: If it is a one line change I might advocate that `frameworkSlug` should be for DOS2. That way we know that our code is using `frameworkFramework` instead or the same valued `frameworkSlug` at the moment.

Or maybe I'm being anal....

👍 either way

Wynndow: Nope, I think that's pretty valid. i've updated. will merge when green.

PR #617, opened 2017-02-16
138702235-save-and-continue-and-save-button-on-declr

benvand:
All on the declarations page:
* Save and continue
* See what is coming Next
* Save and return to the declaration overview.
* Remove 'make decl' button from last page
* Hook up the breadcrumb back to the overview page

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -443,11 +443,13 @@ def framework_supplier_declaration(framework_slug, section_id=None):
                     framework_slug=framework_slug,
                     section_id=content.get_next_editable_section_id()))
 
+    # Get and check the current section.
     section = content.get_section(section_id)
     if section is None or not section.editable:
         abort(404)
+    # Do the same for the next section. This also implies whether or not we are on the last page of the declaration.
+    next_section = content.get_section(content.get_next_section_id(section_id=section.id, only_editable=True))
 
-    is_last_page = section_id == content.sections[-1]['id']
     saved_answers = {}
 
     try:
@@ -457,7 +459,6 @@ def framework_supplier_declaration(framework_slug, section_id=None):
     except APIError as e:
         if e.status_code != 404:
             abort(e.status_code)
-
     if request.method == 'GET':
         errors = {}
         all_answers = saved_answers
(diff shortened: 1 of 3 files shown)


app/main/views/frameworks.py
```diff
@@ -329,31 +330,36 @@ def framework_supplier_declaration(framework_slug, section_id=None):
...
-                    return redirect(
-                        url_for('.framework_dashboard',
-                                framework_slug=framework['slug']))
+                    # Otherwise they may have completed their declaration. Check.
+                    if not next_section:
+                        # They have. Write the flash message.
+                        url = "{}/declaration_complete".format(
+                            url_for('.framework_dashboard', framework_slug=framework['slug'])
+                        )
+                        flash(url, 'declaration_complete')
```
risicle: I don't think we're doing this here - this flash message is to accompany the final "Make declaration" button-press, done from the overview.

I think you can safely obliterate all machinery for setting flash messages from this view.

app/templates/frameworks/edit_declaration_section.html
```diff
@@ -59,29 +60,41 @@
...
+              {% endwith %}
+
             {% else %}
-              {%
-                with
-                label="Save and continue",
-                type="save"
+              {# If there is no next page the 'save and return' button is the primary and only save button #}
+              {% with
+                type="save",
```
risicle: Does jinja allow `... if ... else ...` statements here? Because save<->secondary seems to be the only difference between these two `toolkit/button.html` uses...

benvand: Nope :(

risicle:
Have you managed to get away without breaking any tests here? If so that's a little on the worrying side.

app/main/views/frameworks.py
```diff
@@ -329,31 +330,28 @@ def framework_supplier_declaration(framework_slug, section_id=None):
...
     return render_template(
         "frameworks/edit_declaration_section.html",
         framework=framework,
+        next_section=next_section,
         section=section,
         declaration_answers=all_answers,
-        is_last_page=is_last_page,
         get_question=content.get_question,
-        errors=errors,
+        errors=errors
```
TheDoubleK: @risicle would like the trailing comma left in here :)

app/templates/frameworks/edit_declaration_section.html
```diff
@@ -19,7 +19,8 @@
         "label": "Apply to " + framework.name
       },
       {
-        "label": "Supplier declaration"
+        "link": url_for(".framework_supplier_declaration_overview", framework_slug=framework.slug),
+        "label": "Declaration overview"
```
TheDoubleK: The link text for this breadcrumb in the prototype is "Your declaration"

app/templates/frameworks/edit_declaration_section.html
TheDoubleK: You can do things like `type = "secondary" if next_section else "save"`

TheDoubleK:
Couple of comments but looking good.  Let's get it rebased up and un x-fail the tests.

app/main/views/frameworks.py
benvand: I think they harm readability. For instance. How long does it take to work out what this does?
`x = lambda: 'foo',`
any small gain like reordering arguments and/ or prettier diffs is negated by the inconsistency presented by statements like:
```
x= 'foo', 
len(x)
```
In python commas can _do_ something. For instance one is a string, one is a tuple:
```
('foo')
('foo',)
```
Unless we're asking the comma to _do_ something then it being there has the potential to mislead. 

Why is it there if it's not doing anything? Would be my argument in a nutshell.

We also don't write:
`x=[foo, bar, baz,]`

TheDoubleK:
Looks good to go :+1:

app/main/views/frameworks.py
risicle:
```
x= 'foo', 
len(x)
```

I'm not advocating using trailing commas in places that sequences aren't being specified. And this would be an argument against commas at the ends of lines at all, which we obviously can't do without. There's nothing "trailing"-comma specific about it.

```
('foo')
('foo',)
```

This is exactly a case I would use to argue in *favour* of trailing commas. Firstly, if you want to make a single-element tuple you *have* to use the second form, no way around it, the trailing comma is needed. But then, if you add a second member, the trailing comma is not needed. Remove trailing comma? Leave it? Surely the most consistent thing is to just be able to append another member to the line.

Otherwise you end up with pieces of code like

```
a = ("one",)
b = ("two", "one")
```

which is jarring, even though `b` is just a prepend-ation (?) of `a`.

```
c = ("three", "four")
```

Decide that `c` can lose its first member, selecting and deleting the `"three",` and all of a sudden, bam:

```
c = ("four")
```

You've not got a tuple anymore.

```
x=[foo, bar, baz,]
```

Yes, that's just a quirk of the language - I probably *would* add trailing commas here if the language allowed me (it doesn't).

PR #620, opened 2017-02-17
NO-TICKET-New-tests-base-2

benvand:
* New base class BaseApplicationTestLoggedIn with auto login

### DIFF
--- a/tests/app/helpers.py
+++ b/tests/app/helpers.py
@@ -394,6 +394,16 @@ class BaseApplicationTest(object):
             assert not session.get("_flashes")
 
 
+class BaseApplicationTestLoggedIn(BaseApplicationTest):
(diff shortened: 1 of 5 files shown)


TheDoubleK:
There are still *lots* of tests left doing `self.login()` - have you left these as they are because those test classes have some tests logged in and some not logged in?

I see how this new auto-logged-in class is a nice thing, but I'm not sure if having a mix of multiple different ways we log in tests is a good thing.  Might be nicer to stick to the same pattern everywhere? Also, having a `self.login()` in the test code makes it pretty clear what's going on for that test - no need to hunt round figuring out which base class the current test happens to have inherited from.

benvand:
Yeah I realised that some were going to conflict with upcoming commits. I'll update this PR another time and we can reopen discussion then.

PR #621, opened 2017-02-20
fixing breadcrumbs and links

danielknell:


idavidmcdonald:
Failing tests, looks like just due to the API mocks... :(

PR #623, opened 2017-02-21
NO-TICKET-repair-post-bug

benvand:
* Accept 'no-reuse' instead of 0 as negative for reuse
* POST email_address instead of user object
* Update tests

See issue here for more details:
https://github.com/alphagov/digitalmarketplace-frontend-toolkit/issues/306

### DIFF
--- a/app/main/helpers/frameworks.py
+++ b/app/main/helpers/frameworks.py
@@ -95,22 +95,26 @@ def get_declaration_status(data_api_client, framework_slug):
         return declaration.get('status', 'unstarted')
 
 
-def get_reusable_declaration(supplier_id, client, exclude_framework_slugs=None):
+def get_framework_for_reuse(supplier_id, client, exclude_framework_slugs=None):
     """Given a list of declarations find the most suitable for reuse.
 
      :param supplier_id: supplier whose declarations we are inspecting
      :param client: data client if not frameworks
      :param exclude_framework_slugs: list of framework slugs to exclude from results
-     :return: (declaration, framework)
+     :return: framework
      """
     exclude_framework_slugs = exclude_framework_slugs or []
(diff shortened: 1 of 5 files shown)


TheDoubleK:
Looks good - nice catch that we don't need to pass the framework in the URL too.
:+1: once tests pass.

benvand:
@TheDoubleK 
Also realised we were dealing with `supplier_framework` objects everywhere so made that clear and added a check to see that the declaration is complete before we allow reuse.
Seemed like the right thing to do.

app/main/helpers/frameworks.py
```diff
@@ -104,13 +111,17 @@ def get_reusable_declaration(supplier_id, client, exclude_framework_slugs=None):
      :return: (declaration, framework)
      """
     exclude_framework_slugs = exclude_framework_slugs or []
-    declarations = {i['frameworkSlug']: i for i in client.find_supplier_declarations(supplier_id)['frameworkInterest']}
+    supplier_frameworks = client.find_supplier_declarations(supplier_id)['frameworkInterest']
+    supplier_frameworks_with_completed_declarations = filter(
```
benvand: We are now :|

benvand: Fortunately `get_supplier_framework_info` returns the declaration with the framework info so I swapped out `get_supplier_declaration` for `get_supplier_framework_info`

app/main/helpers/frameworks.py
```diff
@@ -104,13 +104,17 @@ def get_reusable_declaration(supplier_id, client, exclude_framework_slugs=None):
      :return: (declaration, framework)
```
TheDoubleK: Docstring needs to be updated as only returning the framework now.

TheDoubleK:
Sorry, should have spotted this the first time round :/
Merge once Travis is happy :+1:

TheDoubleK:
Going to merge this and fix the validation on the reuse page later.

PR #624, opened 2017-02-21
Fix bug with POST to 'responses/start' if brief response exists

idavidmcdonald:
Reviewers, please read the pivotal story with comments first - https://www.pivotaltracker.com/story/show/139468419

POST route now handles three cases
- no existing brief response
- existing draft brief response
- existing submitted brief response

If you click the 'Start application' button and you have an ongoing application that has not been submitted you should be redirected to the application flow at the start for your existing application which is prefilled (similar to the continue application button).

If you click the 'Start application' button and you have already applied and completed an application you should be redirected to your response page with a error message saying 'you have already applied' or to that effect, again the same as the continue application button.

I wanted to bring in `assert_not_called` from mock 1.3.0 for tests but decided we might as well upgrade all the way to 2.0.0. (http://mock.readthedocs.io/en/latest/changelog.html)

### DIFF
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -91,21 +91,30 @@ def start_brief_response(brief_id):
     if not is_supplier_eligible_for_brief(data_api_client, current_user.supplier_id, brief):
         return _render_not_eligible_for_brief_error_page(brief)
 
-    if request.method == 'POST':
-        brief_response = data_api_client.create_brief_response(
-            brief_id,
-            current_user.supplier_id,
-            {},
-            current_user.email_address,
-        )['briefResponses']
-        brief_response_id = brief_response['id']
-        return redirect(url_for('.edit_brief_response', brief_id=brief_id, brief_response_id=brief_response_id))
-
     brief_response = data_api_client.find_brief_responses(
         brief_id=brief_id,
         supplier_id=current_user.supplier_id,
         status='draft,submitted'
     )['briefResponses']
+
+    if request.method == 'POST':
+        if brief_response:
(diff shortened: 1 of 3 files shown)


app/main/views/briefs.py
```diff
@@ -91,22 +91,31 @@ def start_brief_response(brief_id):
...
     )['briefResponses']
 
+    if request.method == 'POST':
+        if brief_response:
+            if brief_response[0].get('status') == 'submitted':
+                flash('already_applied', 'error')
+                return redirect(url_for(".view_response_result", brief_id=brief_id))
+            if brief_response[0].get('status') == 'draft':
+                return redirect(
+                    url_for('.edit_brief_response', brief_id=brief_id, brief_response_id=brief_response[0]['id'])
```
Wynndow: Why not use `.get()` here, like when pulling out the status of the brief response a few lines above?

app/main/views/briefs.py
```diff
@@ -91,22 +91,31 @@ def start_brief_response(brief_id):
...
+                )
+        else:
+            brief_response = data_api_client.create_brief_response(
+                brief_id,
+                current_user.supplier_id,
+                {},
+                current_user.email_address,
+            )['briefResponses']
+            brief_response_id = brief_response['id']
```
Wynndow: Same as above

idavidmcdonald:
Thanks @Wynndow, decided to remove the unecessary `get`s. I also tidied the code up a bit to get rid of some duplication.

Should be no problems but do have a glance...

PR #625, opened 2017-02-22
Update dependencies for followup questions

samuelhwilliams:
This PR:
* Updates the content-loader to 3.1.1, which enables support for followup questions/questions with dependencies.
* Updates the dm-frontend-toolkit to 21.0.0.
* Updates dm-frameworks to 6.2.0
* Pulls in show-hide-content.js from dm-frontend rather than govuk-frontend.
* xfail test_display while declaration is changing; too many changes and too complex a test.

### DIFF
--- a/app/templates/briefs/_brief_response_data.html
+++ b/app/templates/briefs/_brief_response_data.html
@@ -25,20 +25,24 @@
                 {% endcall %}
               {% endfor %}
             {% elif item.type == "dynamic_list" %}
-              {% set processed_followups = [] %}
+              {% set followup_questions_already_displayed = [] %}
               {% for question in item.questions %}
                   {% if question.type == "boolean" %}
                     {% if question._data['followup'] is defined %}
                       {% for followup_question in question._data['followup'] %}
                         {% call summary.row() %}
-                          {% set processed_followups = processed_followups.extend([followup_question]) %}
+                          {# Keep track of the questions we are displaying here, so we know to skip them as we
+                              continue to loop through item.questions #}
+                          {% set followup_questions_already_displayed = followup_questions_already_displayed.extend([followup_question]) %}
+
                           {{ summary.field_name(question.question) }}
                           {{ summary.text(item.unformat_data(brief_response)[followup_question]) }}
                         {% endcall %}
                       {% endfor %}
                     {% endif %}
                   {% else %}
-                    {% if question.id not in processed_followups %}
+                    {# If we have seen this question before, let's not display it again. #}
+                    {% if question.id not in followup_questions_already_displayed %}
                       {% call summary.row() %}
                         {{ summary.field_name(question.question) }}
                         {{ summary.text(item.unformat_data(brief_response)[question.id]) }}
--- a/app/templates/briefs/_brief_response_data.html
+++ b/app/templates/briefs/_brief_response_data.html
@@ -25,17 +25,20 @@
                 {% endcall %}
               {% endfor %}
             {% elif item.type == "dynamic_list" %}
+              {% set processed_followups = [] %}
               {% for question in item.questions %}
                   {% if question.type == "boolean" %}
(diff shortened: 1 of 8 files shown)


samuelhwilliams:
@allait Updating to include the followup/dependency questions which triggers quite a few changes here, if you could cast your eye over that would be good.
@idavidmcdonald A few changes to briefs tests and one template here, in case you want to run your eye over them.

app/templates/briefs/_brief_response_data.html
```diff
@@ -27,17 +27,19 @@
             {% elif item.type == "dynamic_list" %}
               {% for question in item.questions %}
                   {% if question.type == "boolean" %}
-                    {% if question._data['followup'] %}
+                    {% if question._data['values_followup'] %}
```
allait: `question.values_followup` is a `@property`, so it doesn't exist in `_data`, I think we can still use `followup` for this check, although I'm not sure why we're getting it from `._data` as opposed to the question itself, @idavidmcdonald?

app/templates/briefs/_brief_response_data.html
```diff
@@ -27,17 +27,19 @@
...
                       {% call summary.row() %}
                         {{ summary.field_name(question.question) }}
-                        {{ summary.text(item.unformat_data(brief_response)[question._data['followup']]) }}
+                        {{ summary.text(item.unformat_data(brief_response)[question._data['values_followup']]) }}
```
allait: I haven't realised followup was used here, but this change won't work. Old followup format stored a single string value that matched to follow-up question ID. Now both `.followup` and `.values_followup` are dictionaries, so we either need to either change this part to deal with multiple follow-ups or to make it always take the first one for now.

tests/app/main/test_briefs.py
```diff
@@ -938,8 +938,8 @@ def test_existing_nice_to_have_requirements_evidence_prefills_existing_data(self
 
         # Check yesno radio buttons
         for i in (0, 2):
-            assert len(doc.xpath("//*[@id='input-yesNo-" + str(i) + "-yes' and @checked]")) == 1
-        assert len(doc.xpath("//*[@id='input-yesNo-1-no' and @checked]")) == 1
+            assert len(doc.xpath("//*[@id='input-yesNo-" + str(i) + "-1' and @checked]")) == 1
```
allait: Yeah, sorry about that. I was merging the radio and boolean question templates. I knew the IDs will change, but haven't realised we might be using them in tests.

app/templates/briefs/_brief_response_data.html
samuelhwilliams: I spent quite a lot of changes going over this and changing it to values_followup is what made it work, actually. We can look it over on Monday if you want.

app/templates/briefs/_brief_response_data.html
samuelhwilliams: I'll look into this some more tomorrow morning; thanks.

tests/app/main/test_frameworks.py
```diff
@@ -2272,6 +2272,7 @@ def _extract_section_information(doc, section_title):
             )
         )
 
+    @pytest.mark.xfail
```
idavidmcdonald: Sorry, why are we expecting this test to fail?

samuelhwilliams: Because the supplier declaration has changed from the frameworks repo update in this. I have another PR ready to pull in 6.3.1, and at the same time I will fix this text. However, the test isn't an easy one to follow and I want to spend a little more time on fixing this.

samuelhwilliams: A bit more clarification: the frameworks repo version this PR pulls in (6.2.0) is missing name keys for the questions that this test checks, so the test can't really be modified to work.

PR #627, opened 2017-02-22
FIX: Copying draft services

samuelhwilliams:
This PR:
* Fixes a bug with copying services; one-question-per-page services could not generate the correct URL to redirect to as the logic was missing.


TheDoubleK:
Nice catch.

Might be nice to have a test around this if it's not going to take days to set up...

samuelhwilliams:
Added a test for this.

PR #628, opened 2017-02-22
NO-TICKET-AR-AQ As per request from RH

benvand:
* Change declaration overview page to state 'Answer question' instead of 'Answer required' when an answer still needs to be provided.


benvand:
Here's some I prepared earlier :boat: :smile: 
Before:
overview-orig


After:
overview-aq
overview1

TheDoubleK:
Yep, it's better now, isn't it :)

PR #629, opened 2017-02-22
NO-TICKET-implement-validation

benvand:
* New form `ReuseDeclarationForm`
* Have view use form
* Have template accept form and display validation

### DIFF
--- a/app/main/forms/frameworks.py
+++ b/app/main/forms/frameworks.py
@@ -1,6 +1,6 @@
 from flask.ext.wtf import Form
-from wtforms import BooleanField
-from wtforms.validators import DataRequired, Length
+from wtforms import BooleanField, HiddenField
+from wtforms.validators import DataRequired, InputRequired, Length
 from dmutils.forms import StripWhitespaceStringField
 
 
@@ -34,3 +34,17 @@ class AcceptAgreementVariationForm(Form):
             DataRequired(message="You can only save and continue if you agree to the proposed changes.")
         ]
     )
+
+
+class ReuseDeclarationForm(Form):
+    """Form for the reuse declaration page. One yes no question.
+
+    `reuse` is a yes no whether they want to reuse a framework.
+    `old_framework` is a hidden field allowing us to pass back the framework slug of the framework they are choosing to
+    reuse.
+    """
+    reuse = BooleanField(
+        'Do you want to reuse the answers from your earlier declaration?',
+        validators=[InputRequired(message='You must answer this question.')]
+    )
+    old_framework_slug = HiddenField()
--- a/app/templates/frameworks/reuse_declaration.html
+++ b/app/templates/frameworks/reuse_declaration.html
@@ -28,7 +28,17 @@
 {% endblock %}
 
 {% block main_content %}
-  <div class="grid-row framework-dashboard">
+
+  {% if form_errors %}
+    {%
+      with
+      errors = form_errors
+    %}
+      {% include 'toolkit/forms/validation.html' %}
+    {% endwith %}
+  {% endif %}
+
+  <div class="grid-row">
     <div class="column-two-thirds">
       {% with
          heading = self.page_title_inner(),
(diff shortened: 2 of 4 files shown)


app/templates/frameworks/reuse_declaration.html
```diff
@@ -28,38 +28,45 @@
...
-  <div class="grid-row framework-dashboard">
+  <div class="grid-row">
     <div class="column-two-thirds">
       {% with
          heading = self.page_title_inner(),
          smaller = True
       %}
         {% include "toolkit/page-heading.html" %}
       {% endwith %}
+
```
TheDoubleK: The general pattern we use for validation is that if there are multiple questions on the page then you get this banner with a list of the things with problems.

But for just one question on a page we don't have the banner, just the red bar and message next to the question itself.

(Grab Ralph or Tom if they have a second to confirm, but I'm pretty sure this is right.)

TheDoubleK:
We need new words for the error message - show this to Roz/Andy and they should be able to come up with something.

screen shot 2017-02-23 at 11 14 57

app/main/forms/frameworks.py
```diff
@@ -34,3 +34,17 @@ class AcceptAgreementVariationForm(Form):
...
+class ReuseDeclarationForm(Form):
+    """Form for the reuse declaration page. One yes no question.
+
+    `reuse` is a yes no whether they want to reuse a framework.
+    `old_framework` is a hidden field allowing us to pass back the framework slug of the framework they are choosing to
+    reuse.
+    """
+    reuse = BooleanField(
+        'Do you want to reuse the answers from your earlier declaration?',
+        validators=[InputRequired()]
```
TheDoubleK: I think you can just do `validators=[InputRequired(message="Please answer this")]` to get a different message shown.

But get better words than that :)

tests/app/main/test_frameworks.py
```diff
@@ -2272,6 +2272,7 @@ def _extract_section_information(doc, section_title):
             )
         )
 
+    @pytest.mark.xfail
```
TheDoubleK: Would be nice to figure out how to keep these tests happy if possible.

app/main/forms/frameworks.py
benvand: So I spoke to @laurenceberry about this and he said use the default (although he may not have know what the default was) I'd rather not let this block the PR given I've already asked for confirmation on this.

app/templates/frameworks/reuse_declaration.html
benvand: Yeah confirming with them now

app/main/forms/frameworks.py
benvand: I have asked for new content though so if they get back to me i'll change it :)

tests/app/main/test_frameworks.py
benvand: I'll do my best :smirk:

app/main/forms/frameworks.py
benvand: @andymcaleer said leave it and it might have to be looked at as part of a wider change

TheDoubleK:
There is absolutely no way we can ship it with "This field is required" as a message.

The default isn't even a default we've set - it's a default from WTForms I think.

app/main/forms/frameworks.py
andymcaleer: For future reference, that default message could be something like ‘You must answer this question.' which is less technical and more user-friendly. But changing the default is a bigger piece of work, I guess, so I'm not suggesting we make this change now.

andymcaleer:
In that case (Kev's recent comment), go with ‘You must answer this question.'

benvand:
Ah sorry guys, done this, just waiting on the error placement

TheDoubleK:
I'm 99.9% sure we don't want the top validation banner in, so go with that.

benvand:
@TheDoubleK @laurenceberry 
Above the H1. Thats how we do it on the sign up/sign in pages. (havent checked other 1 q per page pages)

TheDoubleK:
Those sign in pages don't get the next-to-the question validation messages though.  Log in is different from filling-in-the-stuff (although I don't exactly know why).

TheDoubleK:
:+1:

PR #630, opened 2017-02-23
Prefill declaration sections

TheDoubleK:
Here we have working code that pretty much completes:
https://www.pivotaltracker.com/story/show/138702601 (Pre-filling relevant declaration sections with answers from a previous framework)
https://www.pivotaltracker.com/story/show/138703089 (Adding blue banners and validation-style blue messages to pre-filled pages)

~~**NOTE**~~ 
~~!!!!!!!!! There are no new unit tests at all for this new behaviour :scream:  !!!!!!!!!~~

~~Feel free to pick this up and add some relevant tests.~~

~~Or review it and merge it in if you feel like seeing it all working on Friday, and on my honour I will add the required unit tests on Monday.~~

**UPDATE 26/2: Tests added and rebased up-to-date - now 100% ready for review.**

Here's a screenshot:
screen shot 2017-02-26 at 12 13 36

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -514,11 +514,8 @@ def framework_supplier_declaration(framework_slug, section_id=None):
         if len(errors) > 0:
             status_code = 400
         else:
-            validator = get_validator(framework, content, all_answers)
-            if validator.get_error_messages():
+            if not all_answers.get("status"):
                 all_answers.update({"status": "started"})
-            else:
-                all_answers.update({"status": "complete"})
             try:
                 data_api_client.set_supplier_declaration(
                     current_user.supplier_id,
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -471,21 +471,38 @@ def framework_supplier_declaration(framework_slug, section_id=None):
     # Do the same for the next section. This also implies whether or not we are on the last page of the declaration.
     next_section = content.get_section(content.get_next_section_id(section_id=section.id, only_editable=True))
 
-    saved_answers = {}
+    supplier_framework = data_api_client.get_supplier_framework_info(
+        current_user.supplier_id, framework_slug)['frameworkInterest']
+    saved_declaration = supplier_framework.get('declaration', {})
+    section_has_been_prefilled = False
 
-    try:
-        response = data_api_client.get_supplier_declaration(current_user.supplier_id, framework_slug)
-        if response['declaration']:
-            saved_answers = response['declaration']
-    except APIError as e:
-        if e.status_code != 404:
-            abort(e.status_code)
     if request.method == 'GET':
         errors = {}
-        all_answers = saved_answers
+
+        section_errors = get_validator(
+            framework,
+            content,
+            saved_declaration,
+        ).get_error_messages_for_page(section)
+
+        # If there are section_errors it means that this section has not previously been completed
+        if section_errors and section.prefill and supplier_framework['prefillDeclarationFromFrameworkSlug']:
+            # Fetch the old declaration to pre-fill from and pass it through
+            try:
+                declaration_to_reuse = data_api_client.get_supplier_declaration(
+                    current_user.supplier_id,
+                    supplier_framework['prefillDeclarationFromFrameworkSlug']
+                )['declaration']
+                all_answers = declaration_to_reuse
+                section_has_been_prefilled = True
+            except APIError as e:
+                if e.status_code != 404:
+                    abort(e.status_code)
+        else:
+            all_answers = saved_declaration
     else:
         submitted_answers = section.get_data(request.form)
-        all_answers = dict(saved_answers, **submitted_answers)
+        all_answers = dict(saved_declaration, **submitted_answers)
 
         validator = get_validator(framework, content, submitted_answers)
         errors = validator.get_error_messages_for_page(section)
@@ -525,6 +542,7 @@ def framework_supplier_declaration(framework_slug, section_id=None):
         framework=framework,
         next_section=next_section,
         section=section,
+        section_has_been_prefilled=section_has_been_prefilled,
         declaration_answers=all_answers,
         get_question=content.get_question,
         errors=errors
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -488,6 +488,8 @@ def framework_supplier_declaration(framework_slug, section_id=None):
         # If there are section_errors it means that this section has not previously been completed
         if section_errors and section.prefill and supplier_framework['prefillDeclarationFromFrameworkSlug']:
             # Fetch the old declaration to pre-fill from and pass it through
+            # For now we pre-fill a whole section or none of the section
+            # TODO: In future we may need to pre-fill individual questions and add a 'prefilled' flag to the questions
             try:
                 prefill_from_slug = supplier_framework['prefillDeclarationFromFrameworkSlug']
                 framework_to_reuse = data_api_client.get_framework(prefill_from_slug)['frameworks']
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -474,7 +474,7 @@ def framework_supplier_declaration(framework_slug, section_id=None):
     supplier_framework = data_api_client.get_supplier_framework_info(
         current_user.supplier_id, framework_slug)['frameworkInterest']
     saved_declaration = supplier_framework.get('declaration', {})
-    section_has_been_prefilled_from = ""
+    name_of_framework_that_section_has_been_prefilled_from = ""
 
     if request.method == 'GET':
         errors = {}
@@ -498,7 +498,7 @@ def framework_supplier_declaration(framework_slug, section_id=None):
                     prefill_from_slug
                 )['declaration']
                 all_answers = declaration_to_reuse
-                section_has_been_prefilled_from = framework_to_reuse['name']
+                name_of_framework_that_section_has_been_prefilled_from = framework_to_reuse['name']
             except APIError as e:
                 if e.status_code != 404:
                     abort(e.status_code)
@@ -546,7 +546,7 @@ def framework_supplier_declaration(framework_slug, section_id=None):
         framework=framework,
         next_section=next_section,
         section=section,
-        section_has_been_prefilled_from=section_has_been_prefilled_from,
+        name_of_framework_that_section_has_been_prefilled_from=name_of_framework_that_section_has_been_prefilled_from,
         declaration_answers=all_answers,
         get_question=content.get_question,
         errors=errors
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -474,7 +474,7 @@ def framework_supplier_declaration(framework_slug, section_id=None):
     supplier_framework = data_api_client.get_supplier_framework_info(
         current_user.supplier_id, framework_slug)['frameworkInterest']
     saved_declaration = supplier_framework.get('declaration', {})
-    section_has_been_prefilled = False
+    section_has_been_prefilled_from = ""
 
     if request.method == 'GET':
         errors = {}
@@ -489,12 +489,14 @@ def framework_supplier_declaration(framework_slug, section_id=None):
         if section_errors and section.prefill and supplier_framework['prefillDeclarationFromFrameworkSlug']:
             # Fetch the old declaration to pre-fill from and pass it through
             try:
+                prefill_from_slug = supplier_framework['prefillDeclarationFromFrameworkSlug']
+                framework_to_reuse = data_api_client.get_framework(prefill_from_slug)['frameworks']
                 declaration_to_reuse = data_api_client.get_supplier_declaration(
                     current_user.supplier_id,
-                    supplier_framework['prefillDeclarationFromFrameworkSlug']
+                    prefill_from_slug
                 )['declaration']
                 all_answers = declaration_to_reuse
-                section_has_been_prefilled = True
+                section_has_been_prefilled_from = framework_to_reuse['name']
             except APIError as e:
                 if e.status_code != 404:
                     abort(e.status_code)
@@ -542,7 +544,7 @@ def framework_supplier_declaration(framework_slug, section_id=None):
         framework=framework,
         next_section=next_section,
         section=section,
-        section_has_been_prefilled=section_has_been_prefilled,
+        section_has_been_prefilled_from=section_has_been_prefilled_from,
         declaration_answers=all_answers,
         get_question=content.get_question,
         errors=errors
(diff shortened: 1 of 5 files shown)


tests/app/main/test_frameworks.py
```diff
@@ -2805,14 +2810,188 @@ def test_get_with_with_previous_answers(self, data_api_client):
...
+                self.framework(slug='digital-outcomes-and-specialists-2', name='Digital Stuff 2', status='live'),
+            ]
+            # Current framework application information
+            data_api_client.get_supplier_framework_info.return_value = self.supplier_framework(
+                framework_slug="g-cloud-9",
+                declaration={"status": "started"},
+                prefill_declaration_from_framework_slug="digital-outcomes-and-specialists-2"
+            )
+            # The previous declaration to prefill from
+            data_api_client.get_supplier_declaration.return_value = {
```
risicle: We don't assert the call args for `get_supplier_declaration` anywhere here do we?

I added some useful tools at the top of this file `_assert_args_and_raise` and `_assert_args_and_return` which assert the args & kwargs in-the-loop ( only useful for cases where mocked functions should get called only once per test).

tests/app/main/test_frameworks.py
```diff
@@ -2805,14 +2810,188 @@ def test_get_with_with_previous_answers(self, data_api_client):
...
+
+            # Radio buttons have been pre-filled with the correct answers
+            assert len(doc.xpath('//input[@id="input-conspiracy-yes"]/@checked')) == 1
+            assert len(doc.xpath('//input[@id="input-corruptionBribery-no"]/@checked')) == 1
+            assert len(doc.xpath('//input[@id="input-fraudAndTheft-yes"]/@checked')) == 1
+            assert len(doc.xpath('//input[@id="input-terrorism-no"]/@checked')) == 1
+            assert len(doc.xpath('//input[@id="input-organisedCrime-no"]/@checked')) == 1
+
+            # Blue banner message is shown at top of page
+            assert doc.xpath('normalize-space(//div[@class="banner-information-without-action"]/p/text())') == \
```
risicle: `normalize-space(string(//div[@class="banner-information-without-action"]))` would be more robust here but meh

tests/app/main/test_frameworks.py
```diff
@@ -2805,14 +2810,188 @@ def test_get_with_with_previous_answers(self, data_api_client):
...
+            assert len(doc.xpath('//input[@id="input-organisedCrime-no"]/@checked')) == 1
+
+            # No blue banner message is shown at top of page
+            assert len(doc.xpath('//div[@class="banner-information-without-action"]')) == 0
+
+            # No blue information messages are shown next to each question
+            info_messages = doc.xpath('//div[@class="message-wrapper"]//span[@class="message-content"]')
+            assert len(info_messages) == 0
+
+    def test_answers_not_prefilled_if_section_marked_as_prefill_false(self, data_api_client):
```
risicle: Do we test a case where we're trying to prefill from a framework that is missing some of the desired questions?

app/templates/frameworks/edit_declaration_section.html
```diff
@@ -52,11 +59,19 @@
...
+                {{ forms[question.type](
+                    question,
+                    declaration_answers,
+                    errors if errors and (errors[question.id] or question.type == 'multiquestion') else {},
+                    question_number=question.number,
+                    get_question=get_question,
+                    message="This answer is from your {} declaration".format(section_has_been_prefilled_from)
+                            if section_has_been_prefilled_from and
+                                (declaration_answers.get(question.id) is not none or question.type == 'multiquestion')
+                            else None
```
benvand: It would probably look even cleaner without the ternary tbf, especially in Jinja syntax

risicle:
(hold on maybe I should actually try running it before I pass judgement...)

app/templates/frameworks/edit_declaration_section.html
risicle: /me would probably do

```
message=(
        "This answer is from your {} declaration".format(section_has_been_prefilled_from)
    if
        section_has_been_prefilled_from and (declaration_answers.get(question.id) is not none or question.type == 'multiquestion')
    else
        None
)
```

though probably not to everyone's tastes.

tests/app/main/test_frameworks.py
benvand: I'm going to suggest leaving these functions alone for now. They're in the wrong place at the moment.
Happy to discuss these methods' inclusion and location another time.

I'd also say we don't strictly need to assert the calls in this test:
This test asserts the existence of elements and their state on the page. What this test asserts can be achieved entirely independently of the api calls.

If you want, on the other hand, to assert that a given piece of code makes certain calls then check the calls but not the html.

A caveat to this might be if you wanted to ensure that the state on the page can be achieved only with a certain number of api calls, however that's not what _this_ test is meant to do.


I'm not _that_ fussed if we check the calls here or not but I'd question wether or not it's causing fragility in the test suite by enforcing conditions we actually don't care about.

tests/app/helpers.py
```diff
@@ -309,7 +309,7 @@ def framework(
                      'unitSingular': 'service', 'unitPlural': 'service'},
                 ]
             metaframework = "g-cloud"
-        elif slug == 'digital-outcomes-and-specialists':
+        elif 'digital-outcomes-and-specialists' in slug:
```
benvand: Fo sho

tests/app/main/test_frameworks.py
risicle: I would say we care about *all* these conditions *all* the time. "Separate tests for separate concerns" *sounds* nice, but whichever way you look at it, doesn't end up testing as much. Only if this exact function-under-test call was made with the exact same setup prologue would it be not worth making these extra assertions here. And if that were the case, then I would advocate merging those tests to improve run time.

This comes back to the coverage argument. We're kidding ourselves with any coverage reports we're doing if we're not actually asserting all that we could be in all cases.

e.g. from what I can tell, nowhere else in the tests do we test the case where the function under test is called with the declaration being `started`, not having any pre-existing answers and having the `prefill_declaration_from_framework_slug` set. We'd be missing out on the opportunity to ensure that combination doesn't go crazy and idunno call `get_supplier_declaration` for `g-cloud-9` in this case. And we do *care* that it doesn't do that as if it doesn't call it with the right argument, there's no way it can possibly give us the right answer under real life conditions...

app/templates/frameworks/edit_declaration_section.html
TheDoubleK: Not sure how to avoid the ternary without duplicating the whole thing in an `if...else` block (as it used to be), so going to go with @risicle's suggestion which makes it at least a bit easier to read I think.

app/templates/macros/toolkit_forms.html
```diff
@@ -1,9 +1,9 @@
-{% macro text(question_content, service_data, errors, question_number=None, get_question=None) -%}
```
TheDoubleK: The problem is that because the macros are called dynamically they have to all accept the same arguments, so without using `kwargs` in order to add a new argument to be used in one macro we have to add it to the argument list of all macros.  So this seems the exact kind of thing that `kwargs` is designed for?

TheDoubleK: What is the standard python way of doing things?  And if we're in a Jinja template does it still apply?  I only just found out how kwargs works in Jinja so may well be abusing it here through not knowing any better.  But it seemed like a neater solution than adding unused arguments to all macros.

app/main/views/frameworks.py
```diff
@@ -471,21 +471,42 @@ def framework_supplier_declaration(framework_slug, section_id=None):
...
+            # TODO: In future we may need to pre-fill individual questions and add a 'prefilled' flag to the questions
+            try:
+                prefill_from_slug = supplier_framework['prefillDeclarationFromFrameworkSlug']
+                framework_to_reuse = data_api_client.get_framework(prefill_from_slug)['frameworks']
+                declaration_to_reuse = data_api_client.get_supplier_declaration(
+                    current_user.supplier_id,
+                    prefill_from_slug
+                )['declaration']
+                all_answers = declaration_to_reuse
+                section_has_been_prefilled_from = framework_to_reuse['name']
```
TheDoubleK: Yep - will fix.

tests/app/helpers.py
TheDoubleK: It's all fine, nothing to see here :trollface:

tests/app/main/test_frameworks.py
benvand: What you're describing above isn't unit testing. 
A test with more condition checks may cover more code but they are inherently more fragile and add far less value by being less indicative of an overall problem.

Put it this way, we could check, in every one of our unit tests, whether the GOV.UK cookie banner appears.
Say we then add a check to every test to make sure the CrownCopyright appears in the bottom right.

Then, on the next test run, everything breaks. 
Was it the banner or the CC or both?

We don't do the above because it doesn't add value, every test would just break if we changed the banner.
Unless you're checking one specific thing unit test failures help _narrow down_ a problem at best and at worst point you completely in the wrong direction.

Obviously this is a massive oversimplification but it demonstrates why it's important to check only what you want to find out. Tests are cheap, once you've written one it's very easy to write more so if needs be then split them out into several tests that functionally do the same thing but check different conditions.

Then when a test breaks called `test_view_makes_only_one_api_call` fails you know _exactly_ what's wrong instead of a test called `test_declaration_page` failing because someone updated the frontend toolkit.

app/main/views/frameworks.py
TheDoubleK: I'm going with `name_of_framework_that_section_has_been_prefilled_from` - or is that way too long?

tests/app/main/test_frameworks.py
```diff
@@ -2805,14 +2810,188 @@ def test_get_with_with_previous_answers(self, data_api_client):
             doc = html.fromstring(res.get_data(as_text=True))
             assert len(doc.xpath('//input[@id="input-PR1-no"]/@checked')) == 1
 
-    def test_post_valid_data(self, data_api_client):
+    def test_get_with_with_prefilled_answers(self, data_api_client):
         with self.app.test_client():
             self.login()
+            # First call is for the current framework; second call for the framework to pre-fill from
+            data_api_client.get_framework.side_effect = [
```
TheDoubleK: Fair point.  Fixing up.

tests/app/main/test_frameworks.py
TheDoubleK: No but we can do.

tests/app/main/test_frameworks.py
TheDoubleK: It feels to me like testing _all_ the API calls made by a view could fall into the category of "I know too much about the inside of the function under test" kind of testing.

TheDoubleK:
Comments addressed:
* Changed variable name to `name_of_framework_that_section_has_been_prefilled_from` to make it clearer that it is a framework name being passed round and not some other sort of object.
* Fixed tests to not depend on order of calls to `data_api_client.get_framework`
* Add test for case where previous declaration doesn't have answers for all questions in a section so page is only partially pre-filled.
* Use more robust xPath expression for retrieving banner message
* Assert call args to `data_api_client.get_supplier_declaration` in cases where it should be called.

app/templates/macros/toolkit_forms.html
risicle: Yeeeeeeah.... in python, a function defined with only positional parameters would only accept those parameters, you would have to explicitly add a `**kwargs` parameter to allow that, so the syntax diverges a bit here - still, if there's a reason for it, go right ahead. I was just making a comment in case you were doing it just for the hell of it.

app/templates/frameworks/edit_declaration_section.html
```diff
@@ -52,11 +59,22 @@
...
+                {{ forms[question.type](
+                    question,
+                    declaration_answers,
+                    errors if errors and (errors[question.id] or question.type == 'multiquestion') else {},
+                    question_number=question.number,
+                    get_question=get_question,
+                    message=(
+                            "This answer is from your {} declaration".format(name_of_framework_that_section_has_been_prefilled_from)
+                        if
+                            name_of_framework_that_section_has_been_prefilled_from and (declaration_answers.get(question.id) is not none or question.type == 'multiquestion')
```
risicle: Lowercase `none` here - mistake or jinja thing?

galund: I think this highlights a problem - these macros are basically causing us to write Python code in our templates. The benefit from separation of concerns that templating is supposed to bring is not being realised right now, is the point. These definitely aren't things that could be owned by front-end designers/specialists, in their current state, IMO.

risicle: Additionally, `or question.type == 'multiquestion'` here seems to cause the "this answer is from your..." message to appear even on questions that have a missing previous entry as long as they're multiquestions.
20170227missingprefillmultiquestions

risicle: (though maybe this isn't important if we're only catering for all-prefilled-or-none-prefilled cases)

TheDoubleK: Lowercase none is a jinja thing.

Yep the multiquestion thing will need an extra condition in there, good spot, will fix. (At the moment it is all or nothing though.)

George, philosophical questions aside that's a much bigger issue than this pull request can address right now.

tests/app/main/test_frameworks.py
risicle: I don't tend to get too bound up by terms, what's "unit testing" or not "unit testing" - I'm just concerned with what's useful testing (hey maybe I should register _that_ term). When a test fails for me, the first thing I look at isn't the test name (take a survey of test names we have - more often that not they aren't particularly illuminating), I run it under `--pdb` and look at the assertion that failed, which hopefully is written in a way that makes it clear what it's trying to assert for and why.

Yes, testing for the govuk banner is going too far, but this is where you have to apply some pragma. It's also something that should have been tested as part of its own module. Code such as a view should be able to consider something as fundamental as the banner to be a black box that is tested on its own and "works".

And while "tests are cheap", maintaining them certainly is not. How may people when they fix up a particular small test go and similarly fix up the 57 other variants of the test sitting next to it? Copy pasted test variants have a tendency to diverge... 

Test execution time is also something that isn't particularly cheap.

@TheDoubleK I think only when the situation falls into the category of "leave the function some leeway of how it wants to achieve this, but if it's "there's no possible way for it to give the right answer if it doesn't call the right thing", I'd say not. If we want to go mad, I'd say the right thing to do would be for the `get_supplier_declaration` mock to be able to return appropriate mock declarations for either framework depending on how it's called and for the assertions to be able to distinguish between them. But that would be insane to set up manually all the time, which is one reason I'd like to move to a more "comprehensive" api-mocking system.

app/templates/macros/toolkit_forms.html
TheDoubleK: Yep this is just how Jinja does kwargs I'm afraid - you can't pass them in explicitly.  Pretty horrible design choice, I agree :/

risicle:
Anyway, given the changes, I'm sure this will do for G9. :+1:

risicle:
Before you merge - should this PR also take care of removing the ability for the "final page" to mark the declaration as `complete`?

PR #632, opened 2017-02-27
NO-TICKET-normalise-tests-app-helpers

benvand:
* Optimise & order imports
* Double to single quotes
* Normalize function definitions/ signatures
* Normalize variable names


galund:
So personally I think changes like just changing quote marks are a bad thing, because they make following the history of each line significantly harder. Optimising things is okay (preferably if it actually seems to need optimising) and better variable names I have no problem with at all :)

PR #633, opened 2017-02-28
Restore test_display test and pull in latest frameworks

samuelhwilliams:
* Update frameworks to 6.3.1 (pulls in names for some questions in supplier declaration, and some other content tweaks).
* Restores a test which was temporarily xfailed.


samuelhwilliams:
Yes, good catch - I've removed that stub.

idavidmcdonald:
Cool, all good then 👍

PR #634, opened 2017-02-28
140124261-confidence-banner-on-framework-overview

benvand:
* New banner on framework overview page
  * https://preview.development.digitalmarketplace.service.gov.uk/suppliers/frameworks/g-cloud-9
* Tests

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -2,6 +2,7 @@
 from collections import OrderedDict
 from itertools import chain
 
+from datetime import datetime
 from dateutil.parser import parse as date_parse
 from dmapiclient import HTTPError
 from flask import render_template, request, abort, flash, redirect, url_for, current_app, session
@@ -22,7 +23,6 @@ from dmutils.documents import (
     degenerate_document_path_and_return_doc_name, get_signed_url, get_extension, file_is_less_than_5mb,
     file_is_empty, file_is_image, file_is_pdf, sanitise_supplier_name
 )
-
 from ... import data_api_client, flask_featureflags
 from ...main import main, content_loader
 from ..helpers import hash_email, login_required
@@ -68,20 +68,22 @@ def framework_dashboard(framework_slug):
             )
 
     drafts, complete_drafts = get_drafts(data_api_client, framework_slug)
-
     supplier_framework_info = get_supplier_framework_info(data_api_client, framework_slug)
     declaration_status = get_declaration_status_from_info(supplier_framework_info)
-    supplier_is_on_framework = get_supplier_on_framework_from_info(supplier_framework_info)
-
     # Do not show a framework dashboard for earlier G-Cloud iterations
     if declaration_status == 'unstarted' and framework['status'] == 'live':
         abort(404)
 
-    application_made = supplier_is_on_framework or (len(complete_drafts) > 0 and declaration_status == 'complete')
+    supplier_is_on_framework = get_supplier_on_framework_from_info(supplier_framework_info)
+    application_made = complete_drafts and declaration_status == 'complete'
     lots_with_completed_drafts = [lot for lot in framework['lots'] if count_drafts_by_lot(complete_drafts, lot['slug'])]
 
     framework_dates = content_loader.get_message(framework_slug, 'dates')
     framework_urls = content_loader.get_message(framework_slug, 'urls')
+
+    # Confidence banner.
+    if application_made and framework['status'] == 'open':
+        flash('submission_confidence', 'success')
 
     # filenames
     result_letter_filename = RESULT_LETTER_FILENAME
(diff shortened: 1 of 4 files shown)


app/main/views/frameworks.py
```diff
@@ -68,21 +68,23 @@ def framework_dashboard(framework_slug):
...
     declaration_status = get_declaration_status_from_info(supplier_framework_info)
-    supplier_is_on_framework = get_supplier_on_framework_from_info(supplier_framework_info)
-
     # Do not show a framework dashboard for earlier G-Cloud iterations
     if declaration_status == 'unstarted' and framework['status'] == 'live':
         abort(404)
 
-    application_made = supplier_is_on_framework or (len(complete_drafts) > 0 and declaration_status == 'complete')
+    supplier_is_on_framework = get_supplier_on_framework_from_info(supplier_framework_info)
+    application_made = complete_drafts and declaration_status == 'complete'
```
risicle: I'd be slightly careful of changing the meaning of `application_made` here, it means we've got the variable meaning one thing for the entirety of the view, but then a slightly different thing in the template. If the supplier is "on framework" they will have had to make an application, and there should be no danger in them being shown a confidence banner because that's also put behind the condition `framework['status'] == 'open'`.

app/main/views/frameworks.py
```diff
@@ -68,21 +68,23 @@ def framework_dashboard(framework_slug):
...
     lots_with_completed_drafts = [lot for lot in framework['lots'] if count_drafts_by_lot(complete_drafts, lot['slug'])]
 
     framework_dates = content_loader.get_message(framework_slug, 'dates')
     framework_urls = content_loader.get_message(framework_slug, 'urls')
 
+    # Confidence banner.
+    if application_made and framework['status'] == 'open':
+        flash('submission_confidence', 'success')
```
risicle: So we're flashing a message... to ourselves? Is this a normal pattern in our code? (genuine question)

tests/app/main/test_frameworks.py
```diff
@@ -1640,6 +1629,137 @@ def test_make_or_edit_supplier_declaration_shows_correct_page(self, data_api_cli
...
+        }
+        data_api_client.get_supplier_framework_info.return_value = self.supplier_framework(status='complete')
+
+        with self.app.test_client():
+            self.login()
+            resp = self.client.get("/suppliers/frameworks/g-cloud-8")
+
+        assert resp.status_code == 200
+        # Assert mock.
+        flash_mock.assert_called_once_with('submission_confidence', 'success')
```
risicle: Surely we just care that the banner is shown in the right cases - it's an internal implementation detail whether the view is using the flash mechanism to do it surely? Are we overcomplicating something...? Wonder what others think...

app/main/views/frameworks.py
benvand: Yeah, in general we flash a 'key' that corresponds to a message value in the template for banners.
Thought it was a bit weird at first but it does allow quite nicely for the separation of logic and content, although going forward it would be nice if we had some sort of location for all key: message/ value pairs.

tests/app/main/test_frameworks.py
benvand: The idea of this test is to monitor the boundary contract between code and template. If the view doesn't flash the appropriate key then there will be a problem in the template ie the contract between the view and template has changed.

app/main/views/frameworks.py
benvand: Yeah agreed actually I'll update the template to take `supplier_is_on_framework` and do the check itself. An application _having been made_ isn't strictly what the template cares about

TheDoubleK:
I don't think this should be a flash message at all - the template already has all the information it needs to decide whether or not to show the banner - I don't think any new view logic or parameters are required.

I reckon under the existing flash message rendering code in the template, put 
```
{% if application_made and framework.status == 'open' %}
   banner message
{% endif %}
```

And then just need the two tests for when the banner is and isn't shown.

app/main/views/frameworks.py
TheDoubleK: I think we generally flash messages for a redirect - e.g. POST some data, then flash a message to be shown on the redirected page.  I don't think we have GET routes that flash to themselves.

benvand:
So I've made a PR with your suggestion @TheDoubleK 
https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/641

I'd like to register my thoughts on this before we close this PR though.

1. I would prefer the logic that determines whether to show the banner be included in the view and not the template. The new template blurs the line between responsibilities.
2. We have too many ways to introduce a banner onto a page.

Both of the above points can all be solved with a decent refactor of banner handling which I understand is upcoming/ proposed so I'm happy to shelve this for now.

PR #635, opened 2017-02-28
Test `_render_not_eligible_for_brief_error_page` using unit tests not integration and remove old supplier flow

idavidmcdonald:
### Part 1
Several of our brief routes hand off to `_render_not_eligible_for_brief_error_page` if the supplier is not eligible for the brief. We were originally testing this functionality by integration tests covering some but not all of the routes which use it. This was not a great solution as it either meant we were duplicating our tests to cover several different ineligiblity situations for each route, or meant that we were not properly testing all the possible situations for each route.

As we are about to remove the 'legacy flow' for suppliers applying to a brief response, this route had the detailed tests covering all the situations when a supplier was not eligible for the brief. As these tests will shortly be removed it left the situation that we could move these tests to a different route or we could instead unit test the functionality.

Now we have unit tests for `_render_not_eligible_for_brief_error_page` and check all routes that use it do hand off to this method when the supplier is not eligible.

One thing to note is that whilst we will be checking that the correct variables are handed to the template, we do not have tests that test the template. We should add functional tests for this
instead.

Also in a few places I have added `autospec` to our mocks as this reduces testing errors with incorrect assert statements.

### Part 2
We remove the old supplier applies for a brief response flow. Note, we can not remove the feature flags because the response page relies on them. If we wish to remove them we need to have a rethink about we identify a brief being on the old/new flow rather than comparing it's publication date to the feature flag date. This is important for briefs with no responses as we would usually use brief response data to indicate if it is the old or new flow we are dealing with.


tests/app/main/test_briefs.py
```diff
@@ -360,7 +352,7 @@ def test_submit_clarification_question_escapes_html(self, send_email, data_api_c
 
 
 class TestApplyToBrief(BaseApplicationTest):
-    """Tests requests for the new multipage flow for applying for a brief"""
+    """Tests requests for the multipage flow for applying for a brief"""
```
TheDoubleK: You could arguably also remove the word `multipage` here.

tests/app/main/test_briefs.py
```diff
@@ -2402,19 +1912,107 @@ def test_supplier_details_shown(self, data_api_client):
...
+        render_template.assert_called_with(
+            "briefs/not_is_supplier_eligible_for_brief_error.html",
+            clarification_question=False,
+            framework_name='Digital Outcomes and Specialists',
+            lot='digital-specialists',
+            reason='supplier-not-on-lot',
+            data_reason_slug='supplier-not-on-lot',
+        )
+
+    def test_not_on_role(self, render_template, current_user, data_api_client):
```
TheDoubleK: "not on role"?  Really? Looks like it's in now so let's keep it :smile:

TheDoubleK:
This looks fine, though I'm not clear why we can't unit test the `_render_not_eligible_for_brief_error_page` to make sure it's showing the right messages?

tests/app/main/test_briefs.py
idavidmcdonald: We can change it. I believe your comment was about the fact it could be clearer. Maybe I'll do something more like `test_supplier_has_service_on_lot_but_not_for_this_role` or `test_supplier_does_not_have_service_with_correct_role`. Thoughts?

idavidmcdonald:
If we wish to test `_render_not_eligible_for_brief_error_page` including testing a rendered template then we run into issues. This is because `_render_not_eligible_for_brief_error_page` is not itself a route (you can't visit just visit a url using the test client and test this logic as it is a function that just so happens to render a template). If we wanted to test the template rendering we need to be in the app context of having received a request (i.e. come to it by a url) as the context is used is passed to the template rendering. I had a play around trying to mock the app context for a few hours and failed miserably, however if anyone can show me how to do it then all ears open. I had no luck on google/S.O etc.

I do wonder if having a helper method that renders a template is a good thing to be doing with Flask or if just no one else seems to want to be doing it.

tests/app/main/test_briefs.py
TheDoubleK: It's not just the test name though - looks like we have `reason='supplier-not-on-role',` baked in to our app, in which case the test name makes sense.

I can see why it was done (for consistency with "not on lot"), but "does not provide role" would have been more accurately descriptive.  Given that `'supplier-not-on-role'` is a value in the app though let's stick to it here.

TheDoubleK:
Fair enough - you've made the functional tests now too so let's do this. :+1:

PR #636, opened 2017-02-28
[#138323307]declaration overview: use secondary action link for "Return to Application" link

risicle:
Followup to story https://www.pivotaltracker.com/story/show/138323307


samuelhwilliams:
The spacing looks better with this but it feels like it could also using the secondary-button styling to be a bit more consistent with the question-editing pages?

screen shot 2017-03-01 at 08 20 28
screen shot 2017-03-01 at 08 20 37

This works fine though, so 👍 otherwise.

samuelhwilliams:
I guess it's consistent with the service editing summary and overview pages, so 🤷‍♂️

PR #638, opened 2017-02-28
removes hanging bullets and adds secondary bullet style

ralph-hawkins:
Everyone hates hanging bullets, so I removed them. Adds a secondary bullet style to make nested lists clearer.

Looks like this:
screen shot 2017-02-28 at 14 19 03

### DIFF
--- a/app/assets/scss/_supplier-declaration.scss
+++ b/app/assets/scss/_supplier-declaration.scss
@@ -9,13 +9,21 @@
   .question-heading li,
   .hint li {
 
+
     margin: 10px 0 10px 20px;
     list-style: disc;
 
-    @include media(1080px) {
-      margin-left: 0;
-    }
+    > ul {
+        li {
+          list-style: none;
+        } 
+      }
 
+    > ul li:before {
+      content: '\2013';
+      position: absolute;
+      margin-left: -20px;
+    }
   }
 
   .validation-wrapper li {


samuelhwilliams:
Is this something that can/should be done across the whole app rather than just the supplier declaration? Feels like a pattern we would probably want to use everywhere as opposed to just this page. If that is the case, we should probably do equivalent PRs for buyer and admin frontends until we have a way to unify resources like this.

samuelhwilliams:
Would this make more sense to do in dm-frontend-toolkit so that it's consistent across all pages on all apps? Feels like a pattern we might want to use as standard rather than just on the supplier declaration.

samuelhwilliams:
But for now, 👍

ralph-hawkins:
@pcraig3 do you have an opinion on putting this in the toolkit, does the declaration use the same styles?

@superroz how much we want to encourage nested lists?

TheDoubleK:
There are already toolkit styles for lists - can we just use them?
http://alphagov.github.io/digitalmarketplace-frontend-toolkit/shared_scss/lists.html

TheDoubleK:
Oh, sorry - they maybe don't have nested lists.  But if we're using nested lists maybe those styles should live there in the toolkit.

ralph-hawkins:
Spoke to @superroz and we don't use nested bullets anywhere and we don't want to. They are something that we'd like to contained to the declaration. So I'm going to add them here and not in the toolkit.

samuelhwilliams:
We don't want this pattern to spread without people having to justify it, so will merge it to here rather than the general frontend-toolkit.

PR #639, opened 2017-02-28
frameworks dependency bump

risicle:
This should fix multiquestion madness, but tests won't pass until https://github.com/alphagov/digitalmarketplace-frameworks/pull/393 is merged.

Before
20170229_uk_before
After
20170228_uk_after

~~~Also have slipped in a fix whereby the validator was expecting a field named `distortedCompetition` instead of  `distortingCompetition`.~~~

None of these changes were covered by the tests, which is a bit scary...

### DIFF
--- a/app/main/helpers/validation.py
+++ b/app/main/helpers/validation.py
@@ -194,7 +194,7 @@ class DOSValidator(DeclarationValidator):
             "mitigatingFactors": [
                 'misleadingInformation', 'confidentialInformation', 'influencedContractingAuthority',
                 'witheldSupportingDocuments', 'seriousMisrepresentation', 'significantOrPersistentDeficiencies',
-                'distortedCompetition', 'conflictOfInterest', 'distortedCompetition', 'graveProfessionalMisconduct',
+                'distortedCompetition', 'conflictOfInterest', 'distortingCompetition', 'graveProfessionalMisconduct',
                 'bankrupt', 'environmentalSocialLabourLaw', 'taxEvasion'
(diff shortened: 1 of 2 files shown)


TheDoubleK:
DOS had bot `distortingCompetition` and `distortedCompetition` questions.  The DOSValidator is doing the right thing for DOS and doesn't need to be changed.

If one of the questions has been dropped now then it's the G9Validator that should be updated, not the DOSValidator.

TheDoubleK:
Just had a look and all G8, DOS, DOS2 and G9 have both `distorting` and `distorted` competition.  I don't think the validator is wrong.

risicle:
*Oh* - it looks like you're right - I was doing this together with @samuelhwilliams - I guess we were misleading each other.... will remove that commit.

risicle:
Folie a deux corrected, rebased.

samuelhwilliams:
There was no folie a deux.

https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/main/helpers/validation.py#L197

This line listed 'distortedCompetition' twice; one of them should be 'distortingCompetition'.

TheDoubleK:
Sorry @risicle - you can put it back in.
@samuelhwilliams pointed out to me that it's a duplicated key in the list.  Both distorting and distorted should be in there.

risicle:
Merde.

samuelhwilliams:
frameworks-6.3.2 is in, so this should be good to go. 👍

PR #641, opened 2017-03-01
140124261-confidence-banner-on-framework-overview-2

benvand:
* New banner on framework overview page
  * https://preview.development.digitalmarketplace.service.gov.uk/suppliers/frameworks/g-cloud-9
* Tests

### DIFF
--- a/app/templates/frameworks/dashboard.html
+++ b/app/templates/frameworks/dashboard.html
@@ -48,6 +48,16 @@
     {% endfor %}
   {% endwith %}
 
+  {# Confidence Banner #}
+  {% if application_made and framework.status == 'open' %}
+    {% with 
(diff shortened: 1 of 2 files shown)


TheDoubleK:
Think you need `type="success"` in the banner to make it green. It doesn't look right at the moment (see http://alphagov.github.io/digitalmarketplace-frontend-toolkit/notification-banner.html):
screen shot 2017-03-01 at 16 36 37

TheDoubleK:
Just saw your comment about the view should decide if the banner is shown - that's fine, pass it as a parameter to the template in the backend though, not as a flash message in the request :)

TheDoubleK:
Although pretty much everything on that page appears or disappears based on values of things passed in. I don't really see the value of adding yet another parameter that can be derived from things already passed in?

TheDoubleK:
Good to go after a rebase :+1:

PR #642, opened 2017-03-01
140507667-no-supplier-name-in-zendesk-clarification-question

benvand:
When we send supplier clarification questions into Zendesk, we currently
show:
* Supplier name
* User name

People will now be able to see this when they are answering the
question.  We need the people answering questions to be independent.

We need to change this to supplier ID so that it is not obvious who
asked the question

Changes:

* Change template to use supplier id
* Test


TheDoubleK:
Good stuff :+1:

PR #644, opened 2017-03-01
Bugfix: require supplier_framework entry to create and edit drafts

TheDoubleK:
Fixes this bug: https://www.pivotaltracker.com/story/show/138050747

Previously it was possible for a supplier user to enter the "create draft" url directly and begin submitting drafts without ever having registered interest in the framework.

Changes here make it impossible to do the following things to draft services if no supplier_framework entry exists for the current supplier:
* create
* copy
* complete
* edit
* delete
* remove subsection


samuelhwilliams:
Feels like it might be possible to combine those tests into one or extract some of the core code since they're all doing/checking something very similar... but at the moment I think I prefer nice explicit tests to ones that are overly generalised. So happy for this to go in as-is.

PR #645, opened 2017-03-01
BUGFIX: Pass question numbers through to all question types

TheDoubleK:
Fixes this bug: https://www.pivotaltracker.com/story/show/140851855

Some question numbers in supplier declaration were not being displayed following the recent change to use kwargs in the question macros.  This explicitly passes question number to all question types so will fix the issue.


samuelhwilliams:
👍

PR #647, opened 2017-03-02
[140837783] - First question, no anchor

samuelhwilliams:
For each section in the supplier declaration, the first question's edit link should load without any anchors. This will ensure that the top of the page is visible, showing any section description/guidance/preamble for users who click the question-level edit rather than the section-level edit.


risicle:
Sminky pinky :+1:

PR #648, opened 2017-03-02
[#139553967]service submission pages: merge generation of "You need to make the supplier declaration" banner, retarget

risicle:
Story https://www.pivotaltracker.com/story/show/139553967

Previously this code was duplicated between the submission and submission lot pages
and linked to the first question page of the declaration (the old way of doing things).
instead we now point at the declaration overview or the "start declaration" page
depending on the status of the declaration. also improve testing coverage of this
banner's presence/absence.

### DIFF
--- a/app/templates/partials/service_warning.html
+++ b/app/templates/partials/service_warning.html
@@ -1,9 +1,9 @@
-{% if true %}
+{% with declaration_url = url_for('.framework_start_supplier_declaration' if declaration_status == 'unstarted' else '.framework_supplier_declaration_overview', framework_slug=framework.slug) %}
   {%
     with
-    message = 'This service will not be submitted at the deadline because the <a href="'|safe + url_for('.framework_supplier_declaration', framework_slug=framework.slug) + '">supplier&nbsp;declaration</a> hasn’t been made.' | safe,
+    message = 'You need to <a href="'|safe + declaration_url + '">make the supplier&nbsp;declaration</a> before any services can be submitted'|safe,
     type = 'warning'
   %}
     {% include 'toolkit/notification-banner.html' %}
   {% endwith %}
-{% endif %}
+{% endwith %}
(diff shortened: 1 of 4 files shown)


app/templates/partials/service_warning.html
```diff
@@ -1,9 +1,9 @@
-{% if true %}
+{% with declaration_url = url_for('.framework_start_supplier_declaration' if declaration_status == 'unstarted' else '.framework_supplier_declaration_overview', framework_slug=framework.slug) %}
   {%
     with
-    message = 'This service will not be submitted at the deadline because the <a href="'|safe + url_for('.framework_supplier_declaration', framework_slug=framework.slug) + '">supplier&nbsp;declaration</a> hasn’t been made.' | safe,
-    type = 'warning'
+    message = 'You need to <a href="'|safe + declaration_url + '">make the supplier&nbsp;declaration</a> before any services can be submitted'|safe,
+    type = notification_type or 'warning'
```
TheDoubleK: I'm not sure it's worth having the `notification_type` here - can't see when we'd ever use it?

app/templates/partials/service_warning.html
```diff
@@ -1,9 +1,9 @@
-{% if true %}
```
TheDoubleK: Whoa!  What was this `if true` all about?!?

TheDoubleK:
Left a comment about the currently unused param but OK to go in either way :+1:

app/templates/partials/service_warning.html
risicle: Agreed I was just hesitant to break any existing flexibility. Changing

app/templates/partials/service_warning.html
risicle: My guess is it was a previous condition stubbed out and not tidied before commit.

PR #649, opened 2017-03-02
BUGFIX-false-values-in-form

benvand:
* Catch correct false value in reuse form
  * As per
http://stackoverflow.com/questions/38088806/wtforms-booleanfield-treats-json-false-value-as-true

### DIFF
--- a/app/main/forms/frameworks.py
+++ b/app/main/forms/frameworks.py
@@ -45,6 +45,7 @@ class ReuseDeclarationForm(Form):
     """
     reuse = BooleanField(
         'Do you want to reuse the answers from your earlier declaration?',
+        false_values={'False', 'false', ''},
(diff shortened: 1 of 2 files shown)


app/main/forms/frameworks.py
```diff
@@ -45,6 +45,7 @@ class ReuseDeclarationForm(Form):
     """
     reuse = BooleanField(
         'Do you want to reuse the answers from your earlier declaration?',
+        false_values={False, 'False', 'false', ''},
```
TheDoubleK: The WTForms docs say that `false_values` should be a sequence of strings, so not sure about including the raw `False` in here.
http://wtforms.readthedocs.io/en/latest/fields.html

TheDoubleK:
:+1: once Travis is happy.

PR #650, opened 2017-03-02
TestDeclarationOverview: slightly deobfuscate by adding comments

risicle:
@samuelhwilliams mentioned the paratrization was a little hard to decode. Hence added comments. Better, @samuelhwilliams ?

### DIFF
--- a/tests/app/main/test_frameworks.py
+++ b/tests/app/main/test_frameworks.py
@@ -2324,7 +2324,7 @@ class TestDeclarationOverview(BaseApplicationTest):
             empty_declaration,  # noqa
             False,
             prefill_fw_slug,
-            (
+            (   # expected result for "Providing suitable services" section as returned by _extract_section_information
                 "Providing suitable services",
                 "/suppliers/frameworks/g-cloud-9/declaration/edit/providing-suitable-services",
                 (
@@ -2368,7 +2368,8 @@ class TestDeclarationOverview(BaseApplicationTest):
                     ),
                 ),
             ),
-            (
+            (   # expected result for "Grounds for mandatory exclusion" section as returned by
+                # _extract_section_information
                 "Grounds for mandatory exclusion",
                 "/suppliers/frameworks/g-cloud-9/declaration/edit/grounds-for-mandatory-exclusion",
                 (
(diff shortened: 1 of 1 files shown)


TheDoubleK:
Works for me :+1:

benvand:
https://www.python.org/dev/peps/pep-0008/#inline-comments
> Use inline comments sparingly.

The comments don't make this any more readable.
The _statement_ itself does too much. 
It reminds me of when people used to challenge each other to do an entire task in one line on codewars.com. It was always eventually possible, but rarely, if ever advisable.

risicle:
But this is how functional declarative programming is done and though it creates a few challenges over formatting & commenting etc., it means that every section in a set of brackets can be analyzed and understood totally separately from neighbouring scopes because it's all side-effect free.

risicle:
(Yeah yeah "Why don't you move to haskellania if you love your fp so much...")

risicle:
I don't particularly see how using OO here would really benefit - it doesn't help that pytest seems to encourage use of classes are more or less organizational containers, not much more. FP and OO aren't mutually exclusive - they actually work really well together. I don't intend to pull away from OO using FP. And while performance isn't as wonderful as it *could* be, there are many cases where FP approaches being able to make use of lazy evaluation of iterables/generators will cause significant performance *advantages* (lazy evaluation doesn't work so well for key-value structures in python, I tend to use it less in these cases anyway).

It didn't really take much time to write - the time in writing these tests came from deciding what to test, what to assert and what not to assert. Although there's an initial investment in decoding the setup of the test, I think it is far more straightforward to understand a combined test than tests written in a more copy-pasta style that achieve the same amount of coverage (there would be a lot of copy-pasta). It also encourages someone coming to alter something to the view to consider the behaviour in all cases and understand the view as a whole, as opposed to just cherry-picking cases where they *think* their change is relevant and disregard other cases.

I've moved towards writing tests like this because I found my previous ways of doing it inadequate and actually ended up having a higher cognitive load because I had a harder time convincing myself that I'd written a test for all the "important" cases. This way, it's clear what gets tested and what doesn't - the actual test code is pretty compact and the machinery above is just there to feed it with the right combinations.

Finally, when I use this approach I regularly find significant numbers of bugs that would otherwise have made it through to released code, and I don't feel it's adequate to not write this style of tests for my code.

PR #652, opened 2017-03-02
Add checkbox_tree component to supplier app

pcraig3:
This pull request:
- pulls in the latest `frontend-toolkit` with the checkbox_tree pattern
- pulls in the latest `frameworks` with the new service categories
- pulls in the latest `content-loader` which knows how to process checkbox_tree data
- adds a wrapper macro for a `checkbox_tree` question
- brings in the ability to chain show/hide questions 

### no child categories (looks like a flat list)
checkbox_tree

### parent and child categories (has an open/close widget)
checkbox_tree

### DIFF
--- a/app/templates/macros/toolkit_forms.html
+++ b/app/templates/macros/toolkit_forms.html
@@ -126,6 +126,26 @@
   {% endwith %}
 {%- endmacro %}
 
+{% macro checkbox_tree(question_content, service_data, errors) -%}
+  {%
+    with
+    name=question_content.id,
+    question=question_content.question|question_references(kwargs.get_question),
+    question_advice=question_content.question_advice,
+    hint=(question_content.hint or '')|question_references(kwargs.get_question),
+    value=service_data[question_content.id],
+    id=question_content.id,
+    options=question_content.options,
+    type='checkbox_tree',
+    number_of_items=question_content.number_of_items,
+    question_number=kwargs.question_number,
+    hidden=question_content.hidden,
+    error=errors.get(question_content.id)['message']|question_references(kwargs.get_question)
+  %}
+    {% include "toolkit/forms/checkbox-tree.html" %}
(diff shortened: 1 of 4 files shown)


app/templates/macros/toolkit_forms.html
```diff
@@ -126,6 +126,24 @@
   {% endwith %}
 {%- endmacro %}
 
+{% macro checkbox_tree(question_content, service_data, errors, question_number=None, get_question=None) -%}
```
TheDoubleK: The rest of the macros in this file have changed to use kwargs instead of named params.
So you can do
```
{% macro checkbox_tree(question_content, service_data, errors) -%}
...
     question=question_content.question|question_references(kwargs.get_question)
     hint=(question_content.hint or '')|question_references(kwargs.get_question),
```
etc.

TheDoubleK: Might not hurt to additionally pass through 
`question_number=kwargs.question_number` - won't be needed for now but future-proofing for if we ever use one of these in the declaration or add numbers to service questions.

Although neither of those things feel likely, so could skip it.

app/templates/macros/toolkit_forms.html
```diff
@@ -126,6 +126,26 @@
...
+    question_advice=question_content.question_advice,
+    hint=(question_content.hint or '')|question_references(kwargs.get_question),
+    value=service_data[question_content.id],
+    id=question_content.id,
+    options=question_content.options,
+    type='checkbox_tree',
+    number_of_items=question_content.number_of_items,
+    question_number=kwargs.question_number,
+    hidden=question_content.hidden,
+    error=errors.get(question_content.id)['message']|question_references(get_question)
```
TheDoubleK: This `get_question` still needs a `kwargs`.

TheDoubleK:
:+1:

PR #655, opened 2017-03-03
Remove hardcoded framework name

idavidmcdonald:
This should be taken from the framework instead as it may be a DOS2 opportunity for example.


TheDoubleK:
Nice spot - :+1:

PR #659, opened 2017-03-06
Add javascript stuff for category-picker

pcraig3:
The javascript that runs overtop of the category-picker currently makes the assumption that there is only one category-picker component on a page, so it doesn't really work on [the "examples" page of the frontend toolkit](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/forms/checkbox-tree.html) where there are multiple of them.

That being so, I've kept it in the supplier app for the time being until it can be further generalised.

It's also worth pointing out that we don't have a way of including javascript on a per-question basis in the app, so I've just added all the category-picker-specific stuff to the general `application.js` file.  It won't cause any problems or anything, it's just something to be aware of.

This pull request:
- adds a polyfill so that the `<details>` element works on older browsers
- sticks the `counter` element to the top of the browser while scrolling through categories
- handles the logic for the category-picker
  - will tick and untick all non-unique categories at once
  - it will update the counter
  - it will update the number of selected categories within each parent grouping
- brings in the new frontend toolkit with the needed corresponding markup

***

### sticky counter element
sticks at the top and unsticks at the bottom (even if the page changes size) 
sticky_counter

### updates the counter as elements are ticked
counter_updated

### updates the parent context as child categories are ticked
parent_context

### will select/unselect all child categories with the same name
multiple_categories

### DIFF
--- a/app/assets/javascripts/_stick-at-top-when-scrolling.js
+++ b/app/assets/javascripts/_stick-at-top-when-scrolling.js
@@ -1,0 +1,19 @@
+(function(GOVUK, GDM) {
+
+  GOVUK.GDM = GDM;
+
+  GDM.stickAtTopWhenScrolling = function() {
+
+    if (!GOVUK.stickAtTopWhenScrolling) return;
+
+    GOVUK.stickAtTopWhenScrolling.init();
+
+    $('#checkbox-tree__inputs').on('click', 'details', function () {
+      // immediately after the click, the details pane hasn't expanded yet
+      setTimeout(function () {
+        GOVUK.stopScrollingAtFooter.updateFooterTop()
+      }, 100)
+    })
+  }
+
+}).apply(this, [GOVUK||{}, GOVUK.GDM||{}]);
--- a/app/assets/javascripts/category-picker.js
+++ b/app/assets/javascripts/category-picker.js
@@ -1,0 +1,189 @@
+;(function(root) {
+
+  "use strict";
+
+  var GOVUK = root.GOVUK;
+
+  function _unique(array) {
+    return $.grep(array, function(el, index) {
+        return index === $.inArray(el, array);
+    });
+  }
+
+  // wrapper around the categories data
+  var Categories = function (_categories) {
+    var primaries = []
+    var self = this
+
+    this._containers = []
+    this._categories = _categories
+    this.save_last_state()
+  }
+  Categories.prototype.save_last_state = function () {
+    this._lastState = $.extend(true, {}, this._categories)
+  }
+  Categories.prototype.tick_by_name = function (category_names) {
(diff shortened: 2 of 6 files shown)


app/assets/javascripts/category-picker.js
```diff
@@ -0,0 +1,189 @@
...
+      $('.categories-heading:contains("' + parentCategory + '")').next().text(parentContext)
+    }
+
+    renderCategories()
+    setCounter()
+    setParentContextCounters()
+
+    // make clicks on checkboxes update the categories state
+    // TODO: looks like the event is triggered a bunch of times
+    $('#checkbox-tree__inputs').on('click', 'label', function (evt) {
```
carolinegreen: "event", You've typed it out above, it's not that hard :p

app/assets/javascripts/_stick-at-top-when-scrolling.js
```diff
@@ -0,0 +1,19 @@
+(function(GOVUK, GDM) {
+
+  GOVUK.GDM = GDM;
```
carolinegreen: What does GDM stand for?  Guessing, something Digital marketplace?  I'm just curious, don't remember it.

pcraig3: I don't know, actually.  It's basically like the fake experiment about [the monkeys and the ladder](http://www.wisdompills.com/2014/05/28/the-famous-social-experiment-5-monkeys-a-ladder/) now.

app/assets/javascripts/category-picker.js
```diff
@@ -0,0 +1,189 @@
...
+  function _unique(array) {
+    return $.grep(array, function(el, index) {
+        return index === $.inArray(el, array);
+    });
+  }
+
+  // wrapper around the categories data
+  var Categories = function (_categories) {
+    var primaries = []
+    var self = this
```
carolinegreen: I might be missing something here, but why are you assigning this to a self variable?

app/assets/javascripts/_stick-at-top-when-scrolling.js
carolinegreen: Lol, I hadn't heard the monkey story before.  I guess it's another documentation/tech debt? thing.

TheDoubleK: "Government Digital Marketplace" - a relic from the old Grails app I think.

carolinegreen: Ohhh

samuelhwilliams:
This is sick. 👍

TheDoubleK:
Totally fly.

PR #661, opened 2017-03-06
Require Jinja pinned dependancy

idavidmcdonald:
Flask requires Jinja but we found a bug with differing versions of Jinja on our different environments. We require all environments to run the same version and so therefore pin this dependancy.


TheDoubleK:
:+1: once Travis is happy.

PR #662, opened 2017-03-06
framework_supplier_declaration_edit: add extra test, testing validation works properly on re-submit of a previously valid declaration section

risicle:
This test was the result of a wild goose chase chasing a phantom bug (https://www.pivotaltracker.com/story/show/141116685). it's still a valid test though, so it's probably worth adding anyway.


benvand:
https://lgtm.in/p/Iu6g6MKUp

👍

PR #667, opened 2017-03-08
140507667-no-supplier-username-in-zendesk-clarification-question

benvand:
* Remove user name from clarification email
* Update tests

### DIFF
--- a/app/templates/emails/clarification_question.html
+++ b/app/templates/emails/clarification_question.html
@@ -5,8 +5,6 @@
 </head>
 <body>
 Supplier ID: {{ supplier_id }}
-<br />
-User name: {{ user_name }}
 <br /><br />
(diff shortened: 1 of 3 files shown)


app/templates/emails/clarification_question.html
```diff
@@ -6,7 +6,7 @@
 <body>
 Supplier ID: {{ supplier_id }}
 <br />
-User name: {{ user_name }}
+User ID: {{ user_name }}
```
TheDoubleK: I think we just want to drop user altogether and only have Supplier ID.  I don't see any value in showing two IDs here?

benvand:
@TheDoubleK Good point, sorted

TheDoubleK:
:+1: nice one.

PR #668, opened 2017-03-08
Use full words for variables in category-picker.js file

pcraig3:
Super tiny change to use full words rather than their shortened versions as variable names.

'evt' -> 'event'
'idx' -> 'index'

This was something that @carolinegreen mentioned earlier but I forgot to make the change.


carolinegreen:
Much nicer. 👍

PR #669, opened 2017-03-08
FLAKE8-fixes

benvand:
These are flake8 changes being checked against the pep8 checks


app/main/views/briefs.py
```diff
@@ -360,12 +358,7 @@ def view_response_result(brief_id):
         supplier_id=current_user.supplier_id
     )['briefResponses']
 
-    legacy_brief = True
```
benvand: So I'm trying to amalgamate:
```
-    if not (datetime.strptime(current_app.config['FEATURE_FLAGS_NEW_SUPPLIER_FLOW'], "%Y-%m-%d")
-            <= datetime.strptime(brief['publishedAt'], DATETIME_FORMAT)):
+    if not new_supplier_flow_active_for_datetime(breif_published_at):
```
```
-    if current_app.config['FEATURE_FLAGS_NEW_SUPPLIER_FLOW'] and \
-        (datetime.strptime(current_app.config['FEATURE_FLAGS_NEW_SUPPLIER_FLOW'], "%Y-%m-%d")
-            <= datetime.strptime(brief['publishedAt'], DATETIME_FORMAT)):
-        abort(404)
+    if new_supplier_flow_active_for_datetime(brief_published_at):
+            abort(404)
```
```
-    if current_app.config['FEATURE_FLAGS_NEW_SUPPLIER_FLOW']:
-        legacy_brief = (
-            datetime.strptime(current_app.config['FEATURE_FLAGS_NEW_SUPPLIER_FLOW'], "%Y-%m-%d")
-            > datetime.strptime(brief['publishedAt'], DATETIME_FORMAT))
-
+    legacy_brief = not new_supplier_flow_active_for_datetime(brief_published_at)
```

idavidmcdonald: I probably wouldn't go to too much/any effort in trying to tidy up the bit around feature flags as it is about to get stripped out (apart from I think one location) as the old flow is no longer needed. There is currently a pull request up that does this - https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/635

app/main/views/briefs.py
```diff
@@ -360,12 +358,7 @@ def view_response_result(brief_id):
...
-    if current_app.config['FEATURE_FLAGS_NEW_SUPPLIER_FLOW']:
-        legacy_brief = (
-            datetime.strptime(current_app.config['FEATURE_FLAGS_NEW_SUPPLIER_FLOW'], "%Y-%m-%d")
-            > datetime.strptime(brief['publishedAt'], DATETIME_FORMAT))
-
+    legacy_brief = not new_supplier_flow_active_for_datetime(brief_published_at)
     if len(brief_response) == 0:
```
benvand: Resolves to:
`(not feature_flag_date <= brief_published_at)`
which is equal to:
` feature_flag_date > brief_published_at`

benvand:
```
import pytest
from datetime import datetime, timedelta

d1 = datetime.now()
d2 = d1 + timedelta(seconds=1)

@pytest.mark.parametrize(('d1', 'd2'), ((d1, d2), (d1, d1), (d2, d1)))
def test_dates(d1, d2):
    assert (not d1 <= d2) == (d1 > d2)
```

app/main/helpers/validation.py
```diff
@@ -62,9 +62,6 @@ def get_error_message(self, question_id, message_key):
         return default_messages.get(
             message_key, 'There was a problem with the answer to this question')
```
benvand: Removed as never used. Is redefined below.

app/main/helpers/briefs.py
```diff
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 
 import six
```
benvand: unused import

app/main/views/briefs.py
benvand: Ah yeah, either way it works now. I'll look at that other pr today if you like?

tests/app/main/test_briefs.py
```diff
@@ -57,6 +56,41 @@
     ' could provide this specialist role when you applied to the Digital Outcomes and Specialists framework.'
```
benvand: The below is a *Too complex* error. No need to define the class within the method

tests/app/main/test_suppliers.py
```diff
@@ -367,7 +367,7 @@ def test_shows_gcloud_7_closed_message_if_pending_and_application_done(
     @mock.patch("app.main.views.suppliers.data_api_client")
```
benvand: The below test name was declared twice meaning this test was run but the one of the same name above wasnt.

app/main/helpers/validation.py
allait: It defined the expected override in a subclass though. Without it you get a less friendly 'attribute not found' error if you forget the override in a new subclass.

TheDoubleK: Yep I think this is a useful thing to keep in the base class.

tests/app/main/test_frameworks.py
```diff
@@ -3993,7 +3993,8 @@ def test_we_abort_if_agreement_does_not_match_supplier_framework(
         supplier_framework = self.supplier_framework(framework_slug='g-cloud-8', on_framework=True)['frameworkInterest']
         return_supplier_framework.return_value = supplier_framework
 
-        res = self.client.get("/suppliers/frameworks/g-cloud-8/234/signer-details")
+        # Do request
```
TheDoubleK: I don't think we really need `Do request` as a comment on `.get()` and `.post()`... no biggie though.

app/main/helpers/validation.py
benvand: Are you suggesting keep it as documentation? That `NotImplementedError` can never raise. 
```
>>> class Foo(object):
	def bar(self):
		print 'baz'
	def bar(self):
		print 'quux'

		
>>> Foo().bar()
quux
>>> class InheritFoo(Foo):
	pass

>>> InheritFoo().bar()
quux
>>> 
```

tests/app/main/test_frameworks.py
benvand: I'm kind of in 2 minds. 
The way I interpreted the intention here is that the original writer was assigning to `res` to make it clear that the code was getting a response at that point (which I actually quite like) :man_shrugging:

benvand: Decision: They don't make sense unless the rest of the test is heavily commented. I'll take them out for tests without other comments.

app/main/helpers/validation.py
TheDoubleK: Sorry, it wasn't clear from the diff that it was defined further on in the base class.

PR #672, opened 2017-03-09
140921547-Measure-how-people-are-using-the-new-evidence-flow

benvand:
* Add result=success param to url on successful brief submission (for analytics)
* Update tests

https://www.pivotaltracker.com/story/show/140921547

### DIFF
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -225,7 +225,7 @@ def edit_brief_response(brief_id, brief_response_id, question_id=None):
                     brief_response_id,
                     current_user.email_address
                 )
-
+                flash('submitted_first', 'success')
                 return redirect(url_for('.view_response_result', brief_id=brief_id))
 
     previous_question_id = section.get_previous_question_id(question_id)
@@ -274,11 +274,8 @@ def view_response_result(brief_id):
             return redirect(url_for(".start_brief_response", brief_id=brief_id))
     elif brief_response[0].get('essentialRequirementsMet') or all(brief_response[0]['essentialRequirements']):
         result_state = 'submitted_ok'
-        flash('submitted_ok', 'success')
     else:
         result_state = 'submitted_unsuccessful'
-        flash('submitted_unsuccessful', 'error')
-
     brief_response = brief_response[0]
     framework, lot = get_framework_and_lot(
         data_api_client, brief['frameworkSlug'], brief['lotSlug'], allowed_statuses=['live', 'expired'])
--- a/app/templates/briefs/view_response_result.html
+++ b/app/templates/briefs/view_response_result.html
@@ -26,27 +26,50 @@
 
 {% block main_content %}
 
+{# Messages #}
 {% with messages = get_flashed_messages(with_categories=True) %}
+  {# Already applied #}
   {% if ("error", "already_applied") in messages %}
     {%
     with
     message = "You’ve already applied so you can’t apply again.".format(brief.title) if result_state == "submitted_ok"
-                    else "You already applied but you didn’t meet the essential requirements.".format(brief.title),
+      else "You already applied but you didn’t meet the essential requirements.".format(brief.title),
     type = "destructive"
     %}
-    {% include "toolkit/notification-banner.html" %}
+      {% include "toolkit/notification-banner.html" %}
     {% endwith %}
   {% else %}
+    {# Successful application #}
+    {% if result_state == "submitted_ok"  %}
+      {%
+      with
+      message = "Your application has been submitted.".format(brief.title),
+      type = "success"
+      %}
+        {% include "toolkit/notification-banner.html" %}
+      {% endwith %}
+    {# Unsuccessful application #}
+    {% elif result_state == "submitted_unsuccessful" %}
+      {%
+      with
+      message = "You don’t meet all the essential requirements.".format(brief.title),
+      type = "destructive"
+      %}
+        {% include "toolkit/notification-banner.html" %}
+      {% endwith %}
+    {% endif %}
+    {# Flash messages/ first success handling/ default #}
     {% for category, message in messages %}
       {%
       with
-      message = {
-        "submitted_ok": "Your application has been submitted.".format(brief.title),
-        "submitted_unsuccessful": "You don’t meet all the essential requirements.".format(brief.title)
-      }[message] or message,
-      type = "destructive" if category == "error" else "success"
+      message = message,
+      type = "destructive" if category == "error" else category
       %}
-        {% include "toolkit/notification-banner.html" %}
+        {% if message == "submitted_first" %}
+          <span data-analytics="trackPageView" data-url={{ "/opportunities/{}/responses/result?result=success".format(brief.id) }}></span>
+        {% else %}
+          {% include "toolkit/notification-banner.html" %}
+        {% endif %}
       {% endwith %}
     {% endfor %}
   {% endif %}
@@ -68,7 +91,7 @@
 
       {% if 'essentialRequirementsMet' in brief_response %}
         <h3>Shortlist</h3>
-        
+
         <p>When the opportunity closes, the buyer will score your evidence. If you’re one of the top {{ brief.get('numberOfSuppliers') }} suppliers, you'll go through to the evaluation stage.</p>
         <p>The buyer will tell you if you're not successful.</p>
 
--- a/tests/app/main/test_briefs.py
+++ b/tests/app/main/test_briefs.py
@@ -1912,6 +1912,43 @@ class TestResponseResultPage(ResponseResultPageBothFlows, BriefResponseTestHelpe
         assert requirements_data.row(2).cell(0) == "Email address"
         assert requirements_data.row(2).cell(1) == "contact@big.com"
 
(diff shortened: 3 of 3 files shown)


idavidmcdonald:
👍 looks good after tests green

TheDoubleK:
Do we do this elsewhere - i.e. adding URL params just for analytics?

I thought we had some subtler way at one point, but I might be misremembering.

idavidmcdonald:
We use to add a parameter exactly like this for the 'old supplier application' flow, it just wasn't included in the 'new flow'. Can't say I know about elsewhere though...

TheDoubleK:
I'm sure I've had this conversation before with Tom... it's probably fine.  Just a bit weird to have a param that's never used by the view at all, and seems pretty unreliable too - what if people bookmark/refresh/hit back?

idavidmcdonald:
@benvand Might be worth asking Ash about his thoughts on that then...

benvand:
screen shot 2017-03-09 at 17 00 13

TheDoubleK:
Aha!  I knew there was a tricksy way of doing things.

Just to check - is the only addition here the wrapping `<div>`?  Only the diff shows that whole section as new but I think it's just the indenting. 

Final thing - are we sure the new `'grid-row'` doesn't adversely affect the page layout?

benvand:
Yeah found it late yesterday :tada:  I managed to test that it does send the analytics info looking at the network panel in the browser console. I'm going to write a test and tidy this up this morning.

app/templates/briefs/view_response_result.html
```diff
@@ -64,48 +64,51 @@
...
-        <p>
-          Your evidence must describe the skills and experience of the {{ 'person' if brief.lotSlug == "digital-specialists" else 'people' }} who’ll be working on the project.
-        </p>
-        <p>
-          The buyer will score all suppliers who reached the evaluation stage using the weightings they published with their requirements. They’ll provide feedback if you’re unsuccessful.
-        </p>
-      {% else %}
-        {% include 'briefs/_legacy_view_response_result_content.html' %}
-      {% endif %}
+      <div class="grid-row" data-analytics="trackPageView" data-url={{ "/opportunities/{}/responses/result/?result=success".format(brief.id) }}>
```
idavidmcdonald: Should there be the `/` after `result`?

idavidmcdonald: Also this solution will come up with the problem that if you ever visit this page having applied successfully then the `result_state` is always `submitted_ok` and we will trigger a page view every time they visit this page. This is potentially worse than adding the parameter as this parameter would only be added if they've come from the application flow and not just revisiting the result page
image

It could be worth investigating sending a flash message on the final page flow when you submit your application that we can then do a page track based on this flash message. That seems to be a solution that won't expose params in the url and will definitely only happen once, regardless of if they bookmark the page, refresh it or revisit it.

On another note, if we are using the trackPageView functionality, would be worth double checking with Ash if this also works with him. My assumption is it would mean two page views would be registered (one with and one without the result=success param).

idavidmcdonald: And with a quick glance at the view code, we are doing a bit of a weird thing where we are setting flash messages for the result page using the result page view. Usually we would be sending the flash page from the page before (we do this in most of the other brief routes). So potentially this could get some tidy up to stop doing this but might be creeping on the scope of the story so feel free to leave either way.

TheDoubleK: If I recall, that weird flash from the view itself was a thing added by Tom specifically for triggering analytics events only in the right circumstances - I don't think it's used for displaying a message on the page at all (at least, it shouldn't be?).

idavidmcdonald: Line 44 of this template? Seems to be the thing showing the notification message at the top page every time?

benvand: TBH although it's a little out of scope I think the idea about refactoring out the flash messages might be the best way to achieve this. Do we want to do this as part of this ticket? Are you guys happy for me to go ahead?

https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/ca5c06b80a40d2768e67cf6db8436f4119f62a37/app/main/views/briefs.py

idavidmcdonald: I say go for it, as long as the user behaviour remains the same (i.e. will still always be shown the successful message if they applied OK but didn't come direct from the application flow). Maybe worth double checking with Ash now regarding potentially sending two page views just in case that would change anything.

benvand: :+1:

tests/app/main/test_briefs.py
```diff
@@ -1912,6 +1912,44 @@ def test_supplier_details_shown(self, data_api_client):
         assert requirements_data.row(2).cell(0) == "Email address"
         assert requirements_data.row(2).cell(1) == "contact@big.com"
 
+    def test_analytics_and_messages_applied_on_first_submission(self, data_api_client):
+        """Go through submitting to edit_brief_response and the redirect to view_response_result. Assert messages."""
+        self.set_framework_and_eligibility_for_api_client(data_api_client)
+        data_api_client.get_brief.return_value = self.brief
+        # data_api_client.find_brief_responses.return_value = self.brief_responses
```
idavidmcdonald: Not sure what's happening with this line?

tests/app/main/test_briefs.py
```diff
@@ -1912,6 +1912,44 @@ def test_supplier_details_shown(self, data_api_client):
         assert requirements_data.row(2).cell(0) == "Email address"
         assert requirements_data.row(2).cell(1) == "contact@big.com"
 
+    def test_analytics_and_messages_applied_on_first_submission(self, data_api_client):
+        """Go through submitting to edit_brief_response and the redirect to view_response_result. Assert messages."""
```
idavidmcdonald: It looks like our other tests that test flash messages on this page are more unit (checking the flash is set on the previous page and then asserting if you visit this page with a flash message set that it shows the correct message). Either way works with me but just thought I'd point it out. Would be nice if our tests were more consistent but I don't think you should go to the effort of refactoring tests for that reason

tests/app/main/test_briefs.py
```diff
@@ -1912,6 +1912,44 @@ def test_supplier_details_shown(self, data_api_client):
...
+        self.set_framework_and_eligibility_for_api_client(data_api_client)
+        data_api_client.get_brief.return_value = self.brief
+        # data_api_client.find_brief_responses.return_value = self.brief_responses
+        data_api_client.get_brief_response.return_value = self.brief_response()
+        data_api_client.update_brief_response.return_value = {'success': True}
+        data_api_client.find_brief_responses.side_effect = [
+            {'briefResponses': []},
```
idavidmcdonald: Why are two brief responses coming back?

app/main/views/briefs.py
```diff
@@ -225,7 +225,7 @@ def redirect_to_next_page():
                     brief_response_id,
                     current_user.email_address
                 )
-
+                flash('submitted_first', 'success')
```
idavidmcdonald: `submitted_first` sounds a little unclear to me... saying that I don't think I have a better suggestion. The best I can think of is `submitted_on_previous_page`. Happy to keep as is if need be

tests/app/main/test_briefs.py
```diff
@@ -1912,6 +1912,44 @@ def test_supplier_details_shown(self, data_api_client):
...
+        data = res.get_data(as_text=True)
+
+        # Assert the analytics exists
+        assert (
+            '<span data-analytics="trackPageView" '
+            'data-url=/opportunities/1234/responses/result?result=success></span>' in data
+        )
+        assert data.count('data-analytics') == 1
+        # Assert we get the correct banner message (and only the correct one).
+        assert 'Your application has been submitted.' in data
```
idavidmcdonald: Should we be asserting more than the text on the page? e.g. that there is a banner message object and there is only one and that is is the 'success' type? This would be putting it really strictly though so feel free to ignore if you think it's overkill

tests/app/main/test_briefs.py
benvand: One in each view, the test does the whole process from the initial non existence of a brief response, it then does the submit and show result. 
So on the `edit` page we look for a brief response and don't find one so we submit a new brief response, then we end up on the result page where we get the brief response we just submitted to show the result of it.

tests/app/main/test_briefs.py
benvand: Yeah, so I thought about this but I decided the 2 occurrences are part of the same linear process.
What we _really_ care about in this test is that the analytics occur on the first submit. It doesn't matter how we achieve that; using the flash messages or some other freaky method that we haven't invented yet.
That was my thinking and why I didn't break it out.

app/main/views/briefs.py
benvand: `submitted_initial` sounds like it's an initial draft or something
`submitted_first_submit` sounds redundant
`first_submit` sounds like we're not sure if they'll have to submit again (at this point theyve been successful)
`submitted_success_first` ?

idavidmcdonald:
👍 All good to go!

PR #675, opened 2017-03-10
FLAKE8-do-flake8

benvand:
### Swap out pep8 checks for flake8

* New .flake8 file
  * Exclude virtualenvs, cache, bower and node stuff
  * Ignore blank line required before class docstring (pretty standard ignore)
  * Ignore unused imports in __init__ files
  * Ignore import not at top of file in _some_ __init__ files
  * Ignore complexity in views

* Makefile
  * New test_flake8 command to run flake8
  * Remove test_pep8 command
  * replace pep8 command with flake8 in test run command

* requirements_for_test
  * Remove pep8
  * Add flake8 and flake8-putty
  * Alphabetize

* setup.cfg
  * Remove unused pep8 and nose config entries

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,14 @@
+[flake8]
+# Rule definitions: http://flake8.pycqa.org/en/latest/user/error-codes.html
+# D203: 1 blank line required before class docstring
+# F401: 'identifier' imported but unused
+# E402: module level import not at top of file
+# C901: too complex
+exclude = venv,venv3,__pycache__,node_modules,bower_components
+ignore = D203
+max-complexity = 7
+max-line-length = 120
+putty-ignore =
+    **/__init__.py : +F401
+    app/__init__.py, app/main/__init__.py, app/status/__init__.py : +E402
+    app/main/views/*.py : +C901
--- a/Makefile
+++ b/Makefile
@@ -31,10 +31,10 @@ npm_install: package.json
 frontend_build:
 	npm run --silent frontend-build:${GULP_ENVIRONMENT}
 
-test: show_environment test_pep8 test_python test_javascript
+test: show_environment test_flake8 test_python test_javascript
 
-test_pep8: virtualenv
-	${VIRTUALENV_ROOT}/bin/pep8 .
+test_flake8: virtualenv
+	${VIRTUALENV_ROOT}/bin/flake8 .
 
 test_python: virtualenv
 	${VIRTUALENV_ROOT}/bin/py.test ${PYTEST_ARGS}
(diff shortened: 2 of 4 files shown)


.flake8
```diff
@@ -0,0 +1,13 @@
+[flake8]
+# D203: 1 blank line required before class docstring
```
TheDoubleK: Is it also worth including a URL where these rule codes are listed?

.flake8
```diff
@@ -0,0 +1,13 @@
...
+# E402: module level import not at top of file
+# C901: too complex
+exclude = venv,venv3,__pycache__,node_modules,bower_components
+ignore = D203
+max-complexity = 7
+max-line-length = 120
+putty-ignore =
+    **/__init__.py : +F401
+    app/__init__.py, app/main/__init__.py, app/status/__init__.py : +E402
+    app/main/views/*.py : +C901
```
TheDoubleK: Why aren't we enforcing complexity? Is our stuff too complex already?

Makefile
```diff
@@ -31,10 +31,10 @@ npm_install: package.json
 frontend_build:
 	npm run --silent frontend-build:${GULP_ENVIRONMENT}
 
-test: show_environment test_pep8 test_python test_javascript
+test: show_environment test_flake8 test_python test_javascript
 
-test_pep8: virtualenv
```
TheDoubleK: We don't even want an option on doing PEP8?  I guess not if they potentially conflict.

requirements_for_test.txt
```diff
@@ -1,12 +1,13 @@
 -r requirements.txt
-pytest==2.8.2
-pep8==1.5.7
-requests-mock==0.6.0
-mock==2.0.0
-lxml==3.4.4
+coverage==3.7.1
```
TheDoubleK: :clap: nice alphabetisation.

TheDoubleK:
:+1: for the 4th wall.

Makefile
benvand: So PyFlakes actually runs a pep8 check as well as a lint check. I can't see anyone needing to use pep8 as well as the flake8 command. Happy to add it back in later if people realise they want it but hopefully devs will barely notice the change :crossed_fingers: .

.flake8
benvand: Because we perform so much logic in our views, and because our codebase effectively dictates that's where the logic needs to live, it means our views already have high levels of complexity.
I'm sort of OK with that if we keep the complexity down everywhere else. I think it'd be good to have low complexity everywhere but it didn't look viable for a big bang PR.

.flake8
benvand: Yeah, good call.

requirements_for_test.txt
benvand: Alphabetisation FTW. There's a flake8 plugin that enforces alphabetisation in import statements 😨  Not sure I want to lump that on people all at once though :grin:.

PR #677, opened 2017-03-13
Change supplier declaration template column widths

samuelhwilliams:
We now have shorter questions in the supplier declaration, so we should reclaim the unused space and give it to the answers on the dashboard.

### Before:
screen shot 2017-03-13 at 15 36 24

### After:
screen shot 2017-03-13 at 15 33 48

https://trello.com/c/qoqrHUaq/290-make-declaration-summary-page-summary-table-first-column-1-third-width


TheDoubleK:
The answers are generally shorter than the questions though... Whatever, it's in now.

PR #679, opened 2017-03-15
139477629-Shortlisting-content-error

benvand:
As per ticket:
> Nicole has spotted an error in the copy on the 'what happens next' page.
We say that suppliers will provide feedback, when actually all they'll
do at shortlisting is tell a supplier they were not successful.

> It currently says:
'If you’re not successful, the buyer will give you feedback on the
evidence you provided.'

> #Could we please replace with:
'The buyer will tell you if you're not successful.'

https://www.pivotaltracker.com/story/show/139477629


idavidmcdonald:
👍

PR #681, opened 2017-03-16
Clarification around what we do with submitted service data

samuelhwilliams:
## Summary
Ticket: https://trello.com/c/RhoLFz3M/322-clarification-around-what-we-do-with-submitted-service-data

This PR pulls in an updated frameworks repo which contains content to be displayed on the frameworks dashboard explicitly clarifying what we do with any data provided about services submitted to a framework. This is currently available in supplier guides but is not present anywhere on DM pages themselves.

It also adds bullet styling for unordered lists within browse-list-items on the framework dashboard only.

Depends on https://github.com/alphagov/digitalmarketplace-frameworks/pull/408

## Before
screen shot 2017-03-16 at 11 59 36

## After
screen shot 2017-03-16 at 12 00 13

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -15,6 +15,7 @@ from dmutils.email import send_email
 from dmutils.email.exceptions import EmailError
 from dmcontent.formats import format_service_price
 from dmcontent.questions import ContentQuestion
+from dmcontent.errors import ContentNotFoundError
 from dmutils.formats import datetimeformat
 from dmutils import s3
 from dmutils.documents import (
@@ -83,6 +84,11 @@ def framework_dashboard(framework_slug):
 
     framework_dates = content_loader.get_message(framework_slug, 'dates')
     framework_urls = content_loader.get_message(framework_slug, 'urls')
+
+    try:
+        framework_advice = content_loader.get_message(framework_slug, 'advice')
+    except ContentNotFoundError:
+        framework_advice = None
 
     # filenames
     result_letter_filename = RESULT_LETTER_FILENAME
@@ -165,6 +171,7 @@ def framework_dashboard(framework_slug):
         result_letter_filename=result_letter_filename,
         supplier_framework=supplier_framework_info,
         supplier_is_on_framework=supplier_is_on_framework,
+        framework_advice=framework_advice,
     ), 200
 
 
--- a/app/templates/frameworks/_framework_actions.html
+++ b/app/templates/frameworks/_framework_actions.html
@@ -72,6 +72,11 @@
   <a class="browse-list-item-link" href="{{ url_for('.framework_submission_lots', framework_slug=framework.slug) }}">
     <span>Add, edit and complete services</span>
   </a>
+  {% if framework_advice.use_of_service_data %}
+    <div class="browse-list-item-body">
+      {{ framework_advice.use_of_service_data }}
+    </div>
+  {% endif %}
   {% if not counts.complete and not counts.draft %}
   <div class="browse-list-item-status-quiet">
     <p class="browse-list-item-status-title">You need to add and complete services</p>
--- a/app/templates/frameworks/_framework_actions.html
+++ b/app/templates/frameworks/_framework_actions.html
@@ -65,9 +65,7 @@
   </div>
   {% endif %}
 </li>
-{% endif %}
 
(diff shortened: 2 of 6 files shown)


TheDoubleK:
Not part of your changes, but it looks like there are two follow-on sections in that template with `{% if framework.status == 'open' %}` wrapping them.

Might not hurt to merge them into one while we're here?

TheDoubleK:
I worry that this message is G-Cloud specific and won't apply to other frameworks - e.g. for DOS-style frameworks there are no filters or service description pages.

So it might be better to add this to the digitalmarketplace-frameworks repo as a "message", and pull in the message here if one exists for the current framework. (Sorry)

TheDoubleK:
e.g. maybe in a new file here: https://github.com/alphagov/digitalmarketplace-frameworks/tree/master/frameworks/g-cloud-9/messages

I know it seems long-winded, but can't think of a better option off the top of my head.

samuelhwilliams:
Spoken to Roz and she agrees, so will go back and make it so.

Thanks!

app/main/views/frameworks.py
```diff
@@ -84,6 +85,11 @@ def framework_dashboard(framework_slug):
     framework_dates = content_loader.get_message(framework_slug, 'dates')
     framework_urls = content_loader.get_message(framework_slug, 'urls')
 
+    try:
+        use_of_data = content_loader.get_message(framework_slug, 'use-of-data')
```
galund: Probably super-picky, but the 'block' (according to `content_loader.ContentLoader#get_message`) is supposed to be the "place" where the message is used. Whereas the 'key' (if present) is the specific message.

So we have our naming wrong here, maybe. Probably doesn't matter.

TheDoubleK: I think the comment in the content loader is out of date since we introduced `urls` and `dates`.

The first messages was `homepage_sidebar`, which is the only thing that `block` makes sense for.

app/templates/frameworks/_framework_actions.html
```diff
@@ -65,13 +65,16 @@
...
   {% endif %}
 </li>
-{% endif %}
 
-{% if framework.status == 'open' %}
 <li class="browse-list-item">
   <a class="browse-list-item-link" href="{{ url_for('.framework_submission_lots', framework_slug=framework.slug) }}">
     <span>Add, edit and complete services</span>
   </a>
+  {% if use_of_data['services'] %}
```
TheDoubleK: Will this error if `use_of_data` is `None`?  Or does Jinja magically handle it?

bower.json
```diff
@@ -7,6 +7,6 @@
     "jquery-details": "https://github.com/mathiasbynens/jquery-details/archive/v0.1.0.tar.gz",
     "digitalmarketplace-frontend-toolkit": "https://github.com/alphagov/digitalmarketplace-frontend-toolkit.git#22.1.4",
     "govuk_template": "https://github.com/alphagov/govuk_template/releases/download/v0.19.2/jinja_govuk_template-0.19.2.tgz",
-    "digitalmarketplace-frameworks": "https://github.com/alphagov/digitalmarketplace-frameworks.git#6.4.8"
+    "digitalmarketplace-frameworks": "https://github.com/alphagov/digitalmarketplace-frameworks.git#shw-add-use-of-data-messages"
```
TheDoubleK: This will need to update to 6.4.9

app/templates/frameworks/_framework_actions.html
samuelhwilliams: Jinja magically handles it.

bower.json
samuelhwilliams: It will. Check the commit message. :)

TheDoubleK:
Looks good once frameworks version is fixed I think.

bower.json
TheDoubleK: Oh yeah :)

app/templates/frameworks/_framework_actions.html
galund: Does it work the same in Jinja to write `use_of_data.services`. That would seem more idiomatically 'templatey'.

TheDoubleK: Good call, I think that should work, yes.

PR #682, opened 2017-03-16
Update dev config to use a single bucket for all upload types

allait:
Using a single bucket allows us to use S3 bucket domain as our
assets URL without a running a separate nginx proxy.

The bucket is empty at the moment and might get replaced in the
future when we start managing it with Terraform.

### DIFF
--- a/config.py
+++ b/config.py
@@ -106,8 +106,6 @@ class Test(Config):
 
     SECRET_KEY = 'not_very_secret'
 
-    DM_SUBMISSIONS_BUCKET = 'digitalmarketplace-submissions-dev-dev'
-    DM_COMMUNICATIONS_BUCKET = 'digitalmarketplace-communications-dev-dev'
     DM_ASSETS_URL = 'http://asset-host'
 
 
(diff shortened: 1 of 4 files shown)


config.py
```diff
@@ -124,10 +122,10 @@ class Development(Config):
...
     DM_API_AUTH_TOKEN = "myToken"
 
-    DM_SUBMISSIONS_BUCKET = "digitalmarketplace-submissions-dev-dev"
-    DM_COMMUNICATIONS_BUCKET = "digitalmarketplace-communications-dev-dev"
-    DM_AGREEMENTS_BUCKET = "digitalmarketplace-agreements-dev-dev"
-    DM_DOCUMENTS_BUCKET = "digitalmarketplace-documents-dev-dev"
+    DM_SUBMISSIONS_BUCKET = "digitalmarketplace-dev-uploads"
+    DM_COMMUNICATIONS_BUCKET = "digitalmarketplace-dev-uploads"
+    DM_AGREEMENTS_BUCKET = "digitalmarketplace-dev-uploads"
+    DM_DOCUMENTS_BUCKET = "digitalmarketplace-dev-uploads"
```
samuelhwilliams: Is there any reason for still having these as separate variables?

pcraig3:
Seems 👍  to me.

PR #688, opened 2017-03-28
Updated frontend toolkit fixes multiquestion layout issue.

galund:
https://trello.com/c/WmvbdJSR/228-borders-for-follow-up-questions-don-t-line-up

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -5,7 +5,7 @@
     "jquery": "1.11.2",
     "hogan": "3.0.2",
     "jquery-details": "https://github.com/mathiasbynens/jquery-details/archive/v0.1.0.tar.gz",
-    "digitalmarketplace-frontend-toolkit": "https://github.com/alphagov/digitalmarketplace-frontend-toolkit.git#22.1.4",
+    "digitalmarketplace-frontend-toolkit": "https://github.com/alphagov/digitalmarketplace-frontend-toolkit.git#22.1.5",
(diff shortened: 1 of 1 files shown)


samuelhwilliams:
This looks better on blank/fresh pages, but there seems to be a bug when questions get validated that exhibits the original problem of lines being disconnected.

Also the spacing seems to vary a bit between certain questions, and in one of the instances pictured below looks a bit off. Probably can take or leave it either way, though.

All in all, 👍, but a few bits to deal with in a followup PR.

## Good example
screen shot 2017-03-28 at 20 14 20

## Inconsistent spacing
screen shot 2017-03-28 at 20 17 16

## Disconnected with validated questions
screen shot 2017-03-28 at 20 13 53

## Inconsistent disconnected between validated questions
screen shot 2017-03-28 at 20 14 32

pcraig3:
Yeah, I've just noticed this as well.  

The validation thing is a bug that I think we didn't consider. 

The problem of the different spacing between questions has to do with the question type we're using.
List-entry questions have an "add an extra field" button underneath and, as in your example, textareas with word counters I would suspect have extra spacing under them also. 

It seems like those are two different issues: 
1. fixing the disconnected red borders problem
2. removing extra (top/bottom) spacing from question types that are `.related-information`

Probably the first one is more important than the second one.

pcraig3:
However, these changes are still an improvement over what was there before, and [the frontend toolkit has just been re-incremented with another unrelated bugfix](https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/689), so a 👍  from me as well.

PR #691, opened 2017-03-30
[#142242977]fixes bug when the declaration is None instead of empty dict

MuriloDalRi:
Story: https://www.pivotaltracker.com/story/show/142242977


TheDoubleK:
This looks good -  :+1: after a rebase.

It might be worth doing a quick search of the code for `.get('declaration')` to see if this could be an issue elsewhere at all too.

PR #692, opened 2017-04-03
Add documentation about login

carolinegreen:
I don't think it's immediately obvious that this is how the applications are split up.  Also add documentation about nginx configuration for supplier frontend.  I realise this is in digital marketplace manual too, but this is in a private repo so I think it's useful to have it here in the readme too.


TheDoubleK:
Yep makes sense :+1:

PR #693, opened 2017-04-12
Fixes validation masthead on supplier creation

MuriloDalRi:
Story: https://www.pivotaltracker.com/story/show/143115633

Changes template to show masthead if errors >= 1 instead of > 1

### DIFF
--- a/app/templates/suppliers/company_contact_details.html
+++ b/app/templates/suppliers/company_contact_details.html
@@ -17,7 +17,7 @@ with items = [
 
 {% block main_content %}
 
-{% if form.errors|length > 1 %}
+{% if form.errors|length >= 1 %}
     <div class="validation-masthead" role="group" aria-labelledby="validation-masthead-heading">
         <h1 class="validation-masthead-heading" id="validation-masthead-heading">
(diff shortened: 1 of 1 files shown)


risicle:
This is a bit of a suspicious one. When I see code like that (`{% if form.errors|length > 1 %}`) I immediately wonder what made them do that over just `{% if form.errors %}` if all they wanted to do was check if there are *any* errors. So it implies there's a *reason* they didn't want to emit this block in the case of having just a single error.

Digging back into history the explanation/culprit seems to be a9528c5d7c3e6b919f66ef1ee89ed9a387fab7c6

```
This is the only page in the journey that has more
than 1 control so needs a summary
```

So the thinking was that no summary was needed if there's just a single error.

So I would consider this Not A Bug. It was a deliberate decision. Whether we want to revise that decision is another question...

MuriloDalRi:
You're right, it's not really a bug. Is there any reason to change this behaviour or should I delete this and other PRs related to the same code?

risicle:
I say we leave it hanging around until we can get a designer to make a decision...

PR #694, opened 2017-04-12
[#140222101]declaration overview: enable visibility in non-open framework statuses

risicle:
Story https://www.pivotaltracker.com/story/show/140222101

In the story it was decided to cope with all `complete` declarations for non-open frameworks and ignore whether a supplier was `onFramework` or not. Suppliers with non-`complete` declarations for a framework will receive a `410` (see code for reasoning).

One consequence of this is that non-open frameworks end up with no introductory text or paragraph on this page.

~~~Works back to G7 though in an ugly way because pre-G9 questions didn't tend to have "short" question text variants to use in the table. (screenshots coming...)~~~ ~~~~doesn't appear to work properly on G7, will need to look at this...~~~~ ok it does seem to work, it was just the mock declaration that was wrong...

~~~One thing still TBD is where/how this page will get linked to from the framework dashboard in non-open states. Currently it just isn't (but navigating to the url manually will allow a user to see this page, which could still be useful to allow an admin to direct a user to this page if there's some dispute...)~~~ @MuriloDalRi  has done this now

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -404,7 +404,10 @@ def framework_supplier_declaration_overview(framework_slug):
         # no longer can, so it's apparently "gone"
         abort(410)
 
-    content = content_loader.get_manifest(framework_slug, 'declaration').filter(sf["declaration"])
+    try:
+        content = content_loader.get_manifest(framework_slug, 'declaration').filter(sf["declaration"])
+    except ContentNotFoundError:
+        abort(404)
 
     # generate an (ordered) dict of the form {section_slug: (section, section_errors)}.
     # we must perform an actual validation for each section rather than rely on .answer_required as the latter won't
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -387,11 +387,22 @@ def reuse_framework_supplier_declaration_post(framework_slug):
 @main.route('/frameworks/<framework_slug>/declaration', methods=['GET'])
 @login_required
 def framework_supplier_declaration_overview(framework_slug):
-    framework = get_framework(data_api_client, framework_slug, allowed_statuses=['open'])
+    framework = get_framework(data_api_client, framework_slug, allowed_statuses=[
+        "open",
+        "pending",
+        "standstill",
+        "live",
+        "expired",
+    ])
 
     sf = data_api_client.get_supplier_framework_info(current_user.supplier_id, framework_slug)["frameworkInterest"]
     # ensure our declaration is a a dict
     sf["declaration"] = sf.get("declaration") or {}
+
+    if framework["status"] != "open" and sf["declaration"].get("status") != "complete":
+        # 410 - the thinking here is that the user probably *used to* be able to access a page at this url but they
+        # no longer can, so it's apparently "gone"
+        abort(410)
 
     content = content_loader.get_manifest(framework_slug, 'declaration').filter(sf["declaration"])
 
--- a/app/templates/frameworks/_submitted_services.html
+++ b/app/templates/frameworks/_submitted_services.html
@@ -13,8 +13,15 @@
       {% endif %}
     {% endfor %}
   </ul>
-  <a href="{{ url_for('.framework_submission_lots', framework_slug=framework.slug) }}">
-    View submitted services
-  </a>
+  <p>
+    <a href="{{ url_for('.framework_submission_lots', framework_slug=framework.slug) }}">
+      View submitted services
+    </a>
+  </p>
+  <p>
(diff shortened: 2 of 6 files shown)


TheDoubleK:
It might be nice to have conditional breadcrumb text once the framework is in pending/standstill/live - it still says "Apply to <framework>":
screen shot 2017-04-25 at 10 26 57
screen shot 2017-04-25 at 10 26 50

But the page it links to has the title changed to "Your <framework> application", which could be the same for the breadcrumb here.

app/templates/frameworks/_submitted_services.html
```diff
@@ -13,8 +13,15 @@
...
-    View submitted services
-  </a>
+  <p>
+    <a href="{{ url_for('.framework_submission_lots', framework_slug=framework.slug) }}">
+      View submitted services
+    </a>
+  </p>
+  <p>
+    <a href="{{ url_for('.framework_supplier_declaration_overview', framework_slug=framework.slug) }}">
+      View your declaration
```
TheDoubleK: There is a weird case if the supplier is marked on_framework but with an incomplete declaration - we'd show the link but it would 410 - but this should really really never happen. I can't think of any way we'd have a supplier on_framework without having completed their declaration - incomplete declaration means they didn't apply in the first place, so this should be fine I reckon.

tests/app/main/test_services.py
```diff
@@ -26,9 +24,7 @@ def supplier_service_editing_fw_params(request):
     return request.param
 
 
-# and these fixtures really are just a workaround to not being able to use plain function parametrization in unittest-
-# derived classes and should be replaced with that as soon as we're able to do so. hence the hence the oddly specific
-# long names
+# find a better way of doing ofganizing this than using oddly specific long names
```
TheDoubleK: I'm not mad on this comment, but if we have it in we could at least fix `ofganizing`

risicle: Are you suggesting you want our project to be totally unofganized?

TheDoubleK:
Other than the breadcrumbs thing I think this is good.  And we can probably live with them if we have to.

app/templates/frameworks/_submitted_services.html
risicle: I think failing with a 410 isn't the worst thing in the world in that case.

risicle:
Yes I think we should do something about the breadcrumb... I'd rather do something more general than a little ad hoc fix here, but... scope creep...

risicle:
Still need to fix example g7 declaration for this...

app/templates/frameworks/declaration_overview.html
```diff
@@ -17,7 +17,7 @@
       },
       {
         "link": url_for(".framework_dashboard", framework_slug=framework.slug),
-        "label": "Apply to " + framework.name
+        "label": ("Apply to "+framework.name) if framework.status == "open" else ("Your "+framework.name+" application")
```
TheDoubleK: If PEP8 checked Jinja templates it would tell you to put spaces around the `+` operators here.

app/main/views/frameworks.py
```diff
@@ -387,12 +387,23 @@ def reuse_framework_supplier_declaration_post(framework_slug):
...
     sf = data_api_client.get_supplier_framework_info(current_user.supplier_id, framework_slug)["frameworkInterest"]
     # ensure our declaration is a a dict
     sf["declaration"] = sf.get("declaration") or {}
 
+    if framework["status"] != "open" and sf["declaration"].get("status") != "complete":
+        # 410 - the thinking here is that the user probably *used to* be able to access a page at this url but they
+        # no longer can, so it's apparently "gone"
+        abort(410)
+
     content = content_loader.get_manifest(framework_slug, 'declaration').filter(sf["declaration"])
```
TheDoubleK: This will generate a 500 error if someone tries the URL for, e.g. `g-cloud-6` declaration.

Would be nice to `try...except` and 404 if there's no content.
```
        try:
             ...
        except ContentNotFoundError as e:
             abort(404)
```

risicle: But do we even have g6 in the `frameworks` table?

TheDoubleK: Yep.  Also G4, G5...

TheDoubleK: And there are entries in the `supplier_frameworks` table for suppliers who were on G6, even though we didn't do the declaration for that framework. (I think we back-filled them with a migration so that new features would work for existing G6 suppliers.)

TheDoubleK:
Couple of minor final comments but this looks good :+1: after a rebase, with or without suggested changes.

app/main/views/frameworks.py
risicle: Damn and blast it all...

PR #701, opened 2017-05-02
Change "Pass" to "Successful" and deal with failed lots

TheDoubleK:
For this story: https://www.pivotaltracker.com/story/show/144514369

Suppliers prefer the word "Successful" to "Pass", and it matches what we say in the emails to them.

Also, turns out the page was not dealing with DOS-style frameworks properly and showing failed lots as "No application".  This fixes it so that if a lot was submitted but failed it will show "Unsuccessul" rather than "No application".

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -750,10 +750,20 @@ def framework_agreement(framework_slug):
 
     # if there's a frameworkAgreementVersion key, it means we're on G-Cloud 8 or higher
     if framework.get('frameworkAgreementVersion'):
+        def lot_result(drafts_for_lot):
+            if any(draft['status'] == 'submitted' for draft in drafts_for_lot):
+                return 'Successful'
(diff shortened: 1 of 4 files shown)


app/main/views/frameworks.py
```diff
@@ -750,10 +750,20 @@ def framework_agreement(framework_slug):
 
     # if there's a frameworkAgreementVersion key, it means we're on G-Cloud 8 or higher
     if framework.get('frameworkAgreementVersion'):
+        def lot_result(drafts_for_lot):
+            if any(draft['status'] == 'submitted' for draft in drafts_for_lot):
+                return 'Successful'
+            elif any(draft['status'] == 'failed' for draft in drafts_for_lot):
+                return 'Unsuccessful'
+            else:
+                return 'No application'
```
risicle: Point of note that this is "content-y" text going into the view...

TheDoubleK: That's true.  I can't see a particularly neat way round it though, so going to suck it up.

PR #714, opened 2017-06-14
Allow suppliers to view their applications

benvand:
User need
As a supplier, I want to see the opportunities I've applied for, so I can keep track of them

Add page suppliers/framework-slug/opportunities-overview to allow suppliers to view their applications to opportunities.
Add link to this page from supplier dash and from 'application complete' page
Template and view for this page
Tests
Small test refactor in tests/app/main/test_suppliers.py to add mocks to class instead of every method

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -1203,7 +1203,7 @@ def opportunities_dashboard(framework_slug):
     if not (framework['framework'] == 'digital-outcomes-and-specialists' and supplier_framework['onFramework']):
         abort(404)
     opportunities = data_api_client.find_brief_responses(
-        supplier_id=current_user.supplier_id,
+        supplier_id=current_user.supplier_id, framework=framework_slug
     )['briefResponses']
 
     return render_template(
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -1187,3 +1187,27 @@ def view_contract_variation(framework_slug, variation_slug):
         agreed_details=agreed_details,
         supplier_name=supplier_name,
     ), 400 if form_errors else 200
+
+
+@main.route('/frameworks/<framework_slug>/opportunities', methods=['GET'])
+@login_required
+def opportunities_dashboard(framework_slug):
+    try:
+        framework = data_api_client.get_framework(slug=framework_slug)['frameworks']
+        supplier_framework = data_api_client.get_supplier_framework_info(
+            supplier_id=current_user.supplier_id,
+            framework_slug=framework['slug']
+        )['frameworkInterest']
+    except APIError as e:
+        abort(e.status_code)
+    if not (framework['framework'] == 'digital-outcomes-and-specialists' and supplier_framework['onFramework']):
+        abort(404)
+    opportunities = data_api_client.find_brief_responses(
+        supplier_id=current_user.supplier_id,
+    )['briefResponses']
+
+    return render_template(
+        "suppliers/opportunities_dashboard.html",
+        framework=framework,
+        completed=opportunities
+    ), 200
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -63,7 +63,6 @@ def dashboard():
                 framework.get('onFramework') and framework.get('agreementReturned') is False
             )
         })
-
     return render_template(
         "suppliers/dashboard.html",
         supplier=supplier,
--- a/app/templates/briefs/view_response_result.html
+++ b/app/templates/briefs/view_response_result.html
@@ -148,8 +148,8 @@
 
   {%
     with
-      url = url_for(".dashboard"),
-      text = "Your account"
+      url = url_for(".opportunities_dashboard", framework_slug=brief.frameworkSlug),
+      text = "Your {} applications".format(brief.frameworkName)
   %}
     {% include "toolkit/secondary-action-link.html" %}
   {% endwith %}
--- a/app/templates/suppliers/_frameworks_live.html
+++ b/app/templates/suppliers/_frameworks_live.html
@@ -2,7 +2,7 @@
 
 {{ summary.heading("Current services") }}
 {% if frameworks.live %}
-  {{ summary.top_link('View', url_for('.list_services')) }}
+  {{ summary.top_link('View services', url_for('.list_services')) }}
 {% endif %}
 {% call(framework) summary.list_table(
   frameworks.live,
@@ -29,8 +29,14 @@
         </p>
       {% endif %}
     {% endcall %}
+
     {% if framework.onFramework and not framework.needs_to_complete_declaration %}
-      {{ summary.edit_link('View documents and ask a question', url_for('.framework_dashboard', framework_slug=framework.slug)) }}
+        {% call summary.field() %}
+            {% if framework.framework == 'digital-outcomes-and-specialists' %}
+                <a href="{{ url_for('.opportunities_dashboard', framework_slug=framework.slug) }}">View your opportunities</a><br>
+            {% endif %}
+            <a href="{{ url_for('.framework_dashboard', framework_slug=framework.slug) }}">View documents and ask a question</a>
+        {% endcall %}
     {% else %}
       {% call summary.field() %}
       {% endcall %}
--- a/app/templates/suppliers/_frameworks_live.html
+++ b/app/templates/suppliers/_frameworks_live.html
@@ -31,7 +31,7 @@
     {% endcall %}
 
     {% if framework.onFramework and not framework.needs_to_complete_declaration %}
-        {% call summary.field() %}
+        {% call summary.field(action=True) %}
             {% if framework.framework == 'digital-outcomes-and-specialists' %}
                 <a href="{{ url_for('.opportunities_dashboard', framework_slug=framework.slug) }}">View your opportunities</a><br>
             {% endif %}
--- a/tests/app/main/test_suppliers.py
+++ b/tests/app/main/test_suppliers.py
@@ -1,8 +1,8 @@
 # coding=utf-8
-
-import mock
 from flask import session
 from lxml import html
+import mock
+import pytest
 
 from dmapiclient import HTTPError
 from dmutils.email.exceptions import EmailError
@@ -59,9 +59,10 @@ def get_user():
     }]
 
 
+@mock.patch("app.main.views.suppliers.data_api_client", autospec=True)
+@mock.patch("app.main.views.suppliers.get_current_suppliers_users", autospec=True)
 class TestSuppliersDashboard(BaseApplicationTest):
-    @mock.patch("app.main.views.suppliers.data_api_client")
-    @mock.patch("app.main.views.suppliers.get_current_suppliers_users")
+
     def test_shows_supplier_info(self, get_current_suppliers_users, data_api_client):
         data_api_client.get_framework.return_value = self.framework('open')
         data_api_client.get_supplier.side_effect = get_supplier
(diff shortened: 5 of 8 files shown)


tests/app/main/test_suppliers.py
```diff
@@ -59,9 +61,10 @@ def get_user():
     }]
 
 
+@mock.patch("app.main.views.suppliers.data_api_client", autospec=True)
+@mock.patch("app.main.views.suppliers.get_current_suppliers_users", autospec=True)
 class TestSuppliersDashboard(BaseApplicationTest):
-    @mock.patch("app.main.views.suppliers.data_api_client")
-    @mock.patch("app.main.views.suppliers.get_current_suppliers_users")
```
samuelhwilliams: Nice cleaning up.

samuelhwilliams: Nice cleaning up.

tests/app/main/test_suppliers.py
```diff
@@ -1531,3 +1567,136 @@ def test_should_show_email_address_even_when_refreshed(self):
...
+
+            doc = html.fromstring(res.get_data(as_text=True))
+
+            table_caption = "Completed opportunities"
+            list_of_tables = doc.find_class('summary-item-body')
+
+            table = [i for i in list_of_tables if table_caption in i.xpath('string()')][0]
+            first_row, second_row, third_row = table.find_class('summary-item-row')
+
+            assert res.status_code == 200
```
samuelhwilliams: If this is worth doing, it's probably better straight after the get.

app/main/views/suppliers.py
```diff
@@ -444,3 +443,28 @@ def create_your_account_complete():
         "suppliers/create_your_account_complete.html",
         email_address=email_address
     ), 200
+
+
+@main.route('/<framework_slug>/opportunities-overview', methods=['GET'])
```
TheDoubleK: This seems a weird place and URL for this page.

Might be better in `briefs.py` where all of the other `/opportunities` routes are and live under something like `/opportunities/overview`?

Or even in `frameworks.py` where all the '/frameworks/<framework_slug>' routes are and live under something like `'/frameworks/<framework_slug>/opportunities-overview'?

And I know Roz has gone now, but channelling her spirit, I'm not sure "overview" is a good word to have in a URL at all, so just `/frameworks/<framework_slug>/opportunities`. Would be fine here (because this is the supplier app the full URL would end up being `/suppliers/frameworks/<framework_slug>/opportunities`, which makes sense - i.e. the opportunities of interest to X supplier for Y framework).

TheDoubleK: And if all this stuff is going to be moving to the new briefs app then `briefs.py` might be the natural home for it for now, as I assume those are the views being ported across.

I'm not sure how we're going to handle the links from the dashboard though once apps are split - we're going to lose the ability to use `url_for` and probably end up hard-coding routes, which is a real shame.

app/main/views/suppliers.py
```diff
@@ -444,3 +443,28 @@ def create_your_account_complete():
...
+@login_required
+def opportunities_overview(framework_slug):
```
TheDoubleK: I might be missing something but why does this need to take framework as a param?  I can see it makes sense from the point of view of where the link is on the dashboard, but I can't think of any other reason to key on framework?  i.e. if we just had a "View my opportunities" link from the dashboard, not in the frameworks list, and showed a list of all brief responses on whatever framework that would probably work?

As a supplier I just want to see all the things I applied for, I don't particularly care what framework it was on?

app/main/views/suppliers.py
TheDoubleK: Thinking about this more the most natural place for this page seems to me to be in `briefs.py` and with the URL `/opportunities` - I really don't think we need the word overview in there.

app/templates/briefs/view_response_result.html
```diff
@@ -148,8 +148,8 @@ <h2 class="summary-item-heading">What happens next</h2>
 
   {%
     with
-      url = url_for(".dashboard"),
-      text = "Your account"
+      url = url_for(".opportunities_overview", framework_slug=brief.frameworkSlug),
```
TheDoubleK: And then in the code we could call it `opportunities_dashboard`, to mirror the `dashboard` and `framework_dashboard` that we already have.

TheDoubleK: But we should never make the work "dashboard" visible on the site or in any URLs.

app/templates/suppliers/_frameworks_live.html
```diff
@@ -29,8 +29,14 @@
         </p>
       {% endif %}
     {% endcall %}
+
     {% if framework.onFramework and not framework.needs_to_complete_declaration %}
-      {{ summary.edit_link('View documents and ask a question', url_for('.framework_dashboard', framework_slug=framework.slug)) }}
+        {% call summary.field() %}
+            {% if framework.framework == 'digital-outcomes-and-specialists' %}
+                <a href="{{ url_for('.opportunities_overview', framework_slug=framework.slug) }}">View your opportunities</a><br>
```
TheDoubleK: "View your opportunities" feels a bit vague - it would be nice to be more precise that these are the opportunities that you have previously applied for, because I don't think that's obvious from the link text.

And if suppliers click this and see an empty list - as many will if they haven't applied for work recently - it won't be obvious why there's nothing there for them.

app/templates/suppliers/opportunities_overview.html
```diff
@@ -0,0 +1,63 @@
...
+    ],
+    field_headings_visible=True
+) %}
+
+    {% call summary.row() %}
+        {% call summary.field() %}
+            <a href={{ "/{}/opportunities/{}".format(framework.framework, item.briefId) }}>{{ item.brief.title }}</a>
+        {% endcall %}
+        {{ summary.text(item.brief.applicationsClosedAt|dateformat) }}
+        {{ summary.link("View application", url_for(".view_response_result", brief_id=item.briefId)) if not item.brief.status == 'withdrawn' else summary.text() }}
```
TheDoubleK: Would it be helpful to have some text to show that it was withdrawn, rather than just having no link?  I was confused looking at the screenshot why one had a link and the other didn't.

app/templates/suppliers/opportunities_overview.html
```diff
@@ -0,0 +1,63 @@
...
+
+{{ summary.heading("Completed", id="completed-opportunities") }}
+{% call(item) summary.list_table(
+    completed|sort(attribute='submittedAt', reverse=True),
+    caption="Completed opportunities",
+    empty_message="You haven’t applied to any opportunities",
+    field_headings=[
+        "Opportunity",
+        "Application deadline",
+        summary.hidden_field_heading("Remove")
```
TheDoubleK: Why "Remove" if the column actually has "View application" links?  I think the hidden heading is to help people using a screen reader?  If so, this "Remove" will be confusing.

benvand:
@TheDoubleK Will have to chat to the team and get back to you. 
I'm OK with moving the code but that only makes sense if we change the URL. Changing the URL only makes sense if we move the 'View your opportunities' link out of the frameworks table/ list.

Will need to discuss with content and design the following:
Should we move the  'View your opportunities' link out of the frameworks table/ list
Should we change the wording of the 'View your opportunities' link
Should we remove 'overview' from the URL
Should we remove the need for a URL param
Should we have a placeholder to show an opportunity was withdrawn

TheDoubleK:
The URL definitely needs to change one way or another, you can't just plonk `<framework_slug>` at the start of the route.  The way our URLs work is that if you have a parameter like `<framework_slug>` in the URL then it is preceded by the thing it references - i.e. our URLs are always, e.g.:
```
/frameworks/<framework_slug>/...
/services/<service_id>/...
/users/<user_id>/...
/opportunities/<brief_id>/...
```
screen shot 2017-06-15 at 17 44 04

TheDoubleK:
And I think `suppliers.py` is definitely the wrong place for this view code, whether or not the 'View your opportunities' link is in the the frameworks table/list is irrelevant.

app/main/views/suppliers.py
idavidmcdonald: I can't remember exactly why but I think it came out of discussion with @laurenceberry? Potentially something to do with how the supplier page is going to be designed in the future and this is a step towards that.

app/templates/briefs/view_response_result.html
idavidmcdonald: Reasonable suggestion

idavidmcdonald:
I think you are right Kev and we should change the URL as it stands.

If we keep the page to just show opportunities for one framework then I would suggest `frameworks/<framework_slug>/opportunities`.

If we show opportunities for all frameworks then yes probably a route in in briefs.py with `/opportunities` or similar.

As I mentioned in an earlier comment, I think there is a reason based on forward planning for design that we have structured it per framework but this is definitely something worth double checking.

Apart from that, technicallly things in this PR look good. But yes, I think chat with the team @benvand and we can come up with a decision regarding URL and copy changes. Give me a shout tomorrow and I can join you if that would be useful.

app/templates/suppliers/opportunities_overview.html
benvand: This from Tobi:
> Putting the text withdrawn is confusing because it’s does not clearly state who withdrew it or what was withdrawn. Since the other link goes to response, suppliers might be think their application was withdrawn

To further the above the page for the opportunity (currently linked to from the brief title) will contain the withdrawn banner. 
In the future there will be a link in the empty section but we need to decide on what to do with the brief response page itself before we link to it because at the moment it has all of the 'What happens next' text on it which won't apply to a withdrawn brief.

app/templates/suppliers/_frameworks_live.html
benvand: This went through content. The empty text should be 'You haven't applied for any opportunities' So that will indicate what the page is for.

app/main/views/suppliers.py
benvand: Yeah, going forward we'll be adding various links on this page and it was decided that splitting it out by framework would work with a view to going forward, adding frameworks, reusability etc.

app/templates/suppliers/opportunities_overview.html
TheDoubleK: I'm sure with a little thought we could find some words that would clearly convey what has happened in the case of a withdrawn brief, but if we don't want to figure them out right now then fine.

app/templates/suppliers/_frameworks_live.html
TheDoubleK: But you have to click through to find that out...

app/main/views/frameworks.py
```diff
@@ -1187,3 +1187,27 @@ def view_contract_variation(framework_slug, variation_slug):
...
+        framework = data_api_client.get_framework(slug=framework_slug)['frameworks']
+        supplier_framework = data_api_client.get_supplier_framework_info(
+            supplier_id=current_user.supplier_id,
+            framework_slug=framework['slug']
+        )['frameworkInterest']
+    except APIError as e:
+        abort(e.status_code)
+    if not (framework['framework'] == 'digital-outcomes-and-specialists' and supplier_framework['onFramework']):
+        abort(404)
+    opportunities = data_api_client.find_brief_responses(
```
TheDoubleK: This list of opportunities needs to be filtered by the specific framework.  As this is the following URLs show me the same list of things - both DOS and DOS2 opportunities:
```
http://localhost/suppliers/frameworks/digital-outcomes-and-specialists/opportunities
http://localhost/suppliers/frameworks/digital-outcomes-and-specialists-2/opportunities
```

(There should probably be a test for this - i.e. that checks only opportunities in the framework explicitly mentioned in the URL are shown.)

idavidmcdonald: Great spot, will fix!

TheDoubleK:
If you're still waiting for me on this at all, apologies! I'm pretty sure all my comments have been addressed now so go for it once you're happy.

TheDoubleK:
(...and Travis is happy)

PR #715, opened 2017-06-14
Store `supplierId` in brief clarification question audit event

pcraig3:
When a user asks a clarification question on a brief, we want to be able to be able to easily find the supplier the user belongs to so that we can notify them of changes to the brief.

We're pretty lax about what goes into audit events, so it seems like a good idea to keep track of the supplier ID in the audit event created when a clarification question is asked about a brief.

### DIFF
--- a/app/main/helpers/briefs.py
+++ b/app/main/helpers/briefs.py
@@ -65,7 +65,7 @@ def send_brief_clarification_question(data_api_client, brief, clarification_ques
         user=current_user.email_address,
         object_type="briefs",
         object_id=brief['id'],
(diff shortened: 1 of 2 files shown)


TheDoubleK:
Should this change go into the new briefs app? I think brief responses is moving there so changes made here now might not make it across to the new app.  Or is that a way off still?

pcraig3:
There's no brief responses app until after the firebreak, @TheDoubleK, so it seems like put this in now.

TheDoubleK:
OK cool, looks legit then :+1:

PR #716, opened 2017-06-20
Bring in frontend toolkit updates

carolinegreen:
Bring in Digital Marketplace frontend toolkit with the new buttons that work without javascript.  This was a breaking change, so GOV.UK Elements is now a dependency and there is new build process reflecting that.

Additionally the category picker was slightly altered to handle the new markup.

Relies on: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/353

### DIFF
--- a/app/assets/javascripts/category-picker.js
+++ b/app/assets/javascripts/category-picker.js
@@ -100,7 +100,6 @@
       }
     })
     var globalCategories = new Categories(categories_data)
-
     function appendCounter(id) {
        var $_counter = $('#' + id)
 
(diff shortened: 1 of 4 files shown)


galund:
Reopening because we need these changes on the buyer frontend, in order to be able to bring in any further changes to the toolkit.

TheDoubleK:
OK I've had a look locally at:
* supplier sign up and editing 
* Completing a G9 declaration
* Completing a G9 service submission (including the category picker)

and everything seems fine as far as I can see.  Going to merge this once Travis finishes running all the things.

PR #717, opened 2017-06-20
Do not show remove service confirmation message for DOS

idavidmcdonald:
Bug as spotted by Andrew - https://trello.com/c/iptxy2Oy/397-allow-dos-suppliers-to-view-their-services

If the user 'hacked' the url and added `remove_requested=True` the url of a DOS service they should not see the removal confirmation that you usually would for GCloud services.

We decided to not show the message rather than 404 as it's more forgiving to a user.

Before picture:
image

After picture:
image

### DIFF
--- a/tests/app/main/test_services.py
+++ b/tests/app/main/test_services.py
@@ -399,7 +399,7 @@ class TestSupplierEditDosServices(SupplierEditServiceTestsSharedAcrossFrameworks
             **self.framework_kwargs
         )
         res = self.client.get('/suppliers/services/123')
-
+        assert res.status_code == 200
         assert 'View service page on the Digital Marketplace' not in res.get_data(as_text=True)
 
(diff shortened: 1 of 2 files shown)


tests/app/main/test_services.py
```diff
@@ -410,12 +410,26 @@ def test_no_remove_service_section(self, data_api_client):
...
+    def test_no_remove_service_confirmation_prompt(self, data_api_client):
+        self.login()
+        self._setup_service(
+            data_api_client,
+            service_status='published',
+            **self.framework_kwargs
+        )
+        res = self.client.get('/suppliers/services/123?remove_requested=True')
+        assert res.status_code == 200
+        self.assert_not_in_strip_whitespace(
```
risicle: Personally don't like this method but I realize you're just fitting in...

PR #719, opened 2017-07-05
Add default.nix

risicle:
This is something that's going to need extensively more of a writeup than this, but...

_Note: this is meant to be a strictly optional convenience for those of us who want to use it_

This `default.nix` allows users with a working nix install to simply perform a

```
$ nix-shell --pure .
```

in the repository root and get a fully working development environment with all dependencies.

(you may have to blow away your `node_modules/` and `bower_components/` directories before running `make run_all` to get them re-fetched again as they will have been constructed using a different node version)

You can choose the python version by editing the line

```
    pythonPackages = pkgs.python36Packages;
```

to, e.g., `pkgs.python27Packages`. I've set up `nix-shell` to use a separate virtualenv for each python version so you should be able to drop out & back in to different python versions without any extra fuss.

## Caveats

You'll notice that if you use the `--pure` flag with `nix-shell`, it gives you a really very pure environment - it won't even put `vi` in your environment. In fact, nothing from your underlying system installation. This is a great way of ensuring your build *is* actually pure and reproducible, but could indeed be quite annoying for day to day use. There are numerous approaches that could be taken to make this easier. It's up to you which one you prefer. Some suggestions:

* Don't use the `--pure` flag and live with the danger of mixed up libraries. This will allow you to continue using your underlying installed tools.
* Only use `nix-shell` for execution of the actual project commands, either dropping out of `nix-shell` whenever you're not using the project commands, or leaving another shell active in the same directory for performing those commands. This is *probably* the approach I'll take.
* Add your tool of choice to the dependencies using a `local.nix` based on the example at `local.example.nix`, causing nix to provide this tool for you. `local.nix` can be used to personally customize the shell environment in almost any way to your own taste.

Something to note is that this is not a *full* nixification - it just creates a virtualenv and calls pip at the last stage of its initialization. This means that the installed python modules aren't pure in the same way proper nix packages are. The worst effect this should have is maybe occasionally having to blow away your `venv*` dirs after you've updated your nix channels.

The `$PS1` I have provided may not be to everyone's taste. I tried to include a bunch of useful information in it, including a `shortName` for the project which some may find objectionable (so that it's clear if you've absent-mindedly `cd`'d to a different project directory while remaining in a `nix-shell`. As shown in the example, this can also be personalized in `local.example.nix`, or if you like we could sit down and bikeshed what to make the default one...

Mac users will, for now, have to comment out the `watchdog` dependency in `requirements_for_test.txt` as it appears to need some impure dependencies to access system libraries (?). But there are possibly a few workarounds for this that can be discussed. The first that comes to mind is adding the nix package for watchdog (`pythonPackages.watchdog`, which, strangely enough, *does* work fine on macos) to your `local.nix`.

*Some* Mac users (looking at you @Wynndow) have trouble running the frontend tests and node doesn't make it particularly easy to discover the reason for this. WIP.

...probably lots of stuff that I haven't mentioned, but I'd like to get this PR posted at least...

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -5,7 +5,7 @@
 # E402: module level import not at top of file
 # C901: too complex
 # W503: line break before binary operator
-exclude = venv,venv3,__pycache__,node_modules,bower_components
+exclude = venv*,__pycache__,node_modules,bower_components
 ignore = D203,W503
 max-complexity = 7
 max-line-length = 120
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-venv/
+venv*
 .DS_Store
 
 # Byte-compiled / optimized / DLL files
--- a/.gitignore
+++ b/.gitignore
@@ -67,3 +67,6 @@ app/content
 
 # Front end generated assets
 app/static
+
+# local environment customizations
+local.nix
--- a/default.nix
+++ b/default.nix
@@ -11,6 +11,8 @@
       pkgs.nodejs
       pkgs.libffi
       pkgs.libyaml
+      # pip requires git to fetch some of its dependencies
+      pkgs.git
       # for `cryptography`
       pkgs.openssl
     ] ++ pkgs.stdenv.lib.optionals forDev ([
--- a/default.nix
+++ b/default.nix
@@ -43,7 +43,7 @@ in (with args; {
     LANG="en_GB.UTF-8";
 
     shellHook = ''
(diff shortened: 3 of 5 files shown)


allait:
@risicle I think if we're going to add this to the apps we need to document how to use it (since the PR description is not in the commit messages, and I don't think this PR will be easy to find, especially from other repos). If the process is the same between all apps it probably makes sense to add the instructions to the manual. We could also move the example local config there since it appears to be the same between the apps.

risicle:
Yes, this is what I was going to do but I was told that firebreak stuff gets halted after the sprint, so now I'm just dashing to turn these into PRs and will just have to write the proper docs after hours.

risicle:
See https://github.com/alphagov/digitalmarketplace-manual/pull/6 for proposed corresponding manual page.

TheDoubleK:
This worked for me as well as I can tell.  Let's get these in :+1:

risicle:
Hooray!

PR #722, opened 2017-07-11
Return 403 status if supplier ineligible for brief

katstevens:
Simple fix to change the status code returned when a supplier is ineligible for a brief (from 400 to 403).

Ticket: https://trello.com/c/gCEsNuQ3/401-return-403-rather-than-400-http-code-if-supplier-not-eligible-for-brief


katstevens:
Commits now squashed!

PR #723, opened 2017-07-11
Check for empty path before calling get_key

TheDoubleK:
Recent utils version bump brought in updated s3 functions.  `get_key` used to return None for a None input, but now will raise an error instead, so the caller needs to check that the param is not None before calling.


benvand:
👍

PR #725, opened 2017-07-14
Pin requirements

allait:
requirements.txt is generated using a Python 3 virtualenv to keep the py2 backports of Python 3 stdlib packages (like `enum34` and `ipaddress`) out. This leaves those packages unpinned, but shouldn't cause any issues otherwise.

### Reorganize Makefile and requirements files

Renames requirements.txt to requirements-app.txt and requirements_for_test.txt to requirements-dev.txt.

Removes any unused dependencies or dependencies provided by other packages since those will be listed in the full requirements.txt.

Updates Makefile to add freeze-requirements and rename targets to be more consistent with other apps.

(see alphagov/digitalmarketplace-api#615 for details)


Makefile
```diff
@@ -8,53 +8,83 @@ else
...
-test_flake8: virtualenv
+.PHONY: test-flake8
+test-flake8: virtualenv
 	${VIRTUALENV_ROOT}/bin/flake8 .
 
-test_python: virtualenv
-	${VIRTUALENV_ROOT}/bin/py.test ${PYTEST_ARGS}
+.PHONY: test-python
+test-python: virtualenv
+	${VIRTUALENV_ROOT}/bin/py.test ${PYtest-ARGS}
```
Wynndow: Lower case thing.

PR #726, opened 2017-07-28
Service listing page: segregate by framework, adapt URL scheme

risicle:
Story https://trello.com/c/I0beYt9w/40-1-allow-services-page-to-filter-by-framework

In short, a supplier's service listing page has been segregated to provide listings by framework. Though for the hierarchy to continue making _any_ sense, URLs of various pages living beneath `/suppliers/services/<service_id>` had to be moved to beneath `/suppliers/frameworks/<framework_slug>/services/<service_id>`.

The supplier dashboard has had its large "View services" link moved to the per-framework links: 
20170731_servicelisting_00

The actual listing page is now at `/suppliers/frameworks/<framework_slug>/services` and has had its heading changed appropriately:
20170731_servicelisting_01

Service detail pages have been moved to `/suppliers/frameworks/<framework_slug>/services/<service_id>` and its breadcrumb pointing back to the listing page has been adjusted appropriately:
20170731_servicelisting_02

The section editing page has been moved to `/suppliers/frameworks/<framework_slug>/services/<service_id>/edit/<section_name>` and "listing" breadcrumb has been similarly updated:
20170731_servicelisting_03

The removal view has been moved too (though of course you can't see it from this screenshot):
20170731_servicelisting_04

A redirector has been added to cope with old links to services. This will universally redirect any URL beneath `/suppliers/services/<service_id>(/trailing/path)` to `/suppliers/frameworks/<framework_slug>/services/<service_id>(/trailing/path)` by looking up the services `framework_slug` (but notably only *after* checking access rights).

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -5,7 +5,12 @@ from ... import data_api_client, flask_featureflags
 from ...main import main, content_loader
 from ..helpers import login_required
 from ..helpers.services import is_service_associated_with_supplier, get_signed_document_url, count_unanswered_questions
-from ..helpers.frameworks import get_framework_and_lot, get_declaration_status, get_supplier_framework_info
+from ..helpers.frameworks import (
+    get_framework_and_lot,
+    get_declaration_status,
+    get_supplier_framework_info,
+    get_framework,
+)
 
 from dmcontent.content_loader import ContentNotFoundError
 from dmapiclient import HTTPError
@@ -13,18 +18,21 @@ from dmutils import s3
 from dmutils.documents import upload_service_documents
 
 
-@main.route('/services')
+@main.route('/services/<string:framework_slug>')
 @login_required
-def list_services():
-    suppliers_services = sorted(
-        data_api_client.find_services(supplier_id=current_user.supplier_id)["services"],
-        key=lambda service: service['frameworkSlug'],
-        reverse=True
-    )
+def list_services(framework_slug):
+    framework = get_framework(data_api_client, framework_slug, allowed_statuses=['live'])
+
+    suppliers_services = data_api_client.find_services(
+        supplier_id=current_user.supplier_id,
+        framework=framework_slug,
+    )["services"]
 
     return render_template(
         "services/list_services.html",
-        services=suppliers_services), 200
+        services=suppliers_services,
+        framework=framework,
+    ), 200
 
 
 #  #######################  EDITING LIVE SERVICES #############################
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -18,7 +18,7 @@ from dmutils import s3
 from dmutils.documents import upload_service_documents
 
 
-@main.route('/services/<string:framework_slug>')
+@main.route("/frameworks/<string:framework_slug>/services")
 @login_required
 def list_services(framework_slug):
     framework = get_framework(data_api_client, framework_slug, allowed_statuses=['live'])
@@ -38,14 +38,17 @@ def list_services(framework_slug):
 #  #######################  EDITING LIVE SERVICES #############################
 
 
-@main.route('/services/<string:service_id>', methods=['GET'])
+@main.route("/frameworks/<string:framework_slug>/services/<string:service_id>", methods=['GET'])
 @login_required
-def edit_service(service_id):
+def edit_service(framework_slug, service_id):
     service = data_api_client.get_service(service_id)
     service_unavailability_information = service.get('serviceMadeUnavailableAuditEvent')
     service = service.get('services')
 
     if not is_service_associated_with_supplier(service):
+        abort(404)
+
+    if service["frameworkSlug"] != framework_slug:
         abort(404)
 
     framework = data_api_client.get_framework(service['frameworkSlug'])['frameworks']
@@ -67,12 +70,15 @@ def edit_service(service_id):
     )
 
 
-@main.route('/services/<string:service_id>/remove', methods=['POST'])
+@main.route("/frameworks/<string:framework_slug>/services/<string:service_id>/remove", methods=['POST'])
 @login_required
-def remove_service(service_id):
+def remove_service(framework_slug, service_id):
     service = data_api_client.get_service(service_id).get('services')
 
     if not is_service_associated_with_supplier(service):
+        abort(404)
+
+    if service["frameworkSlug"] != framework_slug:
         abort(404)
 
     # dos services should not be removable
@@ -103,24 +109,31 @@ def remove_service(service_id):
             'updated_service_name': updated_service.get('serviceName')
         }, 'remove_service')
 
-        return redirect(url_for(".list_services"))
+        return redirect(url_for(".list_services", framework_slug=service["frameworkSlug"]))
 
     return redirect(url_for(
         ".edit_service",
         service_id=service_id,
+        framework_slug=service["frameworkSlug"],
         remove_requested=True))
 
 
-@main.route('/services/<string:service_id>/edit/<string:section_id>', methods=['GET'])
+@main.route(
+    "/frameworks/<string:framework_slug>/services/<string:service_id>/edit/<string:section_id>",
+    methods=['GET'],
+)
 @login_required
 @flask_featureflags.is_active_feature('EDIT_SECTIONS')
-def edit_section(service_id, section_id):
+def edit_section(framework_slug, service_id, section_id):
     service = data_api_client.get_service(service_id)
     if service is None:
         abort(404)
     service = service['services']
 
     if not is_service_associated_with_supplier(service):
+        abort(404)
+
+    if service["frameworkSlug"] != framework_slug:
         abort(404)
 
     try:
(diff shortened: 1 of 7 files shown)


tests/app/main/test_services.py
```diff
@@ -50,129 +51,289 @@ def supplier_remove_service__post_data(request):
...
+                self._mock_get_service_side_effect,
+                service_status,
+                framework_framework,
+                True,
+            )
+            res = self.client.get('/suppliers/services/31415' + suffix)
+
+            assert res.status_code == 404
+
+    @pytest.mark.parametrize("service_belongs_to_user", (False, True,))
```
idavidmcdonald: We are passing in `service_belongs_to_user` but never using it. I'm not entirely sure why we are parametrizing this case for this test but if it's intentional then will need updating to use the param

tests/app/main/test_services.py
```diff
@@ -50,129 +51,289 @@ def supplier_remove_service__post_data(request):
...
-            res = self.client.get('/suppliers/services')
+            res = self.client.get('/suppliers/frameworks/g-cloud-909/services')
             assert res.status_code == 200
+
+            document = html.fromstring(res.get_data(as_text=True))
+            assert document.xpath(
+                "//*[contains(@class, 'summary-item-no-content')][normalize-space(string())=$t]",
+                t=u"You don’t have any G-Cloud 909 services on the Digital Marketplace",
+            )
+            assert document.xpath(
```
idavidmcdonald: Nitpick but it feels more logical to assert your way down the visual page i.e starting with the heading and then getting to the table

tests/app/main/test_services.py
```diff
@@ -50,129 +51,289 @@ def supplier_remove_service__post_data(request):
...
-                supplier_id=1234)
-            assert "Service name 123" in res.get_data(as_text=True)
-            assert "Software as a Service" in res.get_data(as_text=True)
-            assert "G-Cloud 1" in res.get_data(as_text=True)
+                supplier_id=1234,
+                framework="g-cloud-909",
+            )
 
     @mock.patch('app.data_api_client')
     def test_should_not_be_able_to_see_page_if_made_inactive(self, services_data_api_client):
```
idavidmcdonald: Nitpick. If the test name said explitiely that `user` made in inactive then it's really easy to know when looking at the test name it's the user rather than a service being made inactive

tests/app/main/test_services.py
```diff
@@ -585,6 +765,32 @@ def test_remove_service(
         assert data_api_client.update_service_status.called is (expect_api_call_if_data and post_data)
         assert response.status_code == expected_status_code
 
+    @pytest.mark.parametrize("confirm", (False, True,))
+    def test_should_fail_if_incorrent_framework(
```
idavidmcdonald: Spelling mistake

tests/app/main/test_services.py
```diff
@@ -715,18 +934,33 @@ def test_update_with_under_50_words_error(self, data_api_client):
...
         data_api_client.get_service.return_value = self.empty_service
         res = self.client.post(
-            '/suppliers/services/1/edit/invalid_section'
+            '/suppliers/frameworks/g-cloud-6/services/1/edit/invalid_section'
         )
         assert res.status_code == 404
         self.assert_no_flashes()
+        assert data_api_client.update_service.called is False
+
+    def test_incorrect_framework_fails(self, data_api_client):
```
idavidmcdonald: You have a very similarly named test a few above - `test_edit_service_incorrect_framework` which both GET the same url. The test names are essentially identical but the difference is that one has a service and one doesnt (which then feels a bit weird because if it's the lack of service that's differing, why is the test name about incorrect frameworks).

I assume that you are actually testing correct things here but it might be worth thinking about putting these tests next to each other and renaming them to make it a bit clear what you mean by 'incorrect'

tests/app/main/test_suppliers.py
```diff
@@ -163,11 +177,29 @@ def test_shows_edit_buttons(self, get_current_suppliers_users, data_api_client):
...
+            document = html.fromstring(res.get_data(as_text=True))
+
+            assert document.xpath(
+                "//a[normalize-space(string())=$t][@href=$u][contains(@class, $c)]",
+                t="Edit",
+                u="/suppliers/edit",
+                c="summary-change-link",
+            )
+
+            assert document.xpath(
```
idavidmcdonald: I like how you are asserting your links

app/main/views/services.py
```diff
@@ -5,41 +5,52 @@
...
-def edit_service(service_id):
+def edit_service(framework_slug, service_id):
     service = data_api_client.get_service(service_id)
     service_unavailability_information = service.get('serviceMadeUnavailableAuditEvent')
     service = service.get('services')
 
     if not is_service_associated_with_supplier(service):
         abort(404)
 
+    if service["frameworkSlug"] != framework_slug:
```
risicle: Good... question.

app/main/views/services.py
```diff
@@ -167,12 +194,38 @@ def update_section(service_id, section_id):
...
             service_data=posted_data,
             service_id=service_id,
-            errors=errors
+            errors=errors,
         )
 
     flash({"updated_service_name": posted_data.get("serviceName") or service.get("serviceName")}, 'service_updated')
 
-    return redirect(url_for(".edit_service", service_id=service_id))
+    return redirect(url_for(".edit_service", service_id=service_id, framework_slug=service["frameworkSlug"]))
```
risicle: Probably (you can see I just appended it here)

app/main/views/services.py
```diff
@@ -167,12 +194,38 @@ def update_section(service_id, section_id):
...
+
+
+# we have to split these route definitions in two because without a fixed "/" separating the service_id and
+# trailing_path because otherwise it's not clearly defined where flask should start capturing trailing_path
```
risicle: me english fail here

tests/app/main/test_services.py
risicle: Good catch, wasn't sending it to `_mock_get_service_side_effect`

tests/app/main/test_services.py
risicle: mhm

tests/app/main/test_services.py
risicle: not my test names here, but yes i think they could be improved

tests/app/main/test_services.py
risicle: *You're* a spelling mistake, Dabid.

tests/app/main/test_services.py
risicle: One's a `GET`, the other's a `POST`, and yes, calling the fixture `empty_service` is a bit ambiguous (wasn't me) - what it actually means is a perfectly valid service that doesn't really have any content.

tests/app/main/test_suppliers.py
risicle: You flirt you.

app/main/views/services.py
risicle: I will get back to you on this one.

tests/app/main/test_services.py
idavidmcdonald: Ah sorry, you are right about one is a GET, one is a POST. Slightly more verbose test names could help again here

app/main/views/services.py
risicle: The answer is: probably yes, but we can't do it for this story and remain within time constraints.

app/main/views/services.py
```diff
@@ -59,14 +70,17 @@ def edit_service(service_id):
     )
 
 
-@main.route('/services/<string:service_id>/remove', methods=['POST'])
+@main.route("/frameworks/<string:framework_slug>/services/<string:service_id>/remove", methods=['POST'])
 @login_required
-def remove_service(service_id):
+def remove_service(framework_slug, service_id):
     service = data_api_client.get_service(service_id).get('services')
```
Wynndow: In the `edit` and `update` views below, we check that the service exists here before getting `services`.
```
service = data_api_client.get_service(service_id)
if service is None:
    abort(404)
service = service['services']
```
Is that something we should do here?

app/main/views/services.py
```diff
@@ -95,18 +109,22 @@ def remove_service(service_id):
             'updated_service_name': updated_service.get('serviceName')
         }, 'remove_service')
 
-        return redirect(url_for(".list_services"))
+        return redirect(url_for(".list_services", framework_slug=service["frameworkSlug"]))
 
     return redirect(url_for(
         ".edit_service",
         service_id=service_id,
+        framework_slug=service["frameworkSlug"],
```
Wynndow: Why not just use `framework_slug` instead of pulling it out of the service? Same with like 112.

risicle: Probably for better copypastability between views that might not have `framework_slug` as an argument, or has `framework_slug` but has redefined it to something... else.

app/main/views/services.py
risicle: So this is where half our api functions pass through a 404 and half of them swallow it and return `None`.

Anyway, this bit ain't my code and my way of properly "fixing" it would involve doing all the object fetching using decorators.

TheDoubleK:
Looks a bit odd when there's an ongoing procurement... (The link at the top, above the supplier name.)

Maybe we don't need to fix this right now, but will need to before the next procurement opens:

screen shot 2017-08-16 at 15 14 12

risicle:
Oh that *is* odd. It feels like something that should be in a banner maybe. @laurenceberry ?

PR #727, opened 2017-08-02
(beginnings of) supplier dashboard reorganization

risicle:
**This sits on top of #726, so there will be a lot of chaff in the "files changed" section until that is merged.**

Story https://trello.com/c/5jTvXMoA/42-1-implement-new-layout-for-supplier-dashboard

This rather simply switches the layout of the listing of live frameworks a supplier is on from a table to a list of headed sections containing paragraphs of links:
20170802_supplierdashboard

~~~Affects a disappointingly small number of tests but we don't really have the time to flesh them out.~~~ No sorry I had to flesh them out slightly as the first time I pushed this it was broken in a way that the tests gave me no indication of.

~~~Still awaiting a final decision on the content for the text in the case a supplier is on no frameworks.~~~ We're sticking with the text as-is.

### DIFF
--- a/app/templates/suppliers/_frameworks_live.html
+++ b/app/templates/suppliers/_frameworks_live.html
@@ -1,43 +1,26 @@
-{% import "toolkit/summary-table.html" as summary %}
-
-{{ summary.heading("Current services") }}
-{% call(framework) summary.list_table(
-  frameworks.live,
-  caption='Current services',
-  field_headings=[
-    'Label',
-    'Value',
-    ''
-  ],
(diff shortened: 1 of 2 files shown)


TheDoubleK:
I think we can close this PR as it is subsumed by #728?

risicle:
We *can*, though I _was_ going to merge them individually because people seem to like separated stories as of late. Might make the review process simpler.

risicle:
(good point, I'm bored of juggling rebased heads)

PR #728, opened 2017-08-07
Supplier details relocation & continued supplier dashboard improvements

risicle:
**Again, this branch is on top of #727 which is on top of #726 so there's loads of chaff in the "changed files"**

Story https://trello.com/c/MD0JKN5Q/41-2-create-organisation-details-page

This story is mostly about creating the separate supplier details page, but ended up becoming significant alteration of the supplier dashboard design. What alterations are part of what story has become unclear, but either way what is at the head of this branch is what @laurenceberry wants.

20170807_supplierdashboard


TheDoubleK:
Looks weird when there's an ongoing procurement (link at top of page above the supplier name).

Probably OK for now, but we'll need to fix it before we have another procurement to run.
screen shot 2017-08-16 at 15 14 12

TheDoubleK:
Also for other framework statuses...
screen shot 2017-08-16 at 15 37 21

tests/app/main/test_suppliers.py
```diff
@@ -182,24 +169,30 @@ def test_shows_edit_buttons(self, get_current_suppliers_users, data_api_client):
...
             )
+
             assert not document.xpath(
-                "//tr[./td[normalize-space(string())=$f]][.//a[normalize-space(string())=$t]]",
+                "//*[(.//h3)[1][normalize-space(string())=$f]][.//a[normalize-space(string())=$t]]",
                 f="G-Cloud 7",
                 t="View services",
             )
             assert not document.xpath("//a[@href=$u]", u="/suppliers/frameworks/g-cloud-7/services")
+            assert not document.xpath(
```
TheDoubleK: For asserting things *not* in the page I might prefer to do a much looser match (e.g. "H-Cloud 88" isn't anywhere on the page, rather than just not in this very specific element).  But we do this plenty other places and it follows what was already there, so fine.

TheDoubleK:
The messaging around frameworks with statuses other than `live` is part of the ongoing redesign I think, this all works and does what it says in the ticket so let's get it in :+1:

app/main/views/suppliers.py
```diff
@@ -77,7 +77,24 @@ def dashboard():
     ), 200
 
 
-@main.route('/edit', methods=['GET'])
```
TheDoubleK: Oh I just noticed in the Trello card it says "302 redirect existing edit page to new edit page"... do we care about doing this?  

I can't imagine anyone has the existing edit page bookmarked...

tests/app/main/test_suppliers.py
risicle: Yes indeed, this one I just lazily switched over from a positive assertion...

risicle: Although I think in this case I wouldn't want to go as far as just checking the framework name isn't in the output because I wouldn't want to rule out that there could be some line that says "Hey <framework_name> is coming" or "<framework_name> has ended" - if anyone ever added such a section they would find it hitting this assertion and almost certainly say "that doesn't apply anymore" and delete it, leaving us without this assertion.

 * [ ] But I can certainly slacken it.

app/main/views/suppliers.py
risicle:
~~~Oh forgot about that - I'll do it anyway, well caught.~~~

Actually come to think of it I think we only ever wanted to put the 302 in to make the temporary halfway situation work. And we decided to do away with the temporary situation.

tests/app/main/test_suppliers.py
risicle: Arrrgh actually this one turns out to be really hard to slacken and still have the right meaning. It looks like it's hyper-specific, but it's actually just saying "there isn't an element with an `h3` title of 'H-Cloud 88' which also contains a 'View services' link".

TheDoubleK: Yep don't fret it, I think it's fine as is.

PR #731, opened 2017-08-18
Remove brief response flow

Wynndow:
THIS PR MUST BE MERGED FIRST [https://github.com/alphagov/digitalmarketplace-brief-responses-frontend/pull/4](https://github.com/alphagov/digitalmarketplace-brief-responses-frontend/pull/4)

Now that we have a brief response frontend app that handles the response flow, we can remove the duplicated code and functionality from the supplier app.

The views that were removed are everything behind /suppliers/opportunities.
The view for the supplier opportunities dashboard has also been moved, and the URL changed. A 301 redirect has been put in place to redirect to its new location in the brief responses frontend app.

The new supplier flow feature flag has also been dropped as it's not needed in the supplier app anymore.


app/main/views/frameworks.py
```diff
@@ -1193,22 +1193,4 @@ def view_contract_variation(framework_slug, variation_slug):
 @main.route('/frameworks/<framework_slug>/opportunities', methods=['GET'])
 @login_required
 def opportunities_dashboard(framework_slug):
```
allait: I wonder if it's worth renaming this view (with `_redirect` or `_deprecated`) to make sure it's not used to generate links within the apps anymore. All`url_for` calls should be using `external.opportunities_dashboard`

app/main/views/frameworks.py
```diff
@@ -1193,22 +1193,4 @@ def view_contract_variation(framework_slug, variation_slug):
...
-    opportunities = data_api_client.find_brief_responses(
-        supplier_id=current_user.supplier_id, framework=framework_slug
-    )['briefResponses']
-
-    return render_template(
-        "suppliers/opportunities_dashboard.html",
-        framework=framework,
-        completed=opportunities
-    ), 200
+    return redirect("/suppliers/opportunities/frameworks/{}".format(framework_slug), code=301)
```
Wynndow: Yep, makes sense. Have updated.

PR #735, opened 2017-08-29
Display new supplier fields

TheDoubleK:
For this story in Trello: https://trello.com/c/GFp7ncpN/62-update-the-supplier-page-to-expose-the-migrated-declaration-information-2

**THERE IS AN ACCOMPANYING LEGACY FUNCTIONAL TESTS PR TO MERGE AT THE SAME TIME**: https://github.com/alphagov/digitalmarketplace-functional-tests/pull/336

This adds the new "Registration information" section to the supplier details page, and also updates the edit supplier information form to only show the fields that are now displayed.  Wording is updated to match the prototype, and the fields on the edit page match the names of the fields on the display page (which previously they didn't 😞 ).

Updated page looks like this:
screen shot 2017-08-29 at 18 23 32

And like this for a supplier with more registration info:
screen shot 2017-08-29 at 18 25 47

### DIFF
--- a/app/main/forms/suppliers.py
+++ b/app/main/forms/suppliers.py
@@ -1,5 +1,5 @@
 from flask.ext.wtf import Form
-from wtforms import IntegerField, FieldList
+from wtforms import IntegerField
 from wtforms.validators import DataRequired, ValidationError, Length, Optional, Regexp
 from dmutils.forms import StripWhitespaceStringField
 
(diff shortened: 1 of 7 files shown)


app/main/forms/suppliers.py
```diff
@@ -22,28 +22,22 @@ class EditSupplierForm(Form):
...
-    postcode = StripWhitespaceStringField()
-    website = StripWhitespaceStringField()
-    phoneNumber = StripWhitespaceStringField('Phone number')
-    email = StripWhitespaceStringField('Email address', validators=[
+    contactName = StripWhitespaceStringField('Contact name', validators=[
+        DataRequired(message="You must provide a contact name"),
+    ])
+    email = StripWhitespaceStringField('Contact email', validators=[
         DataRequired(message="You must provide an email address"),
         Regexp("^[^@^\s]+@[^@^\.^\s]+(\.[^@^\.^\s]+)+$",
```
risicle: This is the regex that turns out to be wrong but I'm fixing it in pulling out a specific `EmailField`

TheDoubleK: The only thing changed in this PR is the order of fields and the words ("Email address" -> "Contact email"), so this could go in without the updated regex and functionality wouldn't change from what we have at the moment.

risicle: Absolutely yes just pointing it out

tests/app/main/test_suppliers.py
```diff
@@ -712,22 +723,54 @@ def test_shows_supplier_info(self, data_api_client):
...
+        data_api_client.get_supplier.return_value = get_supplier(companiesHouseNumber=None)
+        with self.app.test_client():
+            self.login()
+
+            res = self.client.get("/suppliers/details")
+            assert res.status_code == 200
+            page_html = res.get_data(as_text=True)
+            document = html.fromstring(page_html)
+            for property_str in ("BZ", "BEL153",):
+                assert document.xpath("//*[normalize-space(string())=$t]", t=property_str), property_str
```
risicle: Ah clever remembering to add `property_str` as the message - I should do this more.

app/templates/suppliers/details.html
```diff
@@ -74,10 +70,74 @@
...
+      {{ summary.text(supplier.companiesHouseNumber) }}
+    {% endcall %}
+  {% endif %}
+
+  {% if supplier.registrationCountry and not supplier.companiesHouseNumber %}
+    {% call summary.row() %}
+      {{ summary.field_name('Where your business was established') }}
+      {{ summary.text(supplier.registrationCountry|upper) }}
+    {% endcall %}
+  {% endif %}
```
risicle: Are we *really* sure we want to show this when we have such incomplete data? (and when we do have data it will presumably come through as its ISO3166 code, leaving some possibly confused users)

TheDoubleK: I don't know... it will only exist for UK-based suppliers who have filled out a recent declaration.  Seeing "GB" shouldn't be that confusing to them?

TheDoubleK: Everyone else will see nothing and be none the wiser.

risicle: Until someone goes and writes a script to populate it, not realizing that it will just show the ISO code raw. Whatever, it doesn't really matter - most people recognize their country's ISO3166-1a2 code anyway

PR #737, opened 2017-09-01
Update content in sign up flow

TheDoubleK:
For this ticket on Trello: https://trello.com/c/wHRA6ScY/74-update-the-sign-up-journey-content-development-1

The working document has got pretty messy, but final changes here are cleared with Constance and I've shown the updated flow to @laurenceberry and he's happy with how it all looks and reads. I've attached screenshots of the full flow below - not all of the pages have been changed, but you can see how it all hangs together better with it all there I think.

There is a corresponding legacy functional tests update here, which should be merged at the same time as this: 

 - [ ] https://github.com/alphagov/digitalmarketplace-functional-tests/pull/339

## Page 1

screen shot 2017-09-04 at 10 48 44

## Page 2

screen shot 2017-09-01 at 15 17 04

## Page 3

screen shot 2017-09-01 at 15 17 12

## Page 4

screen shot 2017-09-01 at 15 17 22

## Page 5

screen shot 2017-09-01 at 15 17 36

## Page 6

screen shot 2017-09-01 at 15 17 51

## Page 7

screen shot 2017-09-01 at 15 18 05

## Page 8

screen shot 2017-09-01 at 15 18 17

### DIFF
--- a/app/templates/suppliers/companies_house_number.html
+++ b/app/templates/suppliers/companies_house_number.html
@@ -1,6 +1,6 @@
 {% extends "_base_page.html" %}
 
-{% block page_title %}Companies House – Create new supplier – Digital Marketplace{% endblock %}
+{% block page_title %}Companies House number – Create a supplier account – Digital Marketplace{% endblock %}
 
 {% block breadcrumb %}
 {%
@@ -36,9 +36,7 @@
             value = form.companies_house_number.data,
             error = form.companies_house_number.errors[0],
             question_advice =
-            'This is a unique 8-character number given to your company by Companies House.
-            Find your <a href="https://beta.companieshouse.gov.uk/">Companies House number</a>.
-            ' | safe
+            '<a href="https://beta.companieshouse.gov.uk/">Find your Companies House number</a>' | safe
         %}
         {% include "toolkit/forms/textbox.html" %}
         {% endwith %}
--- a/app/templates/suppliers/companies_house_number.html
+++ b/app/templates/suppliers/companies_house_number.html
@@ -30,14 +30,16 @@
     <div class="column-two-thirds">
       <form method="POST" action="{{ url_for('.submit_companies_house_number') }}">
         <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
+        {% set question_advice %}
+        <a href="https://beta.companieshouse.gov.uk/">Find your Companies House number</a>
+        {% endset %}
         {%
           with
             question = "Companies House number",
             name = "companies_house_number",
             value = form.companies_house_number.data,
             error = form.companies_house_number.errors[0],
-            question_advice =
-            '<a href="https://beta.companieshouse.gov.uk/">Find your Companies House number</a>' | safe
+            question_advice = question_advice
         %}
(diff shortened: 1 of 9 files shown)


TheDoubleK:
Comment from Constance: 

"Says 'to' twice on page 1
and looking at it with a fresh pair of eyes I'm not sure about: The bit about clicking on the link, distracts from the start button and isn't relevant at that time."

risicle:
(I perfer the term "follow the link" but I suspect that's not what she's getting at)

app/templates/suppliers/companies_house_number.html
```diff
@@ -36,9 +37,7 @@
             value = form.companies_house_number.data,
             error = form.companies_house_number.errors[0],
             question_advice =
-            'This is a unique 8-character number given to your company by Companies House.
-            Find your <a href="https://beta.companieshouse.gov.uk/">Companies House number</a>.
-            ' | safe
+            '<a href="https://beta.companieshouse.gov.uk/">Find your Companies House number</a>' | safe
```
risicle: Side note: this is what I see as one of the significant flaws of the frontend toolkit - I tend to use the `{% set ... %}` tag to set values like this.

app/templates/suppliers/create_new_supplier.html
```diff
@@ -19,26 +19,27 @@
...
+        </p>
+
+        <p>
+          This will take 3 to 5 days.
+        </p>
+        <br />
+        <p>When you have set up your account we will send you a confirmation email. You must click on the link in the
+           email to activate your supplier account.</p>
       </div>
+      <br />
```
risicle: A `<br/>` after a block element? Would be surprised if this had any effect.

risicle: Oh it does. Always thought superfluous `<br/>`s got squashed out.

TheDoubleK: With the `<br />`:
screen shot 2017-09-04 at 10 50 16

Without the `<br />`:
screen shot 2017-09-04 at 10 50 04

TheDoubleK: Whether this is the "right" way to do this is another matter completely :)

TheDoubleK:
Updated content and screenshot for Page 1 ("Create a supplier account") based on comments from Constance.

app/templates/suppliers/companies_house_number.html
TheDoubleK: Auto escaping?  How's that?  I'm gonna try it out now.

risicle:
I assume you've found the non-legacy FTs to be unaffected?

app/templates/suppliers/companies_house_number.html
risicle: The same auto escaping all other regular text in jinja gets, no `|safe` required.

risicle: or maybe a better way to put it... *without* escaping.

risicle: Blimey wasn't even expecting it for that much.

PR #738, opened 2017-09-04
Bugfix: not validating Companies House strictly enough on frontend

TheDoubleK:
We recently (in this commit: https://github.com/alphagov/digitalmarketplace-api/pull/626/commits/fd76ee0bb6f0edacf29ed239d78424650b05e00e) introduced a validator on the API to restrict Companies House numbers to valid formats, but didn't update validation in the frontend flow here at the same time.

Because nothing gets posted to the API until the final step of this flow it is currently possible to submit an invalid CH number here and then have the API post request fail when submitting the final step.

It's not ideal to have the same regex in two places, but this is a current production bug and this will fix it for now.

### DIFF
--- a/app/main/forms/suppliers.py
+++ b/app/main/forms/suppliers.py
@@ -50,7 +50,9 @@ class DunsNumberForm(Form):
 class CompaniesHouseNumberForm(Form):
     companies_house_number = StripWhitespaceStringField('Companies house number', validators=[
         Optional(),
-        Length(min=8, max=8, message="Companies House numbers must have 8 characters.")
+        Regexp(r'^([0-9]{2}|[A-Za-z]{2})[0-9]{6}$',
+               message="Companies House numbers must have either 8 digits or 2 letters followed by 6 digits."
+               )
     ])
 
 
--- a/tests/app/main/test_suppliers.py
+++ b/tests/app/main/test_suppliers.py
@@ -1158,21 +1158,40 @@ class TestCreateSupplier(BaseApplicationTest):
         res = self.client.post(
             "/suppliers/companies-house-number",
             data={
-                'companies_house_number': "short"
+                'companies_house_number': "1234567"
             }
         )
         assert res.status_code == 400
-        assert "Companies House numbers must have 8 characters." in res.get_data(as_text=True)
+        assert ("Companies House numbers must have either 8 digits or 2 letters followed by 6 digits."
+                in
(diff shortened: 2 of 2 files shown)


app/main/forms/suppliers.py
```diff
@@ -50,7 +50,9 @@ class DunsNumberForm(Form):
 class CompaniesHouseNumberForm(Form):
     companies_house_number = StripWhitespaceStringField('Companies house number', validators=[
         Optional(),
-        Length(min=8, max=8, message="Companies House numbers must have 8 characters.")
+        Regexp(r'^([0-9]{2}|[A-Za-z]{2})[0-9]{6}$',
+               message="Companies House numbers must have 8 digits or 2 letters followed by 6 digits."
```
allait: Could we encode the rule as a regexp directly: `[0-9]{8} | [A-Za-z]{2}[0-9]{6}`. It's a bit longer, but it does exactly what the message says, which I think makes it easier to parse.

The rest looks good!

Wynndow: Could this message be confusing? `12345678123456` and `AB123456` both fit the description.

TheDoubleK: Hmm, perhaps. Words are hard.

TheDoubleK: I can see your point but this exactly matches the regex already used in the API, so might be nice to keep them the same? What do you reckon?  Could change in both places, but... pull requests.

Wynndow: "Companies House numbers must start with either 8 digits or 2 letters, followed by 6 digits."

Or something.

TheDoubleK: I did have an "either" in there before but deleted it thinking it was superfluous.  What do you reckon to "Companies House numbers must have *either* 8 digits or 2 letters followed by 6 digits." - better?

Wynndow: Again, could be `12345678` or `AB123456`. I think the comma separates the two parts out.

To be honest, it's probably fine though. I'm just nit picking. Whatever you think is best.

Wynndow: Wait, ignore me. I'm very tired. Yours is good.

allait: Oh, yeah, then keep it as is. I guess the message might change as well and it doesn't describe the format that much better either.

PR #740, opened 2017-09-08
Implement an email sign up capability for suppliers

risicle:
Story https://trello.com/c/V8bUSjq8/77-implement-an-email-sign-up-capability-for-suppliers-2

This story adds a `/supplier/mailing-list` view which will allow a (non-)user to sign up to a mailchimp mailing list:

20170912_mailing_list_before

Taking its its mailchimp configuration from the config variables `DM_MAILCHIMP_USERNAME`, `DM_MAILCHIMP_API_KEY` and `DM_MAILCHIMP_OPEN_FRAMEWORK_NOTIFICATION_MAILING_LIST_ID`. If this fails for some reason the user is brought back to the same page with an error flash message:

20170912_mailing_list_failure

On success, they are redirected to the DM root page complete with success flash message using georgian flashing:

20170912_mailing_list_success

As part of this I consolidated the app's email form field validation:

20170912_mailing_list_invalid
20170912_mailing_list_blank

### DIFF
--- a/app/main/forms/suppliers.py
+++ b/app/main/forms/suppliers.py
@@ -1,7 +1,7 @@
 from flask.ext.wtf import Form
 from wtforms import IntegerField
 from wtforms.validators import DataRequired, ValidationError, Length, Optional, Regexp
-from dmutils.forms import StripWhitespaceStringField
+from dmutils.forms import StripWhitespaceStringField, EmailField, EmailValidator
 
 
 def word_length(limit=None, message=None):
@@ -29,10 +29,8 @@ class EditContactInformationForm(Form):
     contactName = StripWhitespaceStringField('Contact name', validators=[
         DataRequired(message="You must provide a contact name"),
     ])
-    email = StripWhitespaceStringField('Contact email', validators=[
+    email = EmailField('Contact email', validators=[
         DataRequired(message="You must provide an email address"),
-        Regexp("^[^@^\s]+@[^@^\.^\s]+(\.[^@^\.^\s]+)+$",
-               message="Please enter a valid email address")
     ])
     phoneNumber = StripWhitespaceStringField('Contact phone number')
     address1 = StripWhitespaceStringField('Building and street')
@@ -70,8 +68,7 @@ class CompanyContactDetailsForm(Form):
     ])
     email_address = StripWhitespaceStringField('Contact email address', validators=[
         DataRequired(message="You must provide an email address."),
-        Regexp("^[^@^\s]+@[^@^\.^\s]+(\.[^@^\.^\s]+)+$",
-               message="You must provide a valid email address.")
+        EmailValidator(message="You must provide a valid email address."),
     ])
     phone_number = StripWhitespaceStringField('Contact phone number', validators=[
         DataRequired(message="You must provide a phone number."),
@@ -82,6 +79,5 @@ class CompanyContactDetailsForm(Form):
 class EmailAddressForm(Form):
     email_address = StripWhitespaceStringField('Email address', validators=[
         DataRequired(message="You must provide an email address."),
-        Regexp("^[^@^\s]+@[^@^\.^\s]+(\.[^@^\.^\s]+)+$",
-               message="You must provide a valid email address.")
+        EmailValidator(message="You must provide a valid email address."),
     ])
--- a/app/main/forms/suppliers.py
+++ b/app/main/forms/suppliers.py
@@ -29,7 +29,7 @@ class EditContactInformationForm(Form):
     contactName = StripWhitespaceStringField('Contact name', validators=[
         DataRequired(message="You must provide a contact name"),
     ])
-    email = EmailField('Contact email', validators=[
+    email = EmailField('Contact email address', validators=[
         DataRequired(message="You must provide an email address"),
     ])
     phoneNumber = StripWhitespaceStringField('Contact phone number')
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -2,7 +2,7 @@
 
 from itertools import chain
 
-from flask import render_template, request, redirect, url_for, abort, session, Markup
+from flask import render_template, request, redirect, url_for, abort, session, Markup, flash
 from flask_login import current_user, current_app
 import six
 
@@ -10,6 +10,7 @@ from dmapiclient import APIError
 from dmapiclient.audit import AuditTypes
 from dmutils.email import send_email, generate_token
 from dmutils.email.exceptions import EmailError
+from dmutils.email.dm_mailchimp import DMMailChimpClient
 from dmcontent.content_loader import ContentNotFoundError
 
 from ...main import main, content_loader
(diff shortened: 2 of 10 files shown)


app/main/views/suppliers.py
```diff
@@ -467,3 +468,48 @@ def create_your_account_complete():
...
+                # success
+
+                # this message will be consumed by the buyer app which has been converted to display raw "literal"
+                # content from the session
+                flash(Markup(render_template(
+                    "flashmessages/join_open_framework_notification_mailing_list_success.html",
+                    email_address=form.data["email_address"],
+                )), "success")
+
+                return redirect("/")
```
TheDoubleK: I'd probably have said redirect to the "Become a supplier" page (which doesn't exist quite yet but is where this form will be linked from).  And then it would all have been within-app for flash messages.  But I missed that conversation and it's not worth unpicking this now.

app/main/views/suppliers.py
```diff
@@ -467,3 +468,48 @@ def create_your_account_complete():
...
+                # fall through to re-display form with error
+        else:
+            status = 400
+            # fall through to re-display form with errors
+    else:
+        form = EmailAddressForm()
+
+    return render_template(
+        "suppliers/join_open_framework_notification_mailing_list.html",
+        form=form,
```
TheDoubleK: Just double checking... any validation errors are in the form object here and don't need to be passed through to the template separately? (Sorry, can't remember exactly how these work)

requirements-app.txt
```diff
@@ -5,6 +5,6 @@ Flask==0.10.1
 Flask-Login==0.2.11
 Flask-WTF==0.12
 
-git+https://github.com/alphagov/digitalmarketplace-utils.git@27.1.2#egg=digitalmarketplace-utils==27.1.2
+git+https://github.com/alphagov/digitalmarketplace-utils.git@ris-email-form-field#egg=digitalmarketplace-utils==123.123.127
```
TheDoubleK: This will need to change once the utils dependency is in.

tests/app/main/test_suppliers.py
```diff
@@ -1665,3 +1666,192 @@ def test_should_show_email_address_even_when_refreshed(self):
...
+        mailchimp_client_instance.subscribe_new_email_to_list.side_effect = AssertionError("This should not be called")
+
+        mailchimp_client_class.side_effect = assert_args_and_return(
+            mailchimp_client_instance,
+            "not_a_real_username",
+            "not_a_real_key",
+            mock.ANY,
+        )
+
+        with self.client as c:
```
TheDoubleK: What benefit does this `with` give us here?  `c` is only used once in the following block, is there a problem with just doing `response = self.client.get("/suppliers/mailing-list")` as we have in most other tests?

tests/app/main/test_suppliers.py
```diff
@@ -1665,3 +1666,192 @@ def test_should_show_email_address_even_when_refreshed(self):
...
```
TheDoubleK: Same here, I'm not sure what we gain from the `with`?

tests/app/main/test_suppliers.py
```diff
@@ -1665,3 +1666,192 @@ def test_should_show_email_address_even_when_refreshed(self):
...
+                "//*[contains(@class, 'banner-destructive-without-action')][contains(normalize-space(string()), $t)]//"
+                "a[@href=$m][normalize-space(string())=$e]",
+                t="The service is unavailable at the moment",
+                m="mailto:enquiries@digitalmarketplace.service.gov.uk",
+                e="enquiries@digitalmarketplace.service.gov.uk",
+            )
+
+            with self.client.session_transaction() as session:
+                # flash message should have been consumed by view's own page rendering
+                assert not session.get('_flashes')
```
TheDoubleK: We have an `assert_flashes` test helper that might be useful for checking the correct flash message is set before it gets consumed?  Although the presence of the error banner on the page is probably enough to convince us things are working properly.

tests/app/main/test_suppliers.py
```diff
@@ -1665,3 +1666,192 @@ def test_should_show_email_address_even_when_refreshed(self):
...
+                },
+            )
+            assert response.status_code == 503
+            doc = html.fromstring(response.get_data(as_text=True), base_url="/suppliers/mailing-list")
+
+            assert mailchimp_client_instance.subscribe_new_email_to_list.called is True
+
+            form = self._common_page_asserts_and_get_form(doc)
+            assert tuple(inp.value for inp in form.xpath(".//input[@name='email_address']")) == (
+                "squinting@ger.ty",
```
TheDoubleK: I know `.ty` isn't a valid top level domain, but we generally try to use `@example.com` for fake emails in tests, as in "in case"

app/templates/flashmessages/join_open_framework_notification_mailing_list_success.html
```diff
@@ -0,0 +1 @@
+You will receive email notifications to {{ email_address }} when applications are opening.
```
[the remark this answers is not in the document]
risicle: You sure we want a line terminator? It's going to be rendered to a flash message.

app/main/forms/suppliers.py
```diff
@@ -29,10 +29,8 @@ class EditContactInformationForm(Form):
     contactName = StripWhitespaceStringField('Contact name', validators=[
         DataRequired(message="You must provide a contact name"),
     ])
-    email = StripWhitespaceStringField('Contact email', validators=[
+    email = EmailField('Contact email', validators=[
```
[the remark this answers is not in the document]
risicle: > why do we make this an `EmailField` but add an `EmailValidator` to the other `StripWhitespaceStringFields`

Subtly different validation messages: see comments in https://github.com/alphagov/digitalmarketplace-utils/pull/325

app/main/views/suppliers.py
```diff
@@ -467,3 +468,48 @@ def create_your_account_complete():
...
+                current_app.config["DM_MAILCHIMP_API_KEY"],
+                current_app.logger,
+            )
+
+            mc_response = dmmc_client.subscribe_new_email_to_list(
+                current_app.config["DM_MAILCHIMP_OPEN_FRAMEWORK_NOTIFICATION_MAILING_LIST_ID"],
+                form.data["email_address"],
+            )
+            # note we're signalling our flash messages in two separate ways here
+            if mc_response not in (True, False,):
```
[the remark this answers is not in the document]
risicle: It's a `requests` response object.

app/main/views/suppliers.py
```diff
@@ -467,3 +468,48 @@ def create_your_account_complete():
...
+                # success
+
```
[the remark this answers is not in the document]
risicle: Yeah might be an idea.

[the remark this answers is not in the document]
TheDoubleK: Double-checked with briefs team: yes they can re-create their list from data in the database so they didn't feel audit was necessary for them.  But in our case I think it'd be worth creating a new audit event type (in APIClient, confusingly) `email_subscription` or some such, and set it with `data` including the list ID and email address (and maybe also the fact it's Mailchimp service).

app/main/views/suppliers.py
risicle: Yup

tests/app/main/test_suppliers.py
risicle: Oh hell I think I'm supposed to be using `c.session_transaction()`, and the `c` is supposed to be the context that binds the http session together.

tests/app/main/test_suppliers.py
risicle: That does kinda limit you when you're trying to test an email address validator.

app/main/views/suppliers.py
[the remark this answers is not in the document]
risicle: `mailing_list_subscription`

app/main/views/suppliers.py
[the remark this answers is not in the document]
risicle: No hold on I think it's a json dict - can't remember.

tests/app/main/test_suppliers.py
TheDoubleK: Ha!  True.

app/templates/flashmessages/join_open_framework_notification_mailing_list_success.html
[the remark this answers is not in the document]
TheDoubleK: HTML ignores whitespace so should be fine in a flash message?  It just makes me twitch a bit when there isn't one on a file.  I can live with it if you reckon it's better without.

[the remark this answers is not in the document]
risicle: Cool fixed.

config.py
```diff
@@ -33,6 +33,10 @@ class Config(object):
     DM_SUBMISSIONS_BUCKET = None
     DM_ASSETS_URL = None
 
+    DM_MAILCHIMP_USERNAME = None
```
TheDoubleK: We'll need to get these three new environment variables set for Preview/Staging/Production in credentials/PaaS setup before we merge/release this.

TheDoubleK: I don't 100% know how to do that... do you want to figure it out or shall I?

risicle:
Oh, this also depends on a version of the API that's making its way out at the moment anyway without my help.

tests/app/main/test_suppliers.py
```diff
@@ -1665,3 +1667,223 @@ def test_should_show_email_address_even_when_refreshed(self):
...
+                    # by our regex and we want to be able to check their escaping in the flash message
+                    "email_address": "qu&rt@four.pence",
+                },
+            )
+
+            assert response.status_code == 302
+            assert response.location == "http://localhost/"
+            assert mailchimp_client_instance.subscribe_new_email_to_list.called is True
+
+            with c.session_transaction() as session:
```
TheDoubleK: I think this is the same thing that the helper `self.assert_flashes` does.  Might be nice to stick with using that.

risicle: Done. Interesting that `assert_flashes` doesn't seem to give a crap about binding to the same client context, so that casts questions on the use of any of the `with self.client as c:`, so I've removed that too.

app/main/forms/suppliers.py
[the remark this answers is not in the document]
risicle: Done, but I bet this is going to be the straw that broke the FT's back.

PR #741, opened 2017-09-14
Become a supplier page

TheDoubleK:
This depends on this frameworks PR being merged before tests will pass: 

 - [x] https://github.com/alphagov/digitalmarketplace-frameworks/pull/469

For this Trello ticket: https://trello.com/c/kMYuzoge/75-create-the-become-a-supplier-page-2

## One open, one opening
screen shot 2017-09-18 at 09 20 07

## Both closed
screen shot 2017-09-18 at 09 21 16

## One coming one closed
screen shot 2017-09-18 at 09 21 49

### DIFF
--- a/app/main/helpers/frameworks.py
+++ b/app/main/helpers/frameworks.py
@@ -311,3 +311,26 @@ def check_agreement_is_related_to_supplier_framework_or_abort(agreement, supplie
         abort(404)
     if not agreement.get('frameworkSlug') or agreement.get('frameworkSlug') != supplier_framework.get('frameworkSlug'):
         abort(404)
+
+
+def get_frameworks_closed_and_open_for_applications(frameworks):
+
+    all_open_fwks = []
+    all_opening_fwks = []
+    all_closed_fwks = []
+    framework_frameworks = set([framework['framework'] for framework in frameworks])
+
+    # We need to find one framework iteration per framework-framework, open > coming > closed
+    for f_f in framework_frameworks:
+        framework_iterations = [framework for framework in frameworks if framework['framework'] == f_f]
+        open_fwks = [fwk for fwk in framework_iterations if fwk['status'] == 'open']
+        opening_fwks = [fwk for fwk in framework_iterations if fwk['status'] == 'coming']
+        closed_fwks = [fwk for fwk in framework_iterations if fwk['status'] not in ('open', 'coming')]
+        if open_fwks:
+            all_open_fwks.append(open_fwks[0])
+        elif opening_fwks:
+            all_opening_fwks.append(opening_fwks[0])
+        else:
+            all_closed_fwks.append(closed_fwks[0])
+
+    return all_open_fwks, all_opening_fwks, all_closed_fwks
--- a/app/main/helpers/frameworks.py
+++ b/app/main/helpers/frameworks.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 import re
 from datetime import datetime
+from itertools import chain, islice, groupby
 
 from dmutils.formats import DATETIME_FORMAT
 from flask import abort
@@ -314,23 +315,22 @@ def check_agreement_is_related_to_supplier_framework_or_abort(agreement, supplie
 
 
 def get_frameworks_closed_and_open_for_applications(frameworks):
-
-    all_open_fwks = []
-    all_opening_fwks = []
-    all_closed_fwks = []
-    framework_frameworks = set([framework['framework'] for framework in frameworks])
-
-    # We need to find one framework iteration per framework-framework, open > coming > closed
-    for f_f in framework_frameworks:
-        framework_iterations = [framework for framework in frameworks if framework['framework'] == f_f]
-        open_fwks = [fwk for fwk in framework_iterations if fwk['status'] == 'open']
-        opening_fwks = [fwk for fwk in framework_iterations if fwk['status'] == 'coming']
-        closed_fwks = [fwk for fwk in framework_iterations if fwk['status'] not in ('open', 'coming')]
-        if open_fwks:
-            all_open_fwks.append(open_fwks[0])
-        elif opening_fwks:
-            all_opening_fwks.append(opening_fwks[0])
+    # This will find one framework iteration per framework-framework, open > coming > closed
+    def status_priority(status):
+        if status == "open":
+            return 0
+        elif status == "coming":
+            return 1
         else:
-            all_closed_fwks.append(closed_fwks[0])
+            return 2
 
-    return all_open_fwks, all_opening_fwks, all_closed_fwks
+    return tuple(chain.from_iterable(
+        islice(grp, 1)          # take the first framework
+        for _, grp in groupby(  # from each framework_framework
+            sorted(             # listed in priority order
+                (fw for fw in frameworks),
+                key=lambda fw_sort: (fw_sort["framework"], status_priority(fw_sort["status"])),
+            ),
+            key=lambda fw_groupby: fw_groupby["framework"],
+        )
+    ))
--- a/app/templates/suppliers/become_a_supplier.html
+++ b/app/templates/suppliers/become_a_supplier.html
@@ -1,0 +1,91 @@
+{% extends "_base_page.html" %}
+
+{% block page_title %}Become a supplier – Digital Marketplace{% endblock %}
+
+{% block breadcrumb %}
+  {%
+    with items = [
+      {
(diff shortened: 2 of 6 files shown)


app/main/helpers/frameworks.py
```diff
@@ -311,3 +311,26 @@ def check_agreement_is_related_to_supplier_framework_or_abort(agreement, supplie
...
+
+def get_frameworks_closed_and_open_for_applications(frameworks):
+
+    all_open_fwks = []
+    all_opening_fwks = []
+    all_closed_fwks = []
+    framework_frameworks = set([framework['framework'] for framework in frameworks])
+
+    # We need to find one framework iteration per framework-framework, open > coming > closed
+    for f_f in framework_frameworks:
```
risicle: `f_f`? `o_o`.

app/main/helpers/frameworks.py
```diff
@@ -311,3 +311,26 @@ def check_agreement_is_related_to_supplier_framework_or_abort(agreement, supplie
...
+        opening_fwks = [fwk for fwk in framework_iterations if fwk['status'] == 'coming']
+        closed_fwks = [fwk for fwk in framework_iterations if fwk['status'] not in ('open', 'coming')]
+        if open_fwks:
+            all_open_fwks.append(open_fwks[0])
+        elif opening_fwks:
+            all_opening_fwks.append(opening_fwks[0])
+        else:
+            all_closed_fwks.append(closed_fwks[0])
+
+    return all_open_fwks, all_opening_fwks, all_closed_fwks
```
risicle: It should probably be commented that the output of this implicitly depends on the ordering of `frameworks`.

risicle: Clearer *to me* would be e.g.

```
all_open_fwks = tuple(chain.from_iterable(
    islice(grp, 1)  # only want the first of each metaframework
    for _, grp in groupby(
        sorted(
            (fw for fw in frameworks if fw["status"] in ("open",)),
            key=lambda fw: fw["framework"],
        ),
        key=lambda fw: fw["framework"],
    )
))
```

the `("open",)` could then be replaced by function argument and this could generate all three sequences.

risicle: (or maybe a lambda to be able to cope with the `not in`)

app/templates/suppliers/become_a_supplier.html
```diff
@@ -0,0 +1,91 @@
...
+
+      {% if open_fwks or opening_fwks %}
+        <div class="dmspeak">
+          <h2 class="heading-xmedium padding-bottom-small">Services you can apply to sell</h2>
+
+        {% for framework in open_fwks + opening_fwks %}
+
+          {{ content_loader.get_message(framework.slug, 'become-a-supplier', framework.status) }}
+          {% if not loop.last %}
+          <p>&nbsp;</p>
```
risicle: ?

tests/app/main/test_suppliers.py
```diff
@@ -1877,3 +1877,144 @@ def test_post_valid_email_success(self, mailchimp_client_class, data_api_client)
...
+            # Check the right framework content is there
+            assert u'‘Digital Outcomes and Specialists’ is closed for applications.' in data
+            assert u'‘G-Cloud’ is closed for applications.' in data
+
+            # Check the right calls to action are there
+            assert 'Create a supplier account' not in data
+            assert 'Get notifications when applications are opening' in data
+
+    def test_one_open_one_closed_framework(self, data_api_client):
+        data_api_client.find_frameworks.return_value = {
```
risicle: Just to be anal, I'd assert the (lack of) args submitted to `find_frameworks` too.

app/templates/suppliers/become_a_supplier.html
TheDoubleK: It looks shit without it...

risicle: Shouldn't we have a "double-spaced paras" style then?

TheDoubleK: Probably...

app/main/helpers/frameworks.py
TheDoubleK: I find this much more confusing, but would be good to factor out to a function, I'll see what I can do.

TheDoubleK: Actually I that because we only one one framework per metaframework across all three lists this needs to be an atomic operation that generates all three lists in parallel.
If we have a function like this for each status (or set of statuses) then we then additionally need a whole bunch of extra logic to whittle it down (if there's a g-cloud framework in list one then remove all g-clouds from lists 2 and 3, if there's a dos framework in list 2 then remove any dos frameworks from list 3, etc.)

risicle: Well, it's plugging together a few tools to get a result - you can tell approximately what it's trying to do by its use of `groupby` and `sorted`, where the alternative is just "well, it throws around some lists and stuff, picks some values out and then something comes out the end". To me it's like the difference between reading assembly and reading a high-level expression.

risicle: ohhhhright.

risicle: still, easy to misread this style of code.

TheDoubleK: I'm trying to figure out a neat way to do this still... don't hold your breath.

app/templates/suppliers/become_a_supplier.html
TheDoubleK: Better...?
The gap using this style is slightly smaller but looks fine to me:
screen shot 2017-09-15 at 16 59 58

app/main/helpers/frameworks.py
risicle: Well, it's quite a simple change to my method:

```
def status_priority(status):
    if status == "open":
        return 0
    elif status not in ...:
        return 1
    else:
        return 2

displayed_fwks = tuple(chain.from_iterable(
    islice(grp, 1)  # only want the first of each metaframework
    for _, grp in groupby(
        sorted(
            (fw for fw in frameworks if fw["status"] in ("open",)),
            key=lambda fw: (fw["framework"], status_priority(fw["status"])),
        ),
        key=lambda fw: fw["framework"],
    )
))
```

Then you don't even have to `chain` the results back together in the view. Because that's what this function is doing - it's a prioritization, a sort, with some rules over uniqueness which are easily doable using `groupby`

app/templates/suppliers/become_a_supplier.html
risicle: I guess so... feels like something lacking in our stylesheets if we can't get a bulleted list displaying in a way we think "looks good" without wrapping things in extra elements...

app/main/helpers/frameworks.py
TheDoubleK: But if you have a single list of "displayed_frameworks" the logic in the view needs to be completely different, and deal with six possible framework states.

risicle: Oh, they're displayed in yet another different order, aren't they. Well, that's just another sorting, and it's quite easy to filter out just the statuses you want.

Anyway, this feels like a discussion for mon.

app/templates/suppliers/become_a_supplier.html
TheDoubleK: The tricky thing is that the bulleted list comes out of frameworks content, and the lists there don't know if they will be displayed in the middle or end of the page, hence the need to wrap them in the template here.
Otherwise we could just add the style to the list itself.

app/main/helpers/frameworks.py
samuelhwilliams:
```all_open_fwks = tuple(chain.from_iterable(
    islice(grp, 1)  # only want the first of each metaframework
    for _, grp in groupby(
        sorted(
            (fw for fw in frameworks if fw["status"] in ("open",)),
            key=lambda fw: fw["framework"],
        ),
        key=lambda fw: fw["framework"],
    )
))```

x_x

TheDoubleK: I've pushed an updated version of this, Finds the frameworks your way and then splits the list out to pass the separate things into the template, which keeps the template logic the same as it was.  Seems OK to me but see what you reckon?

tests/app/main/test_suppliers.py
TheDoubleK: Done 🍑

TheDoubleK:
Think this is ready to go.  Have updated screenshots to show latest (slightly adjusted) spacing.

PR #744, opened 2017-09-18
templates: use form and assurance macros from frontend toolkit instead of repo local copies

risicle:

There still exists in this repository an `app/templates/macros/forms.html` (which contains... "extensions" to the toolkit-provided macros?)

Do we think these should be folded into the frontend-toolkit too?


TheDoubleK:
OK so this is probably the point where this ticket becomes a "3"...
Looking at the diff between `app/templates/macros/toolkit_forms.html` that's deleted here and the `toolkit/forms/macros/forms.html` that's being dropped in as a replacement, there seem to be some non-trivial differences that mean I'm not 100% convinced that we can just swap-in the toolkit versions.

The main thing seems to be that the supplier app has `|question_references(kwargs.get_question)` on a lot of the values, which I *think* (but am not sure without further investigation) is the thing that makes the between-question references work for the supplier declaration.

So I think this needs thorough testing locally before we go any further - running through a complete G-Cloud 9 application, both declaration and services , should expose any regressions that this change would bring in.

If regressions exist (as I suspect they will), then up to you if you want to fix it all, or just shelve this ticket.

risicle:
Siiii-i-igh so I can't just rely on the tests covering it.

risicle:
The requirements for this are growing rapidly.

One of the main differences between the two files is that the supplier-internal versions filter their input through `question_references`. So with the idea that any app may want to use the question cross-reference feature, it was decided that feature should be included in the toolkit. However, that means the toolkit's macros will implicitly depend on a particular jinja filter which currently only the supplier frontend has. So we have to move the `question_reference` jinja filter to `-utils` and include it in the default selection of filters for this to continue to work across apps.

But `question_reference` has no unit tests, so moving it to `-utils` would require some unit tests written for me to feel ok about it. Thing is, that would require building a mock framework to be any use. And this would cause the `-utils` repo to have a dev-dependency on the content loader, which we've managed to avoid up to now.

PR #745, opened 2017-09-19
Closely pack "Become a supplier" paragraphs

TheDoubleK:
Pulls in new content from:
 - [x] https://github.com/alphagov/digitalmarketplace-frameworks/pull/471

## All closed
screen shot 2017-09-19 at 12 29 09

## All Open/Coming
screen shot 2017-09-19 at 12 29 43

## One of each
screen shot 2017-09-19 at 12 14 20

### DIFF
--- a/app/templates/suppliers/become_a_supplier.html
+++ b/app/templates/suppliers/become_a_supplier.html
@@ -54,9 +54,9 @@
               {{ content_loader.get_message(framework.slug, 'become-a-supplier', framework.status) }}
             </div>
           {% endfor %}
-          </div>
-        {% endif %}
 
+        {% endif %}
+        </div>
 
       {% else %}
         <div class="dmspeak">
--- a/bower.json
+++ b/bower.json
@@ -7,6 +7,6 @@
     "jquery-details": "https://github.com/mathiasbynens/jquery-details/archive/v0.1.0.tar.gz",
     "digitalmarketplace-frontend-toolkit": "https://github.com/alphagov/digitalmarketplace-frontend-toolkit.git#v24.1.1",
     "govuk_template": "https://github.com/alphagov/govuk_template/releases/download/v0.19.2/jinja_govuk_template-0.19.2.tgz",
-    "digitalmarketplace-frameworks": "https://github.com/alphagov/digitalmarketplace-frameworks.git#8.14.2"
+    "digitalmarketplace-frameworks": "https://github.com/alphagov/digitalmarketplace-frameworks.git#8.14.3"
   }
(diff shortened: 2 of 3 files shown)


samuelhwilliams:
I don't think we quote framework names anywhere else, do we? What's the justification for doing it here?

idavidmcdonald:
The content also looks a bit squished together?

samuelhwilliams:
And the links aren't very strongly associated to their 'parents'; could do with the top margin/border being smaller. :trollface:

TheDoubleK:
I've sat with Laurence this morning sorting this out.  He wants it squished together - the problem this fixes is that the content is too spread out.  As for quoting framework names, ¯\\_(ツ)_/¯

idavidmcdonald:
Also should what you can sell for g cloud be:

cloud hosting
cloud software
cloud support

As they are the lot names, at the moment we have something different (ambiguity over things not being cloud and the word technical now added)

TheDoubleK:
Current screenshots, for comparison, are here: 
https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/741

TheDoubleK:
Those words came from Karl, there was some discussion of them on the previous PR I think ¯\\(ツ)/¯

TheDoubleK:
This PR is specifically about spacing.

If we want to change the words that will be a new story.

TheDoubleK:
I'm going to hold off merging this as lots of people don't like it...

TheDoubleK:
@samuelhwilliams > "And the links aren't very strongly associated to their 'parents'"

@laurenceberry says the links aren't associated to a particular framework, so in the case of "both open" or "both closed" the link doesn't need to seem like it's only associated with the final framework listed on the page.

In the "one of each" case it is maybe a bit more room than we'd like, but maybe we live with that because it means we can use the link style straight out of the toolkit rather than create yet another variation.

samuelhwilliams:
My issue is primarily with the 'one of each' page. On the other two versions the link looks OK (i.e. when there's nothing after it with equal spacing).

Re: parents, I meant the heading/section rather than the framework. I understand that at some indeterminate point in the future we'll have more than two frameworks and in that case we don't want the link to stick to the final one.

TheDoubleK:
@samuelhwilliams the content community says the quotes should go.   You can make it happen by adding your thumb here: https://github.com/alphagov/digitalmarketplace-frameworks/pull/472

TheDoubleK:
Laurence has double-checked over this and he's happy so going to merge it.

PR #747, opened 2017-09-22
Move supplier user creation invite emails to notify

Wynndow:
Tests will need this PR on utils to be merged before going green: https://github.com/alphagov/digitalmarketplace-utils/pull/335

Part of this ticket: https://trello.com/c/HlSz9QZZ

### Pull in utils 28.5.0
It includes the shared code for sending user create/invite emails as
well as the external views. 

### Send supplier create emails with Notify
As part of the supplier sign up flow a user creation email is sent to a
given email address. This was done by Mandrill, now it's done by Notify.

The functionality to do this for supplier users and buyer users is
really similar so the code to do it has been abstracted to dmutils. We
import it here and use it.

### Remove old redirect for user creation
Supplier creation was moved to the user app. Any tokens created before
the move will have linked to the supplier app still. Tokens are valid
for one week, so we needed to maintain the functionality of those
tokens by keeping a redirect on the old route to the user app.

It's been over a week so all tokens will now link to the user app. We
can remove this redirect and associate test.

### Import external views from utils
External views are being moved to utils where they can be shared across
all frontend apps.

### Remove unused `invite_user_email` template
The templates are now managed by notify, so can be removed from here.


app/main/views/login.py
```diff
@@ -1,25 +1,16 @@
...
+from flask import flash, redirect, render_template, url_for, current_app
 
 from dmapiclient.audit import AuditTypes
-from dmutils.email import generate_token, send_email
-from dmutils.email.exceptions import EmailError
+from dmutils.email import send_user_account_email
 
 from .. import main
 from ..forms.auth_forms import EmailAddressForm
-from ..helpers import hash_email, login_required
```
allait: Is `hash_email` used by anything else or could we delete the helper function as well? There's a shared one in dmutils now that we should probably switch to if it is used.

Wynndow: It's used in app/main/view/frameworks, but could probable update to use the utils one. I'm going to check they have the same functionality though.

Wynndow: Yep, have removed the helper and used the one from utils.

PR #748, opened 2017-09-22
form macros: base most of these on the upstream toolkit's copy

risicle:

Depends on https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/375

Don't know how much more I need to say.

### DIFF
--- a/app/templates/macros/toolkit_forms.html
+++ b/app/templates/macros/toolkit_forms.html
@@ -1,300 +1,102 @@
+{% import "toolkit/forms/macros/forms.html" as upstream_forms %}
+
 {% macro text(question_content, service_data, errors) -%}
-  {%
-    with
-    question=question_content.question|question_references(kwargs.get_question),
-    question_advice=question_content.question_advice,
-    hint=(question_content.hint or '')|question_references(kwargs.get_question),
-    optional=question_content.optional,
-    name=question_content.id,
-    question_number=kwargs.question_number,
-    value=service_data[question_content.id],
-    unit_in_full=question_content.unit_in_full,
-    unit_position=question_content.unit_position,
-    unit=question_content.unit,
-    smaller=question_content.smaller,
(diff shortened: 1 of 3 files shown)


app/templates/macros/toolkit_forms.html
```diff
@@ -1,300 +1,102 @@
...
-    optional=question_content.optional,
-    id=question_content.id,
+  {{ upstream_forms.pricing(question_content, service_data, errors,
     error=errors.get(question_content.id)['message']|question_references(kwargs.get_question),
-    question_number=kwargs.question_number,
-    message=kwargs.message
-  %}
-    {% include "toolkit/forms/pricing.html" %}
-  {% endwith %}
+    hint=(question_content.hint or '')|question_references(kwargs.get_question),
```
TheDoubleK: `hint` and `question` didn't previously have the `|question_references` - why have you added it for this pricing macro, but not for multiquestion and boolean_list?  Do we know we need it here, or it just feels like we might?

TheDoubleK:
This looks good as far as I can see.  I think I'd like to give it a quick try-out once the toolkit version is merged, before thumbsing up.

app/templates/macros/toolkit_forms.html
risicle: Ah yes I did add that to some of the parameters for symmetry, assuming they were omissions. I can't think how including this feature generally across all similar fields could backfire too much, and it's the "least surprising" thing for the macros to do for a macro user.

risicle:
Please do - the quickest way of doing so is modifying `bower.json` to point to `https://github.com/alphagov/digitalmarketplace-frontend-toolkit.git#ris-form-text-field-add-unit-vars`

risicle:
Interesting that there were no other merges into this repo in the last 12 days so I didn't need to rebase.

PR #750, opened 2017-10-12
form macros: fix for boolean_list and multiquestion macros, add tests

risicle:

Turns out jinja doesn't appear to re-export imported macros after all, but there wasn't a test covering these question types. so add a slightly ugly passthrough hack for the macros and add a couple of (very basic) tests for editing of G9 services.


risicle:
Just going to try & push in some extra test assertions before merging...

PR #755, opened 2017-11-17
Suppliers can upload service definition documents

TheDoubleK:
Tests will fail until this is merged:
 - [x] https://github.com/alphagov/digitalmarketplace-frameworks/pull/482

For this ticket: https://trello.com/c/tZwe5CNc/138-suppliers-can-upload-documents-when-editing-g-cloud-services

This adds a new Documents section to the "edit service" view for G-Cloud 9 services, with a row for service definition document that is visible even if there is no document uploaded (see screenshot below).  Rows will continue to not be shown for empty questions that are not editable (as happens on the DOS version of this page).  Most of the new tests here mirror the tests we have for file uploads during the submissions process

intermediate improvement support digital marketplace 20171116

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -181,6 +181,7 @@ def update_section(framework_slug, service_id, section_id):
     posted_data = section.get_data(request.form)
 
     errors = None
+    # This utils method filters out any empty documents and validates against service document rules
     uploaded_documents, document_errors = upload_service_documents(
         s3.S3(current_app.config['DM_DOCUMENTS_BUCKET']),
         'documents',
@@ -538,6 +539,7 @@ def edit_service_submission(framework_slug, lot_slug, service_id, section_id, qu
         if request.files:
             uploader = s3.S3(current_app.config['DM_SUBMISSIONS_BUCKET'])
             documents_url = url_for('.dashboard', _external=True) + '/assets/'
+            # This utils method filters out any empty documents and validates against service document rules
             uploaded_documents, document_errors = upload_service_documents(
                 uploader, 'submissions', documents_url, draft, request.files, section,
                 public=False)
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -202,6 +202,7 @@ def update_section(framework_slug, service_id, section_id):
                 current_user.email_address)
         except HTTPError as e:
             errors = section.get_error_messages(e.message)
+            # If the service name is empty the final breadcrumb will disappear, so re-populate with the current name
             if not posted_data.get('serviceName', None):
                 posted_data['serviceName'] = service.get('serviceName', '')
 
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -180,25 +180,40 @@ def update_section(framework_slug, service_id, section_id):
 
     posted_data = section.get_data(request.form)
 
-    try:
-        data_api_client.update_service(
-            service_id,
-            posted_data,
-            current_user.email_address)
-    except HTTPError as e:
-        errors = section.get_error_messages(e.message)
-        if not posted_data.get('serviceName', None):
-            posted_data['serviceName'] = service.get('serviceName', '')
+    errors = None
+    uploaded_documents, document_errors = upload_service_documents(
+        s3.S3(current_app.config['DM_DOCUMENTS_BUCKET']),
+        'documents',
+        current_app.config['DM_ASSETS_URL'],
+        service,
+        request.files,
+        section,
+    )
+    if document_errors:
+        errors = section.get_error_messages(document_errors)
+    else:
+        posted_data.update(uploaded_documents)
+
+    if not errors and section.has_changes_to_save(service, posted_data):
+        try:
+            data_api_client.update_service(
+                service_id,
+                posted_data,
+                current_user.email_address)
+        except HTTPError as e:
+            errors = section.get_error_messages(e.message)
+            if not posted_data.get('serviceName', None):
+                posted_data['serviceName'] = service.get('serviceName', '')
+
+    if errors:
         return render_template(
             "services/edit_section.html",
             section=section,
             service_data=posted_data,
             service_id=service_id,
             errors=errors,
-        )
-
+        ), 400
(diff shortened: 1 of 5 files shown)


tests/app/main/test_services.py
```diff
@@ -1030,6 +1058,67 @@ def test_questions_for_this_service_section_can_be_changed(self, data_api_client
...
+
+    def test_upload_question_can_not_be_set_by_form_data(self, data_api_client, s3):
+        data_api_client.get_service.return_value = self.base_service
+        res = self.client.post(
+            '/suppliers/frameworks/g-cloud-9/services/321/edit/documents',
+            data={
+                'serviceDefinitionDocumentURL': 'http://example.com/document.pdf',
+            })
+
+        assert res.status_code == 302
```
katstevens: Don't these three cases (no files, empty/unknown files, url in form data) count as 'errors'? Why do they return 302 instead of 400?

tests/app/main/test_services.py
```diff
@@ -544,6 +544,32 @@ def test_should_view_public_service_with_update_message(self, data_api_client):
...
+        )
+
+        res = self.client.get("/suppliers/frameworks/{}/services/123".format(self.framework_kwargs["framework_slug"]))
+
+        assert res.status_code == 200
+        # There's no serviceDefinitionDocumentURL in the test service data, but the row should still be shown
+        self.assert_in_strip_whitespace(
+            'Service definition document',
+            res.get_data(as_text=True)
+        )
```
katstevens: Should we assert the empty message itself, as well as the label?

app/main/views/services.py
```diff
@@ -180,25 +180,40 @@ def update_section(framework_slug, service_id, section_id):
...
-            service_id,
-            posted_data,
-            current_user.email_address)
-    except HTTPError as e:
-        errors = section.get_error_messages(e.message)
-        if not posted_data.get('serviceName', None):
-            posted_data['serviceName'] = service.get('serviceName', '')
+    errors = None
+    uploaded_documents, document_errors = upload_service_documents(
+        s3.S3(current_app.config['DM_DOCUMENTS_BUCKET']),
```
katstevens: It seems odd to me that the FE apps should have to know about S3. Could the `upload_service_documents` util not just pass in the `current_app.config` and the util function deal with whatever storage implementation we happen to have? Not suggesting we change that now, but maybe one for the tech debt pile...

app/main/views/services.py
```diff
@@ -180,25 +180,40 @@ def update_section(framework_slug, service_id, section_id):
...
+    if not errors and section.has_changes_to_save(service, posted_data):
+        try:
+            data_api_client.update_service(
+                service_id,
+                posted_data,
+                current_user.email_address)
+        except HTTPError as e:
+            errors = section.get_error_messages(e.message)
+            if not posted_data.get('serviceName', None):
+                posted_data['serviceName'] = service.get('serviceName', '')
```
katstevens: I know this chunk was in the previous code, but it's not obvious at all what it's meant to be doing. I get that it's for rendering DOS services that don't have a name, but why are we only setting it in the API error case? Maybe an explanatory comment would help 😸

tests/app/main/test_services.py
TheDoubleK: Yes, probably!

app/main/views/services.py
TheDoubleK: I agree wholeheartedly that this isn't for now :grinning:

app/main/views/services.py
TheDoubleK: If I can work out the point of it I'll add a comment 😛

tests/app/main/test_services.py
TheDoubleK: That's a good question that I don't necessarily have a good answer to right now.  The trying-to-pass-in-in-as-a-key thing is a weird one - because it's a valid key with a valid value then I'm not sure how we'd make it provoke a validation error.  The ignoring of empty/unknown files is existing behaviour in service submissions that someone had a reason for at some point... it is a bit weird though, people might want to be alerted if they try to upload an empty file. Maybe I'll check the history to see if I can find some rationale...

app/main/views/services.py
TheDoubleK: Hmm, looks like I added this back in 2015.  So it must be Important and Useful.

TheDoubleK: And I've remembered what it's for - if someone tries to delete the service name then we want to re-populate the field with the existing service name, rather than just show them an empty box.  It seemed worth it at the time - we really don't want people to have services without a name.

TheDoubleK: Aha - yes, this was more important when the editing was all on one page (as it used to be) rather than one question per page.  The service name used to be the heading for the page, with all the editable fields underneath it. So then if they had deleted their service name and got validation errors the heading disappeared from the page and they didn't know what service they were editing any more.  So this ensured that there was a heading on the page and it was obvious which service they are editing. 

But it still applies for the breadcrumbs now - the final breadcrumb is the service name, so if this has been deleted then the breadcrumb disappears on the validation error page.

tests/app/main/test_services.py
TheDoubleK: If you can live with it then I'm very tempted to stick with the existing ("legacy") behaviour and leave it as it is for now.  But if you think I'm just being lazy then tell me to fix it.

katstevens: Happy to leave it as long as it's documented (my rule of thumb is that if it's not obvious from the code what's happening or an unusual decision has been made, stick an explanatory comment in). This should save us having this same conversation again the next time we touch this code 😸

app/main/views/services.py
katstevens: Makes sense - happy with this comment 👍

tests/app/main/test_services.py
TheDoubleK: I added a comment.  Think this should be good to go now?

TheDoubleK:
PR up for functional tests to cover this: https://github.com/alphagov/digitalmarketplace-functional-tests/pull/412/files

PR #758, opened 2017-12-11
Remove unused CSS

samuelhwilliams:
## Summary
Remove unused CSS from the supplier-frontend, re-order `application.scss` to use the template/form decided upon in https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/672.

## Ticket
https://trello.com/c/iPHuyh35/78-remove-old-scss-supplier-front-end


app/templates/frameworks/contract_start.html
```diff
@@ -70,7 +70,7 @@
       </div>
 
       <div>
-        <h2 class="page-subheading">What you need to do</h2>
+        <h2 class="summary-item-heading">What you need to do</h2>1
```
MuriloDalRi: Typo

samuelhwilliams: 👍

app/assets/scss/application.scss
```diff
@@ -91,42 +91,6 @@ $path: "/suppliers/static/images/";
...
-  display: block;
-  @include core-19;
-  margin: (1.5 * $gutter) 0 (2 * $gutter);
-}
-
 .padding-bottom-small {
   padding-bottom: 10px;
 }
-
-ins {
```
idavidmcdonald: So technically I don't think we are using these two (which are used for showing a contract variation) as g cloud 8 is no longer live and is the only framework that has a contract variation at the moment.

However the route still exists to show contract variations, and if we needed one for g10 then we would need to re-add this css. 

https://github.com/alphagov/digitalmarketplace-frameworks/blob/bb4e019354f7308707d1cf5133e19584cf0f1911/frameworks/g-cloud-8/messages/contract_variation_1.yml

Thoughts?

Of note, we also use a 'diffing' style in the admin app but it uses different styling (obviously) but it feels like something that should be moved in to the toolkit at some point in the future.

app/assets/scss/application.scss
```diff
@@ -48,35 +34,49 @@ $path: "/suppliers/static/images/";
 @import "toolkit/_secondary-action-link.scss";
 @import "toolkit/_service-id.scss";
 @import "toolkit/_temporary-message.scss";
+@import "toolkit/forms/_combinations.scss";
+@import "toolkit/forms/_checkbox-tree.scss";  // Has to be imported after _combinations.scss, which defines the
+                                              // $border-width variable that _checkbox-tree.scss relies upon. :sad:
```
idavidmcdonald: Well that's a shame. Good spot though. I think we shouldn't have one of our partials depending on a variable set in another partial. _checkbox-tree.scss should likely be defining it's own $border-width variable if it needs one (or if it is truely a shared variable then should be stored in a different file and imported in by partials that need it). If you agree, could you write a ticket at the bottom of our icebox for this please :)

app/assets/scss/_updates.scss
```diff
@@ -1,7 +1,3 @@
-#input-clarification_question {
-  height: 20.75em;
-}
```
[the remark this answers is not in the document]
samuelhwilliams: The class is built in the brief-responses frontend through a combination of https://github.com/alphagov/digitalmarketplace-brief-responses-frontend/blob/master/app/main/views/briefs.py#L73 which passes through to https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/toolkit/templates/forms/textbox.html#L64. The same thing doesn't appear to exist any more in supplier-frontend (presumably because the functionality was moved to brief-responses).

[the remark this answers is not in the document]
idavidmcdonald: We may be using it for when a buyer adds a brief clarification question which is likely in this application (they get sent them by email from the supplier form and then fill in what the question is and what the answer is to post it). 

Regardless, it's unlikely this class is being used.

app/assets/scss/application.scss
samuelhwilliams: Yep, ticket made.

app/assets/scss/application.scss
samuelhwilliams: Generic styles on core HTML element types should _definitely_ be in the toolkit, yeah. In light of that, and under the belief that we don't have anything currently using the style, I think we should still opt to remove it for now.

idavidmcdonald: Hmm I'm a little torn. Whilst it feels good to remove it (which can then prompt it to be written in the toolkit when we next need a contract variation), it feels a bit wrong to remove css that an html template of ours is relying on and the route still exists (it just happens that it's not in use at the moment).

app/assets/scss/_updates.scss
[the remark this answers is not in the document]
samuelhwilliams: We're using it here https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/templates/frameworks/updates.html#L133 where suppliers can ask clarification questions about a framework.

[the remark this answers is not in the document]
idavidmcdonald: Ah cool, so we do have an area that it seems appropriate for but technically we aren't using the class due to a underscore mismatch (for which we can't tell if is intentional or unintentional)?

app/assets/scss/application.scss
samuelhwilliams: Is a reasonable point. I could go either way.

Probably the latter (keep it) now that you spell it out.

idavidmcdonald: I've added a ticket at the bottom of our icebox (https://trello.com/c/N712MbsT/106-move-ins-and-dels-diff-styling-into-the-toolkit). Is it worth adding an inline comment? (feel free to say no, I'm not sure myself).

samuelhwilliams: Given adding an inline comment probably takes less time it took you to write this reply, I'd say sure, it can't really not be worth doing. ;)

samuelhwilliams:
OK, should have addressed all your comments @idavidmcdonald

app/assets/scss/_updates.scss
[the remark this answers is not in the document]
samuelhwilliams: Huh, no, I'm saying that **_we are_** using this ID in the supplier app.

screen shot 2017-12-12 at 12 10 11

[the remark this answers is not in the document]
idavidmcdonald: Ah OK, thanks I understand now.

samuelhwilliams:
Ta. I've manually eye-balled it and I've also run all the main routes through local custom visual regression, which highlights no changes. Albeit it won't catch everything (like different versions of the same page) but it should have given good coverage of most of the routes. And it all looked fine.

PR #759, opened 2017-12-12
Reset styles before DM CSS

samuelhwilliams:
Move our DM CSS reset to the top of our `application.css` file so that we clear styling on all HTML core elements, giving us a definitively clean base upon which to buildn.

### DIFF
--- a/app/assets/scss/application.scss
+++ b/app/assets/scss/application.scss
@@ -1,3 +1,7 @@
+// Reset styles on core HTML elements that might be injected by eg a user's browser, so that we have a known clean base
+// for DM styles to build on.
+@import "toolkit/shared_scss/_reset.scss";
+
 // GOVUK Front-end toolkit variables, mixins and functions
 @import "_grid_layout.scss";
 @import "_typography.scss";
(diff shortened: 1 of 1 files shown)


app/assets/scss/application.scss
```diff
@@ -1,3 +1,6 @@
+// Reset styles on core HTML elements so that we have a clean base for DM styles to build on.
```
idavidmcdonald: Nice, might be nice to add something about browser added styles as those are the reason we have a reset stylesheet. Apart from that I like this

samuelhwilliams: Comment clarified. 👍

app/assets/scss/application.scss
```diff
@@ -1,3 +1,7 @@
+// Reset styles on core HTML elements that might be injected by eg a user's browser, so that we have a known clean base
```
idavidmcdonald: Sorry, should it be `that might be injected eg by a user's browser`?

samuelhwilliams: Either/or is fine.

PR #760, opened 2017-12-13
Services can only be edited while framework is live

TheDoubleK:
Addresses this bug: https://trello.com/c/6mO70EhR/186-editing-of-services-from-expired-frameworks

We currently restrict the list of services shown to suppliers to live frameworks, but if the URL is manipulated to have the ID of a service on an expired framework it could still be edited.  This changes things to prevent that from happening.

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -6,10 +6,10 @@ from ...main import main, content_loader
 from ..helpers import login_required
 from ..helpers.services import is_service_associated_with_supplier, get_signed_document_url, count_unanswered_questions
 from ..helpers.frameworks import (
-    get_framework_and_lot,
+    get_framework_and_lot_or_404,
     get_declaration_status,
     get_supplier_framework_info,
-    get_framework,
+    get_framework_or_404,
 )
 
 from dmcontent.content_loader import ContentNotFoundError
@@ -21,7 +21,7 @@ from dmutils.documents import upload_service_documents
 @main.route("/frameworks/<string:framework_slug>/services")
 @login_required
 def list_services(framework_slug):
-    framework = get_framework(data_api_client, framework_slug, allowed_statuses=['live'])
+    framework = get_framework_or_404(data_api_client, framework_slug, allowed_statuses=['live'])
 
     suppliers_services = data_api_client.find_services(
         supplier_id=current_user.supplier_id,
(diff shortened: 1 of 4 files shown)


app/main/views/services.py
```diff
@@ -85,6 +85,9 @@ def remove_service(framework_slug, service_id):
     if service["frameworkFramework"] == 'digital-outcomes-and-specialists':
         abort(404)
 
+    # we don't actually need the framework here; using this to 404 if framework for the service is not live
+    get_framework(data_api_client, service['frameworkSlug'], allowed_statuses=['live'])
```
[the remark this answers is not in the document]
TheDoubleK: It already exists; I decided not to refactor right now.  I can do if you reckon it's worth it?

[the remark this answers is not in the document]
samuelhwilliams: Probably not as part of this, there seem to be quite a few related functions that could also be renamed to do the `_or_404` thing. I think it's OK not to do it as part of this. But if you want to... by all means.

tests/app/main/test_services.py
```diff
@@ -823,6 +886,16 @@ def test_should_fail_if_incorrect_framework(
 @mock.patch('app.main.views.services.data_api_client', autospec=True)
 class TestSupplierEditUpdateServiceSection(BaseApplicationTest):
 
+    def _get_framework_response(self, **kwargs):
+        framework = {
+            "framework": "g-cloud",
+            "name": "G-Cloud 6",
+            "slug": "g-cloud-6",
+            "status": "live",
```
samuelhwilliams: You could use fun new Python 3 syntax here:

```
framework = {"framework": "g-cloud", "name": "G-Cloud 6", "slug": "g-cloud-6", "status": "live", **kwargs}
            
```

TheDoubleK: Will a status in `kwargs` there overwrite the earlier `status`?

samuelhwilliams: Yep! It follows left-to-right order, so:
`{a: 1, **{a: 2}} == {a: 2}`
and
`{**{a: 2}, a: 1} == {a: 1}`

app/main/views/services.py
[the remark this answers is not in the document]
TheDoubleK: Done, with bonus renaming of other methods in the file.

tests/app/main/test_services.py
TheDoubleK: Nice - it's neater like this - done.

PR #761, opened 2017-12-13
join_open_framework_notification_mailing_list: return 400 status code in some mailchimp call failure cases

risicle:
Story https://trello.com/c/g93M7T3f/104-503-responses-for-bad-email-addresses-should-be-400s

Those cases being the ones where we think the failure might somehow be due to the user's email address.

Debatable whether we want to apply this or remain alert to these failures. Note this still doesn't completely quieten these events - the `DMMailChimpClient` from `-utils` itself emits a log message, which we may want to downgrade to a warning in this, less severe, case. But that's a separate issue.

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -513,7 +513,12 @@ def join_open_framework_notification_mailing_list():
             else:
                 # failure
                 flash("mailing_list_signup_error", "error")
-                status = 503
+                if mc_response:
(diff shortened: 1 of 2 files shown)


katstevens:
I know the DM mandrill client does some logging - is there enough detail there or do we need explicit log messages here to highlight the difference in the errors?

risicle:
The logging in `-utils` already knows pretty much all we know about the failure. I'm not suggesting adding log messages here.

PR #762, opened 2017-12-19
Remove bower and upgrade dependencies

MuriloDalRi:
* Remove bower as a dependency and use yarn to manage our frontend dependencies
* Move all our bower dependencies to package.json and update them
* Delete bootstrap.sh and run_tests.sh scripts in favour of the makefile
* Bring node version into line with production
* Update makefile for new yarn commands
* Update README with yarn instructions


samuelhwilliams:
## Added
Gulp silently ignores missing imports without the 'hardFail' option. Let's add it to make sure that every file we want is actually imported.

PR #767, opened 2018-01-08
Pull in latest frameworks version

TheDoubleK:
This will allow us to open G10 and DOS3 on Preview.

The major version bump (9.x to 10.x) is to do with search mappings that are not used by this app, so it's OK to pull in without any further code changes.


benvand:
LGTM 👍

PR #774, opened 2018-01-16
Tweak framework tests in supplier frontend

benvand:
Tidy up 2 tests:
* Remove extraneous, unhelpful assertion and use `self.client`
* Mock session instead of forcing a new context


benvand:
Thanks @CrystalPea and @katstevens ! 🎉

PR #777, opened 2018-01-18
Improve documentation around custom form macros found in this app

idavidmcdonald:
https://trello.com/c/iaRQlj6y/87-toolkit-forms-macros-are-different-in-all-of-the-apps

I originally was hoping to remove the custom `toolkit_forms` macros file from this repo so that instead it could just use the the macros found in the FE toolkit so we had consistency between the apps. However upon on review, we have functionality in this app that relies on these custom macros (question references being parsed for declaration questions and nice date values on file uploads for service editing).

There were however a few small things that I could tidy up
- removed two lines of code from the toolkit_forms macros that were having no effects
- Improved documentation
- Added test coverage

Current behaviour for upload time:
image

Behaviour that would have been changed if we used the toolkit macros rather than the app macros.
image

### DIFF
--- a/app/main/helpers/frameworks.py
+++ b/app/main/helpers/frameworks.py
@@ -154,6 +154,16 @@ def return_supplier_framework_info_if_on_framework_or_abort(data_api_client, fra
 
 
 def question_references(data, get_question):
+    """
+    Replace placeholders for question references with the number of the referenced question
+
+    e.g. "This is my question hint which references question [[anotherQuestion]]" becomes
+    "This is my question hint which references question 7"
+
+    :param data: Object to have placeholders replaced for example a string or Markup object
+    :param get_question: ContentManifest.get_question function
+    :return: Object with same type of original `data` object but with question references replaced
+    """
     if not data:
         return data
     references = re.sub(
--- a/app/templates/macros/toolkit_forms.html
+++ b/app/templates/macros/toolkit_forms.html
@@ -22,7 +22,6 @@
   {{ upstream_forms.list(question_content, service_data, errors,
     error=errors.get(question_content.id)['message']|question_references(kwargs.get_question),
     hint=(question_content.hint or '')|question_references(kwargs.get_question),
-    number_of_items=10,
     question=question_content.question|question_references(kwargs.get_question),
     **kwargs
   ) }}
--- a/app/templates/macros/toolkit_forms.html
+++ b/app/templates/macros/toolkit_forms.html
@@ -40,7 +40,6 @@
   {{ upstream_forms.radios(question_content, service_data, errors,
     error=errors.get(question_content.id)['message']|question_references(kwargs.get_question),
     hint=(question_content.hint or '')|question_references(kwargs.get_question),
-    hint_underneath=(question_content.hint_underneath or False),
     question=question_content.question|question_references(kwargs.get_question),
     **kwargs
   ) }}
--- a/app/templates/macros/toolkit_forms.html
+++ b/app/templates/macros/toolkit_forms.html
@@ -1,3 +1,18 @@
+{#
+The following form macros extend on top of of the macros found in the front end toolkit so we can support additional
+functionality needed for forms for applying for a framework.
+
+Because questions for the supplier framework declaration form have hints, questions and errors that are supposed to
+reference the numbers of other questions, after we get the manifest (which allows us to know the number for each
+question) from the content loader, we must replace the placeholders for question references with their corresponding
+question number. This is done using the `question_references` helper. This behaviour does not seem to be needed for any
+of our other forms. Note, it may be an idea in the future to move this functionality into the content loader as it
+doesn't feel quite right being at the templating level.
+
+So we can format the value of a form upload to be a 'pretty' datetime string for file uploads for editing a service, we
+set a custom filter on its value.
+#}
+
 {% import "toolkit/forms/macros/forms.html" as upstream_forms %}
 
 {% macro text(question_content, service_data, errors) -%}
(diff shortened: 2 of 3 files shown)


app/templates/macros/toolkit_forms.html
```diff
@@ -1,3 +1,14 @@
+{#
+The following form macros extend on top of of the macros found in the front end toolkit so we can support additional
+functionality needed for forms for applying for a framework.
+
+So we can parse question references for the supplier framework declaration forms, we redefine all our individual
+questions here with parsed error messages, hints and question.
+
+So we can format the value of a form upload to be a 'pretty' datetime string for file uploads for editing a
+service, we set a custom filter on it's value.
```
galund: spelling - "its" :)

idavidmcdonald: Fixed, good spot

app/main/helpers/frameworks.py
```diff
@@ -154,6 +154,12 @@ def return_supplier_framework_info_if_on_framework_or_abort(data_api_client, fra
 
 
 def question_references(data, get_question):
+    """
+    Replace placeholders for question references with the number of the referenced question
+
+    e.g. "This is my question hint which references [[anotherQuestion]]" becomes
+    "This is my question hint which references 7"
+    """
```
risicle: It would be nice to document the *weird bit* of this function, which is the `get_question` argument - expected to be a bound callable from what i can remember.

app/templates/macros/toolkit_forms.html
```diff
@@ -65,32 +74,32 @@
   ) }}
 {%- endmacro %}
 
-{% macro upload(question_content, service_data, errors) -%}
-  {{ upstream_forms.upload(question_content, service_data, errors,
+{% macro pricing(question_content, service_data, errors) -%}
+  {{ upstream_forms.pricing(question_content, service_data, errors,
```
risicle: Is this reordering to match the order in another file? If there isn't a clear good reason for it I'd argue it's best not to muddy the blame history of the file.

app/templates/macros/toolkit_forms.html
```diff
@@ -1,3 +1,14 @@
+{#
+The following form macros extend on top of of the macros found in the front end toolkit so we can support additional
+functionality needed for forms for applying for a framework.
+
+So we can parse question references for the supplier framework declaration forms, we redefine all our individual
```
galund: It definitely feels like this is something that doesn't really belong in the templating layer. But I don't really quite understand well enough what it's doing - why do we parse the question when displaying it? If there's _some_ further clarification of what's going on, that we could put in the comments here, then that would be nice.

app/templates/macros/toolkit_forms.html
```diff
@@ -1,3 +1,14 @@
...
+questions here with parsed error messages, hints and question.
+
+So we can format the value of a form upload to be a 'pretty' datetime string for file uploads for editing a
+service, we set a custom filter on its value.
```
galund: We absolutely ought to try and be consistent in how we display dates and times, so this one seems like a less good reason for us to have something custom here.

galund: (Oh no, wait what: it's extracting a date from a URI?? I don't even want to know. *clicks approve*)

app/templates/macros/toolkit_forms.html
idavidmcdonald: I reordered it so the one macro that had the unique kwargs (upload having an additional value kwarg) would be somewhat separate.

After your comment and in retrospect I'm going to undo that change.

app/main/helpers/frameworks.py
idavidmcdonald: Good point, will add

app/templates/macros/toolkit_forms.html
idavidmcdonald: Good point, it feels like this logic should be done by the content loader in preparing the content ready to be passed to the template. Not too sure how hard it would be to move the logic into the content loader but may not be a simple job.

I'll have a quick think but I'll probably not try and move it in as it doesn't feel like a simple job (and if I don't then I'll add a tech debt ticket about it). However I will try and fill out some more details in the documentation here.

idavidmcdonald:
I've started investigating moving the question referencing functionality out of templating and into the content loader. I've put up a strawman so you can see my workings developing (and warn me off if you think it's a bad path to go down or encourage me if you think it's good).

https://github.com/alphagov/digitalmarketplace-content-loader/pull/55

app/templates/macros/toolkit_forms.html
idavidmcdonald: Will rebase on top of this with some even better comments :)

idavidmcdonald:
Cool. I've improved commenting for both of you. I hope it's better (although feel free to let me know if it's not).

Whilst I have started investigating moving functionality into the content loader I do see several potential barriers along the way. Because of that I am going to shortly merge this PR regardless as it should add value right now, won't things harder if we do move functionality to the content loader and is ready to go.

app/templates/macros/toolkit_forms.html
```diff
@@ -1,3 +1,13 @@
+{#
+The following form macros extend on top of of the macros found in the front end toolkit so we can support additional
+functionality needed for forms for applying for a framework.
+
+Because questions for the supplier framework declaration form have hints, questions and errors that are supposed to reference the numbers of other questions, after we get the manifest (which allows us to know the number for each question) from the content loader, we must replace the placeholders for question references with their corresponding question number. This is done using the `question_references` helper. This behaviour does not seem to be needed for any of our other forms. Note, it may be an idea in the future to move this functionality into the content loader as it doesn't feel quite right being at the templating level.
```
galund: Line breaks please?

idavidmcdonald: Thanks, will do

PR #779, opened 2018-01-19
Pull in latest utils

TheDoubleK:
This fixes the S3 files-not-showing-if-no-timestamp bug: 
https://trello.com/c/qBiGn3bP/9-legal-documents-files-not-appearing-on-framework-application-dashboard

## Update

The (unintentional) updating of Werkzeuz caused a test to fail (and, it turns out, the app to break) - see my commit message on new commit, copied here for ease of reference:
```
Running `make freeze-requirements` has, as well as adding utils and ODFPy
updates, pulled in a newer version of Werkzeug (it's gone from 0.12.2 to
0.14.1).  This dependency comes from the Flask libraries specification of
`required: >=0.7`, but pip pulls in the latest available at the time it
runs.

The new version has new behaviour for file uploads. Previously, if no file
 was uploaded Werkzeug would pass an empty FileStorage object to the route
 under the `signature_page` key, but new behaviour is for that key not to
 exist at all under `request.files`.

 The changes here mean that the page will not break if a user submits the
 page with no new file chosen.
```

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -961,23 +961,28 @@ def signature_upload(framework_slug, agreement_id):
 
     agreements_bucket = s3.S3(current_app.config['DM_AGREEMENTS_BUCKET'])
     agreement_path = agreement.get('signedAgreementPath')
-    signature_page = agreements_bucket.get_key(agreement_path) if agreement_path else None
+    existing_signature_page = agreements_bucket.get_key(agreement_path) if agreement_path else None
     upload_error = None
-
     if request.method == 'POST':
+        fresh_signature_page = request.files.get('signature_page')
+
         # No file chosen for upload and file already exists on s3 so can use existing and progress
-        if not request.files['signature_page'].filename and signature_page:
+        if not (fresh_signature_page and fresh_signature_page.filename) and existing_signature_page:
             return redirect(url_for(".contract_review", framework_slug=framework_slug, agreement_id=agreement_id))
 
-        if not file_is_image(request.files['signature_page']) and not file_is_pdf(request.files['signature_page']):
+        # Validate file
+        if not fresh_signature_page:
+            upload_error = "You must choose a file to upload"
+        elif not file_is_image(fresh_signature_page) and not file_is_pdf(fresh_signature_page):
             upload_error = "The file must be a PDF, JPG or PNG"
-        elif not file_is_less_than_5mb(request.files['signature_page']):
+        elif not file_is_less_than_5mb(fresh_signature_page):
             upload_error = "The file must be less than 5MB"
-        elif file_is_empty(request.files['signature_page']):
+        elif file_is_empty(fresh_signature_page):
             upload_error = "The file must not be empty"
 
+        # If all looks good then upload the file and proceed to next step of signing
         if not upload_error:
-            extension = get_extension(request.files['signature_page'].filename)
+            extension = get_extension(fresh_signature_page.filename)
             upload_path = generate_timestamped_document_upload_path(
                 framework_slug,
                 current_user.supplier_id,
@@ -986,7 +991,7 @@ def signature_upload(framework_slug, agreement_id):
             )
             agreements_bucket.save(
                 upload_path,
-                request.files['signature_page'],
+                fresh_signature_page,
                 acl='private',
                 download_filename='{}-{}-{}{}'.format(
                     sanitise_supplier_name(current_user.supplier_name),
@@ -1000,7 +1005,7 @@ def signature_upload(framework_slug, agreement_id):
             data_api_client.update_framework_agreement(agreement_id, {"signedAgreementPath": upload_path},
                                                        current_user.email_address)
 
-            session['signature_page'] = request.files['signature_page'].filename
+            session['signature_page'] = fresh_signature_page.filename
 
             return redirect(url_for(".contract_review", framework_slug=framework_slug, agreement_id=agreement_id))
 
@@ -1008,7 +1013,7 @@ def signature_upload(framework_slug, agreement_id):
         "frameworks/signature_upload.html",
         agreement=agreement,
         framework=framework,
-        signature_page=signature_page,
+        signature_page=existing_signature_page,
         upload_error=upload_error,
     ), 400 if upload_error else 200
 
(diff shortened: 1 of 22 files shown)


tests/app/main/test_frameworks.py
```diff
@@ -7,14 +7,11 @@
...
 
-try:
-    from StringIO import StringIO
-except ImportError:
-    from io import BytesIO as StringIO
 import mock
 import pytest
 from six.moves.urllib.parse import urljoin
 
+from io import BytesIO
```
idavidmcdonald: You could tidy up the order of imports for these files please

https://www.python.org/dev/peps/pep-0008/#imports

```
Imports should be grouped in the following order:

standard library imports
related third party imports
local application/library specific imports
You should put a blank line between each group of imports.
```

TheDoubleK: Why are you complaining when PEP8 isnt? 😛

idavidmcdonald: https://github.com/PyCQA/flake8-import-order Don't make me install the flake8 plugin... (although I am getting tempted)

katstevens: Do it! I am bad at remembering to tidy imports...

idavidmcdonald: Ha, I thought it would be Ben who would spot it...

I'll take a quick glance properly. Although initially it looks like it requires more config than I expected (I don't think it runs straight out of the box).

TheDoubleK: Don't fret, I've done it and also removed Six from the app while I was at it.

app/main/views/frameworks.py
```diff
@@ -961,23 +961,28 @@ def signature_upload(framework_slug, agreement_id):
...
+        fresh_signature_page = request.files.get('signature_page')
+
         # No file chosen for upload and file already exists on s3 so can use existing and progress
-        if not request.files['signature_page'].filename and signature_page:
+        if not (fresh_signature_page and fresh_signature_page.filename) and existing_signature_page:
             return redirect(url_for(".contract_review", framework_slug=framework_slug, agreement_id=agreement_id))
 
-        if not file_is_image(request.files['signature_page']) and not file_is_pdf(request.files['signature_page']):
+        # Validate file
+        if not fresh_signature_page:
```
idavidmcdonald: Just to check, this change is because you spotted that we aren't covering this case currently, right?

tests/app/main/test_frameworks.py
```diff
@@ -4284,6 +4277,30 @@ def test_upload_signature_page(
         assert res.status_code == 302
         assert res.location == 'http://localhost/suppliers/frameworks/g-cloud-8/234/contract-review'
 
+    @mock.patch('dmutils.s3.S3')
+    @mock.patch('app.main.views.frameworks.file_is_empty')
```
idavidmcdonald: I don't think we need to be mocking `file_is_empty` given that we shouldn't reach that part of logic in the view?

tests/app/main/test_frameworks.py
```diff
@@ -4284,6 +4277,30 @@ def test_upload_signature_page(
...
+    ):
+        self.login()
+
+        data_api_client.get_framework.return_value = get_g_cloud_8()
+        data_api_client.get_framework_agreement.return_value = self.framework_agreement()
+        return_supplier_framework.return_value = self.supplier_framework(
+            framework_slug='g-cloud-8',
+            on_framework=True
+        )['frameworkInterest']
+        s3.return_value.get_key.return_value = None
```
idavidmcdonald: I assume this set up is just copied from another test but a small thing that might improve readability would be an inline comment saying why we are setting the s3 return value as such ie. `no existing file`

requirements-app.txt
```diff
@@ -5,6 +5,7 @@ Flask==0.10.1
 Flask-Login==0.2.11
 Flask-WTF==0.12
 
-git+https://github.com/alphagov/digitalmarketplace-utils.git@31.1.0#egg=digitalmarketplace-utils==31.1.0
+git+https://github.com/alphagov/digitalmarketplace-utils.git@31.1.1#egg=digitalmarketplace-utils==31.1.1
+git+https://github.com/alphagov/odfpy.git@ee4482a#egg=odfpy==1.3.6dev
```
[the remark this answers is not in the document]
TheDoubleK: Yes it should have been done in a previous bump. It isn't used, but utils expecs it to be there I think?

app/main/views/frameworks.py
TheDoubleK: This is because of the new Werkzeug behaviour of (apparently) not passing empty files through to the view in the request. Previously if the button was pressed to POST to this route with no file selected then `request.files['signature_page']` existed in the view as an empty FileStorage object.  This has now changed and it strips the key out from the request altogether, so we need to deal with the case that the key doesn't exist.

tests/app/main/test_frameworks.py
```diff
@@ -1,26 +1,19 @@
...
+import mock
+import pytest
 from collections import OrderedDict
+from io import BytesIO
 from itertools import chain
-
-from dmapiclient import HTTPError
+from lxml import html
+from urllib.parse import urljoin
 from werkzeug.datastructures import MultiDict
```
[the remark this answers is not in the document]
TheDoubleK: Should be in requirements-dev?  It's only used for tests, not in the app.  I think...

tests/app/main/test_frameworks.py
TheDoubleK: Yep this follows the same pattern from other tests on this route.

app/main/views/frameworks.py
```diff
@@ -3,28 +3,27 @@
...
 
 from dateutil.parser import parse as date_parse
-from dmapiclient import HTTPError
 from flask import render_template, request, abort, flash, redirect, url_for, current_app, session
 from flask_login import current_user
 import flask_featureflags as feature
-import six
 
 from dmapiclient import APIError
+from dmapiclient import HTTPError
```
katstevens: Can we put this import in the line above?

TheDoubleK: Good spot.  Will do.

PR #785, opened 2018-02-08
Update content for DUNS number

samuelhwilliams:
## Summary
Update the content when asking suppliers for their DUNS number. Also add breadcrumbs to the DUNS number page and the 'create new supplier' page, which doesn't have the breadcrumb it ought to (according to the prototype). We've had to change from `single-question-page` to `dmspeak` on the top-level div for two reasons.
1) `single-question-page` hides the "DUNS number" header above the hint text.
2) `dmspeak` adds some padding around the paragraphs, which are otherwise too tightly packed.

## Before
screen shot 2018-02-08 at 12 55 24

## After
screen shot 2018-02-08 at 14 08 08

## Ticket
https://trello.com/c/ae5UzOAn/39-sign-up-flow-update-copy-on-duns-number-page

### DIFF
--- a/app/templates/suppliers/create_new_supplier.html
+++ b/app/templates/suppliers/create_new_supplier.html
@@ -8,6 +8,10 @@
       {
         "link": "/",
         "label": "Digital Marketplace"
+      },
+      {
+        "link": url_for('.become_a_supplier'),
+        "label": "Become a supplier"
       }
     ]
   %}
--- a/app/templates/suppliers/duns_number.html
+++ b/app/templates/suppliers/duns_number.html
@@ -8,6 +8,14 @@
     {
       "link": "/",
       "label": "Digital Marketplace"
+    },
+    {
+      "link": url_for('.become_a_supplier'),
+      "label": "Become a supplier"
+    },
+    {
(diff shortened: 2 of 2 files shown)


laurenceberry:
Something looks off with the spacing for the <ul> and <li>, it should all look like one paragraph. I think if you wrap it within the <p> above it should look OK.

laurenceberry:
Something looks off with the spacing for the ul and li, it should all look like one paragraph. I think if you wrap it within the p tag above it should look OK.

samuelhwilliams:
Nope, doesn't work.

See this is what I'm getting at. I mean I know it looks off, but I don't know what pattern we have to fix this. If we need something custom I can add that, but we really shouldn't have to.

laurenceberry:
No we shouldn't have to create new patterns. This is how all bulleted lists are displayed on the service so it seems strange it breaks here, perhaps related to .dmspeak?

An example of where we use a bulleted list: /suppliers/frameworks/digital-outcomes-and-specialists-3/declaration/reuse

samuelhwilliams:
Updated.

laurenceberry:
Looks good to me now

PR #787, opened 2018-02-09
Add page to edit organisation size

samuelhwilliams:
## Summary
Adds a new page to the supplier account that allows a supplier to edit their organisation size. This page hasn't been plumbed up to anything yet - that will be done in a later ticket.

There's no flash message confirming the action the user has taken included as part of this ticket. It feels like it might be good to add this as part of this ticket but I'll need the content (@constancecerf?)

## Examples
screen shot 2018-02-09 at 14 12 53
screen shot 2018-02-09 at 14 13 32

## Ticket
https://trello.com/c/eB9u3xe7/40-new-editable-page-build-organisation-size-page

### DIFF
--- a/app/main/forms/suppliers.py
+++ b/app/main/forms/suppliers.py
@@ -1,6 +1,6 @@
 from flask_wtf import Form
-from wtforms import IntegerField
-from wtforms.validators import DataRequired, ValidationError, Length, Optional, Regexp
+from wtforms import IntegerField, RadioField
+from wtforms.validators import DataRequired, ValidationError, Length, Optional, Regexp, InputRequired
 
 from dmutils.forms import StripWhitespaceStringField, EmailField, EmailValidator
 
@@ -82,3 +82,36 @@ class EmailAddressForm(Form):
         DataRequired(message="You must provide an email address."),
         EmailValidator(message="You must provide a valid email address."),
     ])
+
+
+class CompanyOrganisationSizeForm(Form):
+    OPTIONS = [
+        {
+            "value": "micro",
+            "label": "Micro",
+            "description": "Under 10 employees and 2 million euros or less in either annual turnover or balance "
+                           "sheet total",
+        },
+        {
+            "value": "small",
+            "label": "Small",
+            "description": "Under 50 employees and 10 million euros or less in either annual turnover or balance "
+                           "sheet total",
+        },
+        {
+            "value": "medium",
+            "label": "Medium",
+            "description": "Under 250 employees and either 50 million euros or less in either annual turnover or "
+                           "43 million euros or less in annual balance sheet total",
+        },
+        {
+            "value": "large",
+            "label": "Large",
+            "description": "250 or more employees and either over 50 million euros in annual turnover or over 43 "
+                           "million euros in balance sheet total",
+        },
+    ]
+
+    organisation_size = RadioField('Organisation size',
+                                   validators=[InputRequired(message="You must choose an organisation size.")],
+                                   choices=[(o['value'], o['label']) for o in OPTIONS])
--- a/app/templates/suppliers/edit_supplier_organisation_size.html
+++ b/app/templates/suppliers/edit_supplier_organisation_size.html
@@ -24,12 +24,12 @@
 {% block main_content %}
 
 <div class="single-question-page">
-  {% with heading = "What size is your organisation?",
-          smaller = True %}
-    {% include "toolkit/page-heading.html" %}
-  {% endwith %}
   <div class="grid-row">
     <div class="column-two-thirds">
+      {% with heading = "What size is your organisation?", smaller = True %}
+        {% include "toolkit/page-heading.html" %}
+      {% endwith %}
+      
       <form method="POST" action="{{ url_for('.edit_supplier_organisation_size') }}">
         <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
         {% set hint %}
--- a/app/templates/suppliers/edit_supplier_organisation_size.html
+++ b/app/templates/suppliers/edit_supplier_organisation_size.html
@@ -1,0 +1,61 @@
+{% extends "_base_page.html" %}
+
+{% block page_title %}Organisation size – Digital Marketplace{% endblock %}
+
+{% block breadcrumb %}
+  {% with items = [
+    {
+      "link": "/",
+      "label": "Digital Marketplace"
+    },
+    {
+       "link": url_for('.dashboard'),
+       "label": "Your account"
+    },
+    {
+      "link": url_for('.supplier_details'),
+      "label": "Company details"
+    },
+  ] %}
+    {% include "toolkit/breadcrumb.html" %}
+  {% endwith %}
+{% endblock %}
+
+{% block main_content %}
+
+<div class="single-question-page">
+  {% with heading = "What size is your organisation?",
+          smaller = True %}
+    {% include "toolkit/page-heading.html" %}
+  {% endwith %}
+  <div class="grid-row">
+    <div class="column-two-thirds">
+      <form method="POST" action="{{ url_for('.edit_supplier_organisation_size') }}">
+        <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
+        {% set hint %}
+          This information will be used to report on the number of contracts that go to small and medium sized
+          enterprises (SMEs).
+        {% endset %}
+        
+        {% with question = "What size is your organisation?",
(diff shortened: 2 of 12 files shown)


TheDoubleK:
I don't think we have flash message confirmation - you just go back to the overview page and can see the updated value there.

app/templates/suppliers/edit_supplier_organisation_size.html
```diff
@@ -0,0 +1,61 @@
...
+      "link": "/",
+      "label": "Digital Marketplace"
+    },
+    {
+       "link": url_for('.dashboard'),
+       "label": "Your account"
+    },
+    {
+      "link": url_for('.supplier_details'),
+      "label": "Company details"
```
TheDoubleK: Did this breadcrumb text come from the prototype?  In the "registered address" branch we were going with "Supplier details" for the breadcrumb text, as that's the heading on the page (I think). 

But that is going to change with an upcoming ticket.

Either way we should agree what the breadcrumb text is and be consistent.

app/templates/suppliers/edit_supplier_organisation_size.html
```diff
@@ -0,0 +1,61 @@
...
+    },
+  ] %}
+    {% include "toolkit/breadcrumb.html" %}
+  {% endwith %}
+{% endblock %}
+
+{% block main_content %}
+
+<div class="single-question-page">
+  {% with heading = "What size is your organisation?",
```
katstevens: Should the heading go inside the `column-two-thirds` div? That's a genuine question - a similar template on the Briefs FE has the div classes nested as `grid-row > column-two-thirds > single-question-page`, but I have no idea what the right order is meant to be.

tests/app/main/test_suppliers.py
```diff
@@ -1982,3 +1982,63 @@ def test_one_open_one_closed_framework(self, data_api_client):
...
+            doc = html.fromstring(res.get_data(as_text=True))
+            error = doc.xpath('//span[@id="error-organisation_size"]')
+
+            assert len(error) == 1, \
+                'Only one validation message should be shown.'
+
+            assert error[0].text.strip() == 'You must choose an organisation size.', \
+                'The validation message is not as anticipated.'
+
+    @pytest.mark.parametrize('size', (None, 'micro', 'small', 'medium', 'large'))
```
TheDoubleK: We could test that an unexpected value posted here doesn't call the API, e.g. `ginormous` or something shouldn't be saved (I don't think this is enforced at the API level, so we will to enforce it at the frontend POST route here).

app/main/forms/suppliers.py
```diff
@@ -82,3 +82,36 @@ class EmailAddressForm(Form):
...
+        {
+            "value": "large",
+            "label": "Large",
+            "description": "250 or more employees and either over 50 million euros in annual turnover or over 43 "
+                           "million euros in balance sheet total",
+        },
+    ]
+
+    organisation_size = RadioField('Organisation size',
+                                   validators=[InputRequired(message="You must choose an organisation size.")],
```
katstevens: What's the difference between `DataRequired` (which we already import) and `InputRequired`?

tests/app/main/test_suppliers.py
TheDoubleK: I'm going to add an enum to the API schema to enforce this.  But still, would be nice to not make the call in the first place if someone posts something stupid.

app/templates/suppliers/edit_supplier_organisation_size.html
TheDoubleK: I'm pretty sure the heading should be inside the two-thirds.
Double check with @laurenceberry maybe?

tests/app/main/test_suppliers.py
TheDoubleK: OK so the values are enforced by the API model, so API will error for a bad value, so we probably don't need to validate here after all.

samuelhwilliams: Yeah sorry, it is enforced by the API so I didn't bother. I could add a comment if you think that would prevent any confusion in the future?

app/templates/suppliers/edit_supplier_organisation_size.html
samuelhwilliams: This page's HTML was kindly donated by our existing template at https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/templates/suppliers/company_name.html, so if it's broken, I'll need to fix it there as well.

samuelhwilliams: But going through all the templates in the suppliers folder, `become_a_supplier.html`, `companies_house_number.html`, `company_contact_details.html`, `company_name.html`, (etc I give up now) are all the same.

samuelhwilliams: So frigging inconsistent. :sadface:

samuelhwilliams: Maybe this single question page is a template that should go out to utils and be generalised in some fashion. 🤷‍♂️

app/templates/suppliers/edit_supplier_organisation_size.html
samuelhwilliams: `Company details` is indeed from the prototype.

app/templates/suppliers/edit_supplier_organisation_size.html
TheDoubleK: The two-thirds column only makes any difference if the heading is long enough to get wrapped by it, so I have a feeling that only long headings ever got moved inside.

samuelhwilliams: Fair enough.

samuelhwilliams: Although maybe we don't want longer headings to be wrapped at the two-thirds size because it doesn't cause such an accessibility issue with wider headings given their increased font size.

TheDoubleK: I'm pretty sure they make Laurence and Ralph sad though.

samuelhwilliams: Happy to put it inside. Can you give some input please @laurenceberry?

app/main/forms/suppliers.py
samuelhwilliams: Essentially, `InputRequired` allows us to specify a custom message and `DataRequired` doesn't. WTForms also recommends using InputRequired generally.

After digging into what's actually going on under the hood this morning, it looks like DataRequired has some kind of incompatibility with SelectField. InputRequired makes sure that the user has submitted some kind of value; DataRequired makes sure the user has submitted a value that evaluates to 'truthy' after being coerced to whatever format the form wants.

There's an existing issue on the project that I've tacked my little note onto: https://github.com/wtforms/wtforms/issues/289

app/templates/suppliers/edit_supplier_organisation_size.html
katstevens: Sorry for opening a can of worms 😸

app/main/forms/suppliers.py
katstevens: That's strange. We have a custom message on a `DataRequired` validator on the Briefs FE: https://github.com/alphagov/digitalmarketplace-briefs-frontend/blob/master/app/main/forms/awards.py#L11. However if `InputRequired` is recommended then by all means we should use that instead (we use both on the Briefs FE!).

samuelhwilliams: I suspect that works because you're coercing to an int rather than a string. int(None) throws an error, which I guess gets caught and handled. str(None) ends up with 'None' that then fails  a later check. Sounds pretty hinky all-in-all 🤔 

But yeah, it sounds like InputRequired is what we should be using instead.

samuelhwilliams:
@TheDoubleK updated and moved headings as requested. Is this good now?

app/templates/suppliers/edit_supplier_organisation_size.html
CrystalPea: @constancecerf any input? :)

app/main/forms/suppliers.py
CrystalPea: @Wynndow do you think we should maybe change our validators on address and country forms?

Wynndow: Yep, it's probably a good idea.

TheDoubleK:
Good to go after a rebase.

PR #788, opened 2018-02-09
G Cloud 9 Contract variation updates

DilwoarH:
As a supplier, I can review the proposed contract variation for the GCloud9 framework agreement and call-off contract, so that I understand what it includes and decide whether to accept it.

Ticket: https://trello.com/c/ipSUk2Yo/104-add-final-copy-for-contract-variation-for-gcloud-9-and-update-feature-to-reflect-the-correct-dates

screenshot-localhost-2018 02 09-16-39-24

### DIFF
--- a/app/assets/scss/_variables.scss
+++ b/app/assets/scss/_variables.scss
@@ -1,2 +1,2 @@
-$diff-insertion-color: #acf2bd;
-$diff-deletion-color: #fdb8c0;
+$diff-insertion-green: #acf2bd;
+$diff-deletion-red: #fdb8c0;
--- a/app/assets/scss/_variables.scss
+++ b/app/assets/scss/_variables.scss
@@ -1,0 +1,2 @@
+$diff-insertion-color: #acf2bd;
+$diff-deletion-color: #fdb8c0;
--- a/app/assets/scss/application.scss
+++ b/app/assets/scss/application.scss
@@ -104,11 +104,11 @@ $path: "/suppliers/static/images/";
 // app-specific.
 ins {
   text-decoration: none;
-  background-color: $diff-insertion-color;
+  background-color: $diff-insertion-green;
   padding-right: 3px;
 }
 
 del {
-  background-color: $diff-deletion-color;
+  background-color: $diff-deletion-red;
   padding-right: 3px;
 }
--- a/app/assets/scss/application.scss
+++ b/app/assets/scss/application.scss
@@ -53,6 +53,7 @@ $path: "/suppliers/static/images/";
 
 
 // App-specific styles
+@import "_variables.scss";
 @import "_assurance-question.scss";           // This could probably be removed as the only usage is in service editing
                                               // and editing expired services is not officially supported (it can be
                                               // done only by hacking URLs). However, without also removing supporting
@@ -103,11 +104,11 @@ $path: "/suppliers/static/images/";
 // app-specific.
 ins {
   text-decoration: none;
-  background-color: #cce0d6;
-  padding: 0 3px;
+  background-color: $diff-insertion-color;
+  padding-right: 3px;
 }
 
 del {
-  background-color: #efcfd1;
-  padding: 0 3px;
+  background-color: $diff-deletion-color;
+  padding-right: 3px;
 }
--- a/app/templates/frameworks/contract_variation.html
+++ b/app/templates/frameworks/contract_variation.html
@@ -47,28 +47,36 @@
     {% endfor %}
   {% endwith %}
 
-
+    <div class="grid-row">
+      <div class="column-one-whole">
+        {% if form.errors %}
+          {%
+            with
+            errors = form_errors
+          %}
+            {% include 'toolkit/forms/validation.html' %}
+          {% endwith %}
+        {% endif %}
+      </div>
+    </div>
     <div class="grid-row">
         <div class="column-two-thirds">
-            {% with
-              heading = "{}:{}contract variation".format(framework.name, " " if variation_details.get('countersignedAt') else " proposed "),
-              smaller = True
-            %}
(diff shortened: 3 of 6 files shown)


app/templates/frameworks/contract_variation.html
```diff
@@ -69,6 +69,8 @@
           <div class="section-description">
           {% if variation_details.get('countersignedAt') and agreed_details %}
             {{ variation.variation_description_in_place }}
+          {% elif not variation_details.get('countersignedAt') and agreed_details and variation.variation_description_not_in_place_but_agreed %}
```
TheDoubleK: The logic here is reasonably complex so it would be nice to have some additional tests around the expected text in different cases.

DilwoarH: @TheDoubleK added in new test:
https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/788/files#diff-860e899ae62cbab98fee8fa871bcc53bR4994

tests/app/main/test_frameworks.py
```diff
@@ -4980,8 +4988,35 @@ def test_shows_signer_details_and_no_form_if_already_agreed(self, data_api_clien
...
+        data_api_client.get_supplier_framework_info.return_value = already_agreed
+
+        res = self.client.get("/suppliers/frameworks/g-cloud-9/contract-variation/1")
+        page_text = res.get_data(as_text=True)
+        doc = html.fromstring(page_text)
+
+        assert res.status_code == 200
+        assert len(doc.xpath('//h1[contains(text(), "G-Cloud 9: proposed contract variation")]')) == 1
+        assert len(doc.xpath('//h2[contains(text(), "Contract variation status")]')) == 1
+        assert u"<span>William Drăyton<br />agreed@email.com<br />Friday 19 August 2016 at 4:47pm</span>" in page_text
```
TheDoubleK: We shouldn't need the `u` in front of strings anywhere since we are Python 3 and everything is unicode by default.  Would be good to remove these as we come across them, rather than adding more.

DilwoarH:
@TheDoubleK - replaced all of the `u"` in that file

TheDoubleK:
Nice one, yep merge once tests pass.

app/templates/frameworks/contract_variation.html
```diff
@@ -77,6 +79,35 @@
...
+          <div class="section-summary">
+            {{ summary.heading("Contract variation status", id="contract_variation_status") }}
+            {% call(item) summary.list_table(
+              [
+                {"key": "Agreed by", "value": ("%s<br />%s<br />%s" % (agreed_details.agreedUserName, agreed_details.agreedUserEmail, agreed_details.agreedAt|datetimeformat))|safe},
+                {"key": "Countersigned by", "value": ("%s<br />%s<br />%s" % 
+                  (
+                    variation_details.countersignerName, 
+                    variation_details.countersignerRole, 
+                    agreed_details.agreedAt|datetimeformat if variation_details.get('countersignedSameAsAgreed') else variation_details.countersignedAt|dateformat
```
TheDoubleK: I don't think we need `countersignedSameAsAgreed` anywhere at all now, as it's not legally valid.

DilwoarH: @TheDoubleK removed

app/assets/scss/application.scss
```diff
@@ -103,11 +103,11 @@ $path: "/suppliers/static/images/";
 // app-specific.
 ins {
   text-decoration: none;
-  background-color: #cce0d6;
+  background-color: #acf2bd;
```
samuelhwilliams: We shouldn't use custom colour codes here, probably. Can we using existing colours imported by one of our toolkits?

If we can't, we should at least define these as sass variables so that we can give them a semantic name.

app/assets/scss/application.scss
```diff
@@ -103,11 +103,11 @@ $path: "/suppliers/static/images/";
...
   padding-right: 3px;
 }
 
 del {
-  background-color: #efcfd1;
+  background-color: #fdb8c0;
```
samuelhwilliams: Same here.

app/assets/scss/application.scss
DilwoarH: We can't use any existing colours because they do not follow any existing pattern as far as I am aware. I have created a variables sass file and also created a trello ticket to get this moved: https://trello.com/c/a90iPGOH/117-move-ins-del-to-frontend-toolkit.

app/assets/scss/_variables.scss
```diff
@@ -0,0 +1,2 @@
+$diff-insertion-color: #acf2bd;
```
idavidmcdonald: A small part of me wants these in here - https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/toolkit/scss/_digital-marketplace-colours.scss

It feels like 'diff' colours are something we have used in more than one place so could standardise them there.

Although I'm not too sure if this suggestion is actually too much work for the benefit. I'll let you make the call @DilwoarH

app/assets/scss/_variables.scss
```diff
@@ -0,0 +1,2 @@
+$diff-insertion-green: #acf2bd;
```
idavidmcdonald: A small part of me wants these in here - https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/toolkit/scss/_digital-marketplace-colours.scss

It feels like 'diff' colours are something we have used in more than one place so could standardise them there.

Although I'm not too sure if this suggestion is actually too much work for the benefit. I'll let you make the call @DilwoarH

Also, I think `diff-insertion-color` is more semantic than `diff-insertion-green` - should the semantic name need to know anything about it maybe being green or red or whatever colour we so wish?

PR #789, opened 2018-02-12
update framework to 10.4.0

DilwoarH:
Required PR:
https://github.com/alphagov/digitalmarketplace-frameworks/pull/494

Ticket: https://github.com/alphagov/digitalmarketplace-frameworks/pull/494


TheDoubleK:
I think the process is to update the frameworks dependency in `package.json` and then run `yarn install` which will generate any required changes to `yarn.lock`.  We shouldn't be changing `yarn.lock` manually, as any changes will be overwritten the next time someone runs `yarn install`.

TheDoubleK:
So this is the line you need to edit, before running yarn: https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/package.json#L24

DilwoarH:
@TheDoubleK - updated

PR #790, opened 2018-02-12
Build a page for suppliers to update address and country details

CrystalPea:
## Trello ticket: 
https://trello.com/c/FeBysBI7/37-new-editable-page-build-company-address-page-inc-country-picker

## Work done:
- Integrate country picker into Digital Marketplace set-up
 - Make `edit_registered_address` page with country picker
- Make sure both forms (for address and for country) run validation
- Prepopulate registrationCountry field if user has provided a country
- Add top of the page errors
- Add default empty option to country picker if user has not provided a country before
- Apply validation styling manually for country picker field
- Validate registration country against country-picker list of countries
- Load list of countries in helpers (instead of in the views)

For more detail see individual commit messages.

@Wynndow, @CrystalPea and @TheDoubleK worked on developing this feature.

### How the page looks like when it's happy:
screen shot 2018-02-12 at 16 43 11

_____

### How the page looks when user tries to submit empty country field:
screen shot 2018-02-13 at 14 44 21

______

### How the page looks for users who have javascript turned-off (This is a fallback option, for example when they use a screen reader):
screen shot 2018-02-13 at 11 26 22


app/main/forms/suppliers.py
```diff
@@ -34,11 +34,37 @@ class EditContactInformationForm(Form):
...
+
+class EditRegisteredCountryForm(Form):
+    def __init__(self, valid_countries=None, **kwargs):
+        super().__init__(**kwargs)
+        if valid_countries:
+            self.registrationCountry.validators[1].values = [country[1] for country in valid_countries]
+
+    registrationCountry = StripWhitespaceStringField('Country', validators=[
+        DataRequired(message="You need to enter the country."),
+        AnyOf(values=[], message="This is not a valid country"),
```
CrystalPea: Hi @constancecerf,

We identified an edge case where user tries to manipulate html to enter an invalid country to the database. We built a validation safeguarding against that unlikely but possible scenario. And this validation needs a message informing the user what they did wrong. 

For now we used: "This is not a valid country". Let me know if this is alright or if we need to change it :).

app/templates/suppliers/registered_address.html
```diff
@@ -0,0 +1,148 @@
...
+    url: '{{ asset_path }}location-autocomplete-graph.json'
+  });
+
+  $('#location-autocomplete').keyup(function(event) {
+    // Clear the input value of the country select input, if the autocomplete is cleared.
+    if ($(this).val().length < 2) {
+      $('#location-autocomplete-select').val('');
+    };
+  });
+</script>
```
CrystalPea: Hi @DilwoarH. We needed a couple scripts to make our feature act in an expected manner, but we are not sure if they can stay in the template or should they be moved out and if it's the latter, then how to do it. Any advice?

laurenceberry:
Looking good.

One thing I've noticed is the error banner is below the H1, usually the error banner sits above the H1 on the page.

CrystalPea:
@laurenceberry I think we copied the pattern for the error banner location from "Edit company info" page:

screen shot 2018-02-12 at 16 48 19

CrystalPea:
@laurenceberry ok, I will be changing it for the address page now then ☺️ . Thanks for the confirmation :3

app/main/helpers/__init__.py
```diff
@@ -2,6 +2,8 @@
 from datetime import datetime
 from functools import wraps
 import hashlib
+import os
```
idavidmcdonald: Do we need these imports in here when the rest of the file hasn't changed?

app/main/helpers/suppliers.py
```diff
@@ -0,0 +1,24 @@
...
+
+def load_countries():
+    helpers_path = os.path.abspath(os.path.dirname(__file__))
+    countryfile = os.path.join(helpers_path, '../../static/location-autocomplete-canonical-list.json')
+    with open(countryfile) as f:
+        return json.load(f)
+
+
+def get_country_name_from_country_code(country_code):
+    if country_code and country_code != 'gb':
```
idavidmcdonald: I can't quite articulate why but I want to suggest swapping the order of the logic statements so the special case comes first e.g.
```
if country_code and country_code == 'gb':
elif country_code:
else:
```

I think I like the special case being first and then more normal behaviour continuing afterwards but I'm not too fussed eitherway

app/main/helpers/suppliers.py
```diff
@@ -0,0 +1,24 @@
...
+        potential_names = [country[0] for country in COUNTRY_LIST if country[1] == country_code]
+        return potential_names[0] if potential_names else ''
+    elif country_code:
+        # We need to support the old code style ('gb') until after the existing data has updated.
```
idavidmcdonald: after the existing data has updated?

What existing data? What process will update it? Is this something that will happen naturally or are running a migration later?

app/main/helpers/suppliers.py
```diff
@@ -0,0 +1,24 @@
...
+        return 'United Kingdom'
+    else:
+        # In the case that a suppliers registration country isn't set we maintain exiting behavior.
```
idavidmcdonald: I assume this is `existing`? Also existing behaviour based on what - where is this already being done - us on the marketplace or a country picker existing behaviour thing?

app/main/views/suppliers.py
```diff
@@ -85,13 +86,84 @@ def supplier_details():
     except APIError as e:
         abort(e.status_code)
     supplier['contact'] = supplier['contactInformation'][0]
+    country_name = get_country_name_from_country_code(supplier['registrationCountry'])
```
idavidmcdonald: Is there a reason we aren't doing something like `supplier['registrationCountry']` which might be more consistent (if that is useful)?

app/main/forms/suppliers.py
```diff
@@ -28,20 +29,41 @@ class EditSupplierForm(Form):
...
+    ])
+    city = StripWhitespaceStringField('Town or city', validators=[
+        InputRequired(message="You need to enter the town or city."),
+    ])
+    postcode = StripWhitespaceStringField('Postcode', validators=[
+        InputRequired(message="You need to enter the postcode."),
+    ])
+
+
+class EditRegisteredCountryForm(Form):
```
idavidmcdonald: Why are these separate forms?

app/templates/suppliers/registered_address.html
```diff
@@ -0,0 +1,147 @@
...
+      {% with
+        heading = "What is your registered office address?",
+        smaller = true
+      %}
+        {% include 'toolkit/page-heading.html' %}
+      {% endwith %}
+    </div>
+  </div>
+
+  <form action="{{ url_for('.update_registered_address') }}" method="post" enctype="multipart/form-data">
```
idavidmcdonald: Do we need this enctype? I think we only need it if there is a file upload which I don't think there is for this form?

https://stackoverflow.com/questions/4526273/what-does-enctype-multipart-form-data-mean

gulpfile.js
```diff
@@ -154,6 +155,20 @@ function copyFactory(resourceName, sourceFolder, targetFolder) {
 
 }
 
+function copyFiletypeFactory(resourceName, sourceFolder, sourceExtension, targetFolder) {
```
idavidmcdonald: `sourceFileExtension` would be a little bit more readable

gulpfile.js
```diff
@@ -154,6 +155,20 @@ function copyFactory(resourceName, sourceFolder, targetFolder) {
...
+
+  return function() {
+    return gulp
+      .src(sourceFolder + "/**/" + "*." + sourceExtension, { base: sourceFolder })
```
idavidmcdonald: Can we join the `"/**/" + "*."`?

app/main/helpers/suppliers.py
CrystalPea: What you are referring to is called a guard statement I think and I agree that it's a good practice :)

http://www.thechrisoshow.com/2009/02/16/using-guard-clauses-in-your-ruby-code/

app/main/helpers/__init__.py
Wynndow: Nope, they were leftovers after a refactor, I'll whip them out. I'm surprised flake-8 did't catch them though.

app/main/helpers/suppliers.py
Wynndow: Yeah, I see what you mean. Something about checking the negative case rather than the positive feels a bit odd.

app/main/helpers/suppliers.py
Wynndow: This is referring to the existing country code data be have for UK suppliers. It'll get updated by a migration. I'll make the comment more explicit.

app/main/helpers/suppliers.py
Wynndow: Yep existing. Currently any supplier that isn't UK based just has an empty string in the database in the registration country field.

app/main/views/suppliers.py
Wynndow: Do you mean something like `country_name = supplier['registrationCountry']`?
If so, it's because the data we get from the country picker is a country code, and that's what we're going to store in the db. Makes it easy to replay the data into the country picker form. We have the helper so we can convert the country codes to human readable names.

app/main/forms/suppliers.py
Wynndow: It's because we're making two separate calls to the apiclient to update the supplier table, as well as update the contact information table. We _could_ do it with one form, but it would involve picking out all the data individually from the form, rather than just passing all the data in as one argument. It was decided it was neater to use separate forms.

app/templates/suppliers/registered_address.html
Wynndow: Yeah, I think you're right. That link is an interesting read too!

app/assets/scss/application.scss
```diff
@@ -111,3 +111,18 @@ del {
   background-color: #efcfd1;
   padding: 0 3px;
 }
+
+
+// These styles are to customise the country picker widget supplied by registers. The country picker's styles are
+// included in a separate stylesheet which is brought in by the country picker npm package. This makes is difficult
+// to update the source, so we do it here. It's important that the country picker's style sheet is loaded in the
+// template before our own CSS, so the styles defined here override and take precedence.
+#location-autocomplete, .autocomplete__hint {
```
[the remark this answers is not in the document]
TheDoubleK: We thought about whether we'd ever have another use for the country picker ever again and came to the conclusion it's extremely unlikely (in that nobody could thing of any possible scenario where we might want it).  That doesn't mean the styling shouldn't be in the toolkit though, but personally I don't mind it here.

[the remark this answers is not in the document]
idavidmcdonald: I don't mind either way at the moment I think. It doesn't feel that likely that we would put this in other apps at the moment. If this was was duplicated then I think yes it should go in the toolkit but as it's a one off I think it's also fine in here.

So yeah no definitive opinion from me I'm afraid

app/main/helpers/suppliers.py
samuelhwilliams: Think he means the typo, you currently have `exiting`

samuelhwilliams: But also might be need to detail what 'existing behaviour' is a little bit (and maybe a datestamp for quick point of reference)

app/assets/scss/application.scss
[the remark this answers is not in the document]
Wynndow: Yeah, my opinion was the same as Davids. I think if we ever need to include the country picker somewhere else in our apps then at that point we can do the work to move it to the toolkit. For now I think it's fine here.

app/main/forms/suppliers.py
```diff
@@ -28,20 +29,41 @@ class EditSupplierForm(Form):
...
     contactName = StripWhitespaceStringField('Contact name', validators=[
-        DataRequired(message="You must provide a contact name"),
+        InputRequired(message="You must provide a contact name"),
     ])
     email = EmailField('Contact email address', validators=[
-        DataRequired(message="You must provide an email address"),
+        InputRequired(message="You must provide an email address"),
     ])
     phoneNumber = StripWhitespaceStringField('Contact phone number')
+    # TODO: remove these fields when address is removed from the supplier details edit page
```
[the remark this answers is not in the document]
Wynndow: I think it's covered here: https://trello.com/c/A3ovgHlk

app/main/views/suppliers.py
idavidmcdonald: No sorry, I meant
`supplier['registrationCountry'] = get_country_name_from_country_code(supplier['registrationCountry'])`
or something like
`supplier['registrationCountryName'] = get_country_name_from_country_code(supplier['registrationCountry'])`

app/assets/scss/application.scss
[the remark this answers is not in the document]
samuelhwilliams: Fair points.

I guess my concern is just that, when looking for existing styles, people aren't likely to scour every app to find if we have anything already implement - they'll check maybe the app they're in and the frontend-toolkit. If they don't find anything there, there's a risk they'll re-implement styles in their own app and now we have two fractured bits of CSS. If we have the default approach of adding styles to the toolkit, this risk is minimised.

This is unlikely to happen in this instance (ie with the country picker), but with other components I could imagine it happening - and I'm pretty sure we saw evidence of it having happened before on our last mission. Having a shared starting position of implementing styles in the toolkit would help to minimise the risk overall and the effort/cognitive load of remembering to check every app every time you're adding what might be a new style.

app/main/views/suppliers.py
```diff
@@ -85,13 +86,84 @@ def supplier_details():
...
     return render_template(
         "suppliers/details.html",
         supplier=supplier,
+        country_name=country_name,
     ), 200
 
 
+@main.route('/registered-address/edit', methods=['GET'])
+@login_required
+def edit_registered_address(registered_address_form=None, registered_country_form=None, error=None):
```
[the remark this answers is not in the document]
Wynndow: Yeah, I spent a bit of time at the end of yesterday getting this working and it ends up being really quite messy. Rather than populating the data in the forms when it's instantiated by passing in the data as kwargs with **supplier (or whatever), you need to assign each one individually.
I think it makes sense on views with forms without many fields, but once you have a few it gets a bit cumbersome.

app/main/forms/suppliers.py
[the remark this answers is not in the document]
TheDoubleK: Yep: https://trello.com/c/A3ovgHlk/51-remove-address-from-edit-supplier-details-page

app/templates/suppliers/registered_address.html
```diff
@@ -0,0 +1,147 @@
...
+  %}
+    {% include "toolkit/breadcrumb.html" %}
+  {% endwith %}
+{% endblock %}
+
+
+{% block main_content %}
+  {% if error %}
+  <div class="validation-masthead" aria-labelledby="validation-masthead-heading">
+    <h3 class="validation-masthead-heading" id="validation-masthead-heading">
```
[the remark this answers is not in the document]
idavidmcdonald: I think that would give you two h1's though (toolkit/page-heading.html already will be on the page). I _think_ that is to be avoided although there seems to be some inconsistent information on the web about whether it should or not...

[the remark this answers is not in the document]
samuelhwilliams: It does, but that seems to be accepted practice. Certainly on DM we do it a lot. Might be worth getting confirmation/input from @laurenceberry.

app/main/forms/suppliers.py
```diff
@@ -28,20 +29,41 @@ class EditSupplierForm(Form):
...
     address1 = StripWhitespaceStringField('Building and street')
     city = StripWhitespaceStringField('Town or city')
     postcode = StripWhitespaceStringField('Postcode')
 
 
+class EditRegisteredAddressForm(Form):
+    id = IntegerField()
+    address1 = StripWhitespaceStringField('Building and street', validators=[
+        InputRequired(message="You need to enter the street address."),
+    ])
```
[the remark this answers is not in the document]
TheDoubleK: Yep, probably for all string input fields in our WTForms.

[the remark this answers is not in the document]
TheDoubleK: 255?  For historical reasons?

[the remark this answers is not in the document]
Wynndow: Works for me. We could be more restrictive around postcodes. I'm inclined to think it doesn't matter though, it's just to stop someone entering a silly amount of text.

app/main/helpers/suppliers.py
```diff
@@ -0,0 +1,24 @@
...
+        return ''
+
+
+COUNTRY_LIST = load_countries()
```
TheDoubleK: Should this be `COUNTRY_TUPLE` if we do Sam's suggestion? Or `COUNTRY_DATA`?

app/main/views/suppliers.py
```diff
@@ -85,13 +86,84 @@ def supplier_details():
...
+    try:
+        data_api_client.update_supplier(
+            current_user.supplier_id,
+            registered_country_form.data,
+            current_user.email_address,
+        )
+
+        data_api_client.update_contact_information(
+            current_user.supplier_id,
+            registered_address_form.id.data,
```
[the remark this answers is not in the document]
TheDoubleK: Interesting point.

If the GET and POST routes were merged together as you suggest then the Supplier is "got" as the first thing there.  The additional API fetch before posting doesn't feel like a problem, and I'm sure we have that in some other places.

I think you're right it would be nice to avoid leaking these IDs.

app/templates/suppliers/details.html
```diff
@@ -100,11 +100,11 @@
   {% if supplier.registrationCountry and not supplier.companiesHouseNumber %}
     {% call summary.row() %}
       {{ summary.field_name('Where your business was established') }}
-      {{ summary.text(supplier.registrationCountry|upper) }}
+      {{ summary.text(country_name) }}
```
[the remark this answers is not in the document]
TheDoubleK: ALL SHOUTY CAPS was good for two-letter country codes like GB and UK.

app/templates/suppliers/details.html
```diff
@@ -100,11 +100,11 @@
...
     {% endcall %}
   {% endif %}
 
-  {% if supplier.otherCompanyRegistrationNumber and not supplier.companiesHouseNumber and supplier.registrationCountry != 'gb' %}
+  {% if supplier.otherCompanyRegistrationNumber and not supplier.companiesHouseNumber and country_name != 'United Kingdom' %}
```
TheDoubleK: Minor thing but it feels like it would be nice to have the logic based on the value in supplier rather than on the derived country name if possible - something like `supplier.registrationCountry != 'country:gb'?

Ah, but then that won't cover the "legacy" case of just `gb` will it?

OK probably fine like this then.

app/templates/suppliers/details.html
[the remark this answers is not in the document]
samuelhwilliams: Ah I thought that still displayed the full string. Fair enough. 👍

gulpfile.js
TheDoubleK: We already have `copyFileFactory` that copies all files, and `copyFiletypeFactory` seems a logical extension.  Could be `copyFilesWithExtensionFactory`? But `sourceFileExtension` doesn't mean a lot to me.

idavidmcdonald: that is for the argument name rather than the function name (if that wasn't clear)

app/main/helpers/suppliers.py
```diff
@@ -0,0 +1,24 @@
+import os
+import json
+
+
+def load_countries():
+    helpers_path = os.path.abspath(os.path.dirname(__file__))
+    countryfile = os.path.join(helpers_path, '../../static/location-autocomplete-canonical-list.json')
+    with open(countryfile) as f:
+        return json.load(f)
```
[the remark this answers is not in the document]
Wynndow: Will do. Don't understand the bit about semantically expecting it though?

app/main/views/suppliers.py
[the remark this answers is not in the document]
Wynndow: OK, i did the thing.

app/main/views/suppliers.py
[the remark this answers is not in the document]
Wynndow: OK, I've combined the get and post now, and the apiclient call to update contact info is using the id from the supplier. I've also dropped the hidden field in the form in the template which is nice.

app/templates/suppliers/registered_address.html
```diff
@@ -0,0 +1,147 @@
+{% extends "_base_page.html" %}
+{% import "macros/forms.html" as forms %}
+{% import "toolkit/summary-table.html" as summary %}
+{% block head %}
+  <link rel="stylesheet" href="{{ asset_path }}stylesheets/location-autocomplete.min.css" />
```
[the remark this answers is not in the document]
Wynndow: Calling `super` on the line below this one means the head block from the base page is included, and so `asset_fingerprinter.get_url` is called. Is that what you mean?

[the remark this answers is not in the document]
Wynndow: I've updated. The asset fingerprinter looks like it created a unique hash for the asset, which prevents it from being cached if it changes. It appends the hash to the end of the url as a parameter like so: `<link type="text/css" rel="stylesheet" media="screen" href="/static/stylesheets/application.css?6d455fbc43f38555db44c03b76c325aa" />`. I guess it forces the browser to reload it if it changes.

app/templates/suppliers/registered_address.html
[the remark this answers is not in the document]
Wynndow: Visually they're the same. I'll check with laurence.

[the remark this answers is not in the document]
Wynndow: H1 has it.

app/templates/suppliers/registered_address.html
```diff
@@ -0,0 +1,147 @@
...
+            {% endfor %}
+          {% endfor %}
+          </ul>
+          {% endif %}
+
+          {% if registered_country_form.errors %}
+          <ul>
+          {% for field_name, field_errors in registered_country_form.errors|dictsort %}
+            {% for error in field_errors %}
+            <li><a href="javascript:document.getElementById('location-autocomplete').focus()" class="validation-masthead-link">{{ registered_country_form[field_name].label.text }}</a></li>
```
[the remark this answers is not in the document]
Wynndow: Yep. really odd. I've already spent a bit of time thinking about it, so don't really want to spend a lot more...

app/templates/suppliers/registered_address.html
```diff
@@ -0,0 +1,147 @@
...
+
+  <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
+
+  <div class="grid-row">
+    <div class="column-two-thirds">
+
+      {{ forms.input('id', registered_address_form.id.data, type='hidden') }}
+      {{ forms.question_input('address1', 'Building and street', value=registered_address_form.address1.data, errors=registered_address_form.address1.errors) }}
+      {{ forms.question_input('city', 'Town or city', value=registered_address_form.city.data, errors=registered_address_form.city.errors) }}
+      {{ forms.question_input('postcode', 'Postcode', value=registered_address_form.postcode.data, errors=registered_address_form.postcode.errors) }}
```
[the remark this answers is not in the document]
Wynndow: yeah, this seems to be a much cleaner way of calling the textbox. I'd like to keep using it as it reduces 8 lines of code to 1. I think I'll move them to the toolkit.

app/templates/suppliers/registered_address.html
```diff
@@ -0,0 +1,147 @@
...
+
+      {% if registered_country_form.registrationCountry.errors %}
+      <div class="validation-wrapper">
+        <style>
+          #location-autocomplete {border: 5px solid #B10E1E;}
+          .autocomplete__hint {border: 5px solid;}
+        </style>
```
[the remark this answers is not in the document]
Wynndow: The reason we've done this is due to the JavaScript the country picker is running. When the JS loads it adds a new input field and gives it the id of `location-autocomplete`. 
Because the input is dynamically generated we can't conditionally assign it a class in the template, so we need to target and update that elements style here.
It's the same for the `autocomplete__hint` class which is defined in a stylesheet brough in by the node package for country picker. We can't edit the existing class or apply a new class to the element we need to (again dynamically generated) so we need to override it here.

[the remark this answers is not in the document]
samuelhwilliams: I guess I'm being dense. If those static styles just lived in application.scss, wouldn't they be applied when the elements get dynamically created by the location picker?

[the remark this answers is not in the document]
Wynndow: No, it is a bit confusing.

We do have some static styles in the application.scss to already override these styles (adds some padding to make the country picker match our other input fields). The issue is that when that field has validation errors, those styles need to be overridden in a different way (increasing the border size and applying the validation colour). Normally in the template we'd have an if statement which applied a new class (with these new styles) to the element when there were validation errors. But the element is not defined in our template, it's generated dynamically on page load by the javascript. So we can't apply a new class to it, so have to update the styles that are already applied.

app/main/views/suppliers.py
Wynndow: I suppose we could do, but if feels a bit weird to add extra info to the thing we got back from the api. It'd mean unpacking it again in the template when it's used, rather than just using this variable.

gulpfile.js
Wynndow: Yeah, I updated the argument name, made sense

app/main/forms/suppliers.py
```diff
@@ -28,20 +29,43 @@ class EditSupplierForm(Form):
 class EditContactInformationForm(Form):
     id = IntegerField()
     contactName = StripWhitespaceStringField('Contact name', validators=[
-        DataRequired(message="You must provide a contact name"),
+        InputRequired(message="You must provide a contact name"),
     ])
     email = EmailField('Contact email address', validators=[
-        DataRequired(message="You must provide an email address"),
+        InputRequired(message="You must provide an email address"),
```
samuelhwilliams: Do we want to add Length validators to these (and other similar places) as well?

app/main/helpers/suppliers.py
```diff
@@ -0,0 +1,28 @@
...
+    countryfile = os.path.join(helpers_path, '../../static/location-autocomplete-canonical-list.json')
+    with open(countryfile) as f:
+        return (json.load(f))
+
+
+def get_country_name_from_country_code(country_code):
+    if country_code and country_code == 'gb':
+        # We need to support the old country code style ('gb') until after the existing country code data we have in the
+        # database has been updated by a script we'll run.
+        return 'United Kingdom'
```
samuelhwilliams: If we plan to remove this after the migration, can we add a TODO here (with 'remove after x') and/or capture it on another ticket?

app/main/views/suppliers.py
```diff
@@ -91,10 +94,81 @@ def supplier_details():
...
+            error=None,
+        ), 400
+
+    registered_address_form.address1.data = supplier['contact'].get('address1')
+    registered_address_form.city.data = supplier['contact'].get('city')
+    registered_address_form.postcode.data = supplier['contact'].get('postcode')
+
+    registered_country_form.registrationCountry.data = supplier.get('registrationCountry')
+
+    return render_template(
```
samuelhwilliams: We could probably refactor this down a bit so that we only have a single render_template on this one template, but not sure if it would be cleaner or not. The two POST-based ones could be combined relatively easily though.

app/templates/suppliers/registered_address.html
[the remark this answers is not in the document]
samuelhwilliams: Righto, totally follow this now. Thanks for explaining. 👍

app/main/forms/suppliers.py
Wynndow: We could. Not sure it's part of this ticket particularly though. We have other tickets that will touch these.

samuelhwilliams: Sure, although changing them to InputRequired isn't part of this ticket either. If we don't do it here, let's make sure we make a ticket to add these validators so we don't forget. It can probably be a tech debt ticket (one of us should just slap our name on it straight away probably though, so it doesn't get lost). Happy to do it if you just want to Ship The Thing.

Wynndow: Yeah I guess. Changing to InputRequired was a two second find and replace, adding validators is a bit more involved I think due to the tests involved and making sure the content of the messages was right etc. Feels like more creep.

I ❤️🚢

app/main/views/suppliers.py
Wynndow: How were you thinking of refactoring the posts?

samuelhwilliams: You could do something like this for the just POST section:

```python
    if request.method == 'POST':
        error = None
        address_valid = registered_address_form.validate_on_submit()
        country_valid = registered_country_form.validate_on_submit()

        if address_valid and country_valid:
            try:
                data_api_client.update_supplier(
                    current_user.supplier_id,
                    registered_country_form.data,
                    current_user.email_address,
                )

                data_api_client.update_contact_information(
                    current_user.supplier_id,
                    supplier['contact']['id'],
                    registered_address_form.data,
                    current_user.email_address
                )
                
            except APIError as e:
                error = e
                
            else:
                return redirect(url_for(".supplier_details"))

        return render_template(
            "suppliers/registered_address.html",
            supplier=supplier,
            countries=COUNTRY_TUPLE,
            registered_address_form=registered_address_form,
            registered_country_form=registered_country_form,
            error=error.message if error else None,
        ), 400
```


Or something like this for the full method (although I think the flow gets a bit confusing to follow here, so this maybe isn't a good idea):
```python
def edit_registered_address():
    try:
        supplier = data_api_client.get_supplier(
            current_user.supplier_id
        )['suppliers']
    except APIError as e:
        abort(e.status_code)
    supplier['contact'] = supplier['contactInformation'][0]

    error = None
    http_status = 200
    registered_address_form = EditRegisteredAddressForm()
    registered_country_form = EditRegisteredCountryForm()

    if request.method == 'POST':
        address_valid = registered_address_form.validate_on_submit()
        country_valid = registered_country_form.validate_on_submit()

        if address_valid and country_valid:
            try:
                data_api_client.update_supplier(
                    current_user.supplier_id,
                    registered_country_form.data,
                    current_user.email_address,
                )

                data_api_client.update_contact_information(
                    current_user.supplier_id,
                    supplier['contact']['id'],
                    registered_address_form.data,
                    current_user.email_address
                )

            except APIError as e:
                error = e

            else:
                return redirect(url_for(".supplier_details"))

        http_status = 400

    else:
        registered_address_form.address1.data = supplier['contact'].get('address1')
        registered_address_form.city.data = supplier['contact'].get('city')
        registered_address_form.postcode.data = supplier['contact'].get('postcode')

        registered_country_form.registrationCountry.data = supplier.get('registrationCountry')

    return render_template(
        "suppliers/registered_address.html",
        supplier=supplier,
        countries=COUNTRY_TUPLE,
        registered_address_form=registered_address_form,
        registered_country_form=registered_country_form,
        error=error.message if error else None,
    ), http_status
```

Wynndow: Ah nice. I totally forgot that `else` was a thing.

Wynndow: Looking at it, I don't think that the flow in the second option is that confusing.

app/main/forms/suppliers.py
samuelhwilliams: Yeah that's fair, it's a bit creepy.

app/main/helpers/__init__.py
samuelhwilliams: These could still be removed.

app/main/views/suppliers.py
```diff
@@ -91,13 +94,75 @@ def supplier_details():
...
+
+        registered_country_form.registrationCountry.data = supplier.get('registrationCountry')
+
+    return render_template(
+        "suppliers/registered_address.html",
+        supplier=supplier,
+        countries=COUNTRY_TUPLE,
+        registered_address_form=registered_address_form,
+        registered_country_form=registered_country_form,
+        error=error,
```
samuelhwilliams: Is this definitely correct? Not sure if there needs to be the `error.message if error else None` logic here for it to render the correct output. I think we used `e.message` before when we rendered it in the except block.

app/templates/suppliers/registered_address.html
```diff
@@ -0,0 +1,126 @@
...
+        "link": "/",
+        "label": "Digital Marketplace"
+      },
+      {
+        "link": url_for(".dashboard"),
+        "label": "Your account"
+      },
+      {
+        "link": url_for(".supplier_details"),
+        "label": "Supplier details"
```
samuelhwilliams: I think we're going for "Company details" as the breadcrumb now. That's what's in the prototype kit at the moment. Kev's going to make a ticket to go through and make sure everything else is updated, but we can probably do this bit now.

samuelhwilliams: https://trello.com/c/yxK7OIOC/83-update-breadcrumbs-for-supplier-details-to-company-details

app/main/views/suppliers.py
Wynndow: Yep, you're right. I've update. Also added a test that should have caught this.

app/templates/suppliers/registered_address.html
Wynndow: Cool, have changed.

.flake8
```diff
@@ -10,8 +10,7 @@ ignore = D203,W503
 max-complexity = 7
 max-line-length = 120
 per-file-ignores =
-    **/__init__.py : F401
     app/__init__.py : E402
-    app/main/__init__.py : E402
-    app/status/__init__.py : E402
+    app/main/__init__.py : E402, F401
+    app/status/__init__.py : E402, F401
```
samuelhwilliams: 👍

PR #791, opened 2018-02-13
Revert back to eea/odfpy

samuelhwilliams:
Update utils to use eea/odfpy rather than our interim alphagov/odfpy
fork, which we created to allow us to patch a bug. Now that the bug has
been accepted into the main library as of release 1.3.6
(https://github.com/eea/odfpy/pull/72) we no longer need to support
own our fork.

Also adds a comment when registering the external blueprint from 
dm-utils to ensure it is registered last.

### DIFF
--- a/tests/app/main/test_services.py
+++ b/tests/app/main/test_services.py
@@ -1618,7 +1618,7 @@ class TestEditDraftService(BaseApplicationTest):
 
         s3.return_value.save.assert_called_once_with(
             'g-cloud-7/submissions/1234/1-service-definition-document-2015-01-02-0304.pdf',
-            mock.ANY, acl='private'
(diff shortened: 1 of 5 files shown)


tests/app/main/test_services.py
```diff
@@ -1618,7 +1618,7 @@ def test_file_upload(self, data_api_client, s3):
 
         s3.return_value.save.assert_called_once_with(
             'g-cloud-7/submissions/1234/1-service-definition-document-2015-01-02-0304.pdf',
-            mock.ANY, acl='private'
+            mock.ANY, acl='bucket-owner-full-control'
```
galund: Needs mentioning in the commit message too?

samuelhwilliams: Done. 👍

PR #794, opened 2018-02-16
Handle suppliers with no `registrationCountry` key

Wynndow:
When suppliers are serialized by the api any `None` value keys are
dropped before being returned in the response.

The `registrationCountry` field in the suppliers table is a nullable
field. Whilst in practice in production we don't have any null values in
this field (it's either an empty string or 'gb' at the time of writing),
we should cover this case just in case.

This issue was picked up by the functional tests on preview. The
suppliers generated by the functional tests have null values for
`registrationCountry`. This may also need looking at.

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -94,7 +94,7 @@ def supplier_details():
     except APIError as e:
         abort(e.status_code)
     supplier['contact'] = supplier['contactInformation'][0]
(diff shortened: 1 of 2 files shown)


samuelhwilliams:
And now I'm reading https://medium.com/@cvitullo/overly-defensive-programming-e7a1b3d234c2 and thinking we shouldn't patch this because we will always expect to get data back for this key and we should instead have just fixed our functional test. 🤔

PR #796, opened 2018-02-19
Add edit trading status page

samuelhwilliams:
## Summary
Adds a page for suppliers to edit their trading status.

## Ticket
https://trello.com/c/Wvvtb1yW/52-new-editable-page-trading-status

### DIFF
--- a/app/main/forms/suppliers.py
+++ b/app/main/forms/suppliers.py
@@ -142,3 +142,40 @@ class CompanyOrganisationSizeForm(Form):
     organisation_size = RadioField('Organisation size',
                                    validators=[InputRequired(message="You must choose an organisation size.")],
                                    choices=[(o['value'], o['label']) for o in OPTIONS])
+
+
+class CompanyTradingStatusForm(Form):
+    OPTIONS = [
+        {
+            'value': 'limited company (LTD)',
+            'label': 'limited company (LTD)',
+        },
+        {
+            'value': 'limited liability company (LLC)',
+            'label': 'limited liability company (LLC)',
+        },
+        {
+            'value': 'public limited company (PLC)',
+            'label': 'public limited company (PLC)',
+        },
+        {
+            'value': 'limited liability partnership (LLP)',
+            'label': 'limited liability partnership (LLP)',
+        },
+        {
+            'value': 'sole trader',
+            'label': 'sole trader',
+        },
+        {
+            'value': 'public body',
+            'label': 'public body',
+        },
+        {
+            'value': 'other',
+            'label': 'other',
+        },
+    ]
+
+    trading_status = RadioField('Trading status',
+                                validators=[InputRequired(message="You must choose a trading status.")],
+                                choices=[(o['value'], o['label']) for o in OPTIONS])
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -18,6 +18,7 @@ from ..forms.suppliers import (
     CompanyContactDetailsForm,
     CompanyNameForm,
     CompanyOrganisationSizeForm,
+    CompanyTradingStatusForm,
     DunsNumberForm,
     EditContactInformationForm,
     EditRegisteredAddressForm,
@@ -265,9 +266,14 @@ def edit_supplier_organisation_size():
 
     if request.method == 'POST':
         if form.validate_on_submit():
-            data_api_client.update_supplier(supplier_id=current_user.supplier_id,
-                                            supplier={"organisationSize": form.organisation_size.data},
-                                            user=current_user.email_address)
+            try:
+                data_api_client.update_supplier(supplier_id=current_user.supplier_id,
+                                                supplier={"organisationSize": form.organisation_size.data},
+                                                user=current_user.email_address)
+
+            except APIError as e:
+                abort(e.status_code)
+
             return redirect(url_for('.supplier_details'))
 
         current_app.logger.warning(
@@ -283,6 +289,45 @@ def edit_supplier_organisation_size():
     form.organisation_size.data = supplier.get('organisationSize', None)
 
     return render_template('suppliers/edit_supplier_organisation_size.html', form=form)
+
+
+@main.route('/trading-status/edit', methods=['GET', 'POST'])
+@login_required
+def edit_supplier_trading_status():
+    form = CompanyTradingStatusForm()
+
+    if request.method == 'POST':
+        api_error = None
+        if form.validate_on_submit():
+            try:
+                data_api_client.update_supplier(supplier_id=current_user.supplier_id,
+                                                supplier={"tradingStatus": form.trading_status.data},
+                                                user=current_user.email_address)
+
+            except APIError as e:
+                abort(e.status_code)
+
+            return redirect(url_for('.supplier_details'))
+
+        current_app.logger.warning(
+            "supplieredit.fail: trading-status:{tstatus}, errors:{tstatus_errors}",
+            extra={
+                'tstatus': form.trading_status.data,
+                'tstatus_errors': ",".join(form.trading_status.errors)
+            })
+
+        return render_template("suppliers/edit_supplier_trading_status.html", form=form, api_error=api_error), 400
+
+    supplier = data_api_client.get_supplier(current_user.supplier_id)['suppliers']
+
+    prefill_trading_status = None
+    if supplier.get('tradingStatus'):
+        if supplier['tradingStatus'] in map(lambda x: x['value'], form.OPTIONS):
+            prefill_trading_status = supplier['tradingStatus']
+
+    form.trading_status.data = prefill_trading_status
+
+    return render_template('suppliers/edit_supplier_trading_status.html', form=form)
 
 
 @main.route('/supply', methods=['GET'])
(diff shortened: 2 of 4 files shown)


app/main/forms/suppliers.py
```diff
@@ -139,3 +139,40 @@ class CompanyOrganisationSizeForm(Form):
     organisation_size = RadioField('Organisation size',
                                    validators=[InputRequired(message="You must choose an organisation size.")],
                                    choices=[(o['value'], o['label']) for o in OPTIONS])
+
+
+class CompanyTradingStatusForm(Form):
+    OPTIONS = [
+        {
+            'value': 'limited company (LTC)',
```
TheDoubleK: Do we want to store the whole thing with the acronym as a value?  Or we could stick with the existing radio button values, ie just "public limited company", "limited company", "limited liability company", "limited liability partnership"?

Also, we've never had "public bodies" as an option before, so if we make everyone else with existing data "other" then some of these might have been "public bodies".... I guess that's a question for the migration though.

app/main/views/suppliers.py
```diff
@@ -263,6 +264,45 @@ def edit_supplier_organisation_size():
...
+    supplier = data_api_client.get_supplier(current_user.supplier_id)['suppliers']
+
+    prefill_trading_status = None
+
+    if supplier.get('tradingStatus'):
+        if supplier['tradingStatus'] in map(lambda x: x['value'], form.OPTIONS):
+            prefill_trading_status = supplier['tradingStatus']
+
+        else:
+            # The last option in the form is 'other', which we'll use as a fallback if a value exists but doesn't match
```
TheDoubleK: But we're going to have fixed these with a data migration?  Still nice to have the fallback I guess, but I wonder if it's better to not prefill at all in the case we don't have a match, to force people to make a fresh choice?

app/templates/suppliers/edit_supplier_trading_status.html
```diff
@@ -0,0 +1,60 @@
...
+      "link": url_for('.supplier_details'),
+      "label": "Company details"
+    },
+  ] %}
+    {% include "toolkit/breadcrumb.html" %}
+  {% endwith %}
+{% endblock %}
+
+{% block main_content %}
+
```
TheDoubleK: We need a validation masthead here 😛

app/main/forms/suppliers.py
```diff
@@ -139,3 +139,40 @@ class CompanyOrganisationSizeForm(Form):
...
+            'label': 'limited company (LTC)',
```
TheDoubleK: "limited company" should be "LTD" rather than "LTC", according to the ticket.

tests/app/main/test_suppliers.py
```diff
@@ -2282,3 +2282,63 @@ def test_existing_org_size_sets_current_selection(self, data_api_client, existin
...
+            res = self.client.get("/suppliers/trading-status/edit")
+            assert res.status_code == 200, 'The edit trading-status page has not loaded correctly.'
+
+    def test_no_selection_triggers_input_required_validation(self, data_api_client):
+        with self.app.test_client():
+            self.login()
+
+            res = self.client.post("/suppliers/trading-status/edit")
+            doc = html.fromstring(res.get_data(as_text=True))
+            error = doc.xpath('//span[@id="error-trading_status"]')
```
TheDoubleK: Add assertion for validation masthead here :)

tests/app/main/test_suppliers.py
```diff
@@ -2282,3 +2282,63 @@ def test_existing_org_size_sets_current_selection(self, data_api_client, existin
...
+
+            assert len(error) == 1, \
+                'Only one validation message should be shown.'
+
+            assert error[0].text.strip() == 'You must choose a trading status.', \
+                'The validation message is not as expected.'
+
+    @pytest.mark.parametrize('trading_status', (None, 'limited company (LTC)', 'other'))
+    def test_post_choice_triggers_api_supplier_update_and_redirect(self, data_api_client, trading_status):
```
TheDoubleK: Could also have a test for error handling if the POST fails for whatever reason.

app/main/forms/suppliers.py
samuelhwilliams: We display this back to the user on the summary page, and have nothing else that would translate it back ala what the content loader does, so I think we do need to store the entire value so that we can present back to the user the same thing they selected.

samuelhwilliams: As for the migration, that's exactly the point I've raised on the scripts PR.

app/main/views/suppliers.py
samuelhwilliams: https://github.com/alphagov/digitalmarketplace-scripts/pull/198

app/templates/suppliers/edit_supplier_trading_status.html
samuelhwilliams: Added

app/main/forms/suppliers.py
samuelhwilliams: Correct, updated.

tests/app/main/test_suppliers.py
samuelhwilliams: Yep

tests/app/main/test_suppliers.py
samuelhwilliams: Added.

samuelhwilliams:
@TheDoubleK all comments addressed

tests/app/helpers.py
```diff
@@ -493,6 +494,11 @@ def assert_no_flashes(self):
         with self.client.session_transaction() as session:
             assert not session.get("_flashes")
 
+    def assert_validation_masthead(self, res, heading="There was a problem with your answer to:"):
```
TheDoubleK: I like this... should it go into the test-utils repo to be shared across all apps?

app/main/views/suppliers.py
```diff
@@ -242,27 +243,84 @@ def edit_supplier_organisation_size():
...
 
-        current_app.logger.warning(
-            "supplieredit.fail: organisation-size:{osize}, errors:{osize_errors}",
-            extra={
-                'osize': form.organisation_size.data,
-                'osize_errors': ",".join(form.organisation_size.errors)
-            })
+            except APIError as e:
+                api_error = e.message
+                current_app.logger.warning(f"supplieredit.fail: api error:{api_error}")
```
TheDoubleK: Does this need the `extra` to set the api_error value?

samuelhwilliams: Nah, it's a formatted string.

tests/app/helpers.py
samuelhwilliams: Probably! 🤔

samuelhwilliams: Given the first PR on test-utils hasn't gone in yet, so there's nothing to work off, I'll add this to my list of things to do. If we can go ahead with this as-is for now, I'll refactor it out later. 👍

app/main/views/suppliers.py
TheDoubleK: Ah, but I think something to do with the way our JSON logging works might need it done with the "extra" fields, and that's why it's done that way everywhere else we log stuff.

app/templates/suppliers/edit_supplier_organisation_size.html
```diff
@@ -25,7 +25,11 @@
 
 <div class="single-question-page">
   <div class="grid-row">
-    {% if form.errors %}
+    {% if api_error %}
```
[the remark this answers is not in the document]
Wynndow: If we're going to remove the api_error validation mastheads, it'll need to come off the country picker page too.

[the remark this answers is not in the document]
samuelhwilliams: But I think you're right that we probably shouldn't suppress the error here, so I'll go ahead and let it bubble up.

app/main/views/suppliers.py
samuelhwilliams: Well, it's gone now anyway. 👍

samuelhwilliams:
@TheDoubleK updated and should be good now

app/main/views/suppliers.py
```diff
@@ -276,25 +277,78 @@ def edit_supplier_organisation_size():
...
         if form.validate_on_submit():
-            data_api_client.update_supplier(supplier_id=current_user.supplier_id,
-                                            supplier={"organisationSize": form.organisation_size.data},
-                                            user=current_user.email_address)
+            try:
+                data_api_client.update_supplier(supplier_id=current_user.supplier_id,
+                                                supplier={"organisationSize": form.organisation_size.data},
+                                                user=current_user.email_address)
+
+            except APIError as e:
```
TheDoubleK: I think we agreed to do do this for API failures?
```
except APIError as e:
    abort(e.status_code)
```

TheDoubleK: To be consistent with elsewhere

TheDoubleK: Actually, this might not be part of your ticket and will go away after a rebase...?

app/main/views/suppliers.py
TheDoubleK: Yeah, so now the script has run we have no data in the database that doesn't match one of these options.  So I don't think we need to force the prefill of "other" at all - better to make people choose an option if we don't have anything for them.

samuelhwilliams: Yep nice, removed.

samuelhwilliams:
Updated @TheDoubleK

PR #797, opened 2018-02-20
Add validation mastheads

samuelhwilliams:
## Summary
When https://trello.com/c/eB9u3xe7/40 was getting QA'd, it was highlighted that we hadn't added a validation masthead at the top of the page to highlight that something was wrong. This has now been added and retrofitted to some other existing single-question-pages that were missing the same, so that we're more consistent.

## Note
Checking with @laurenceberry whether the masthead should be inside the column-two-thirds or outside.

## Ticket
https://trello.com/c/eB9u3xe7/40

### DIFF
--- a/app/templates/suppliers/duns_number.html
+++ b/app/templates/suppliers/duns_number.html
@@ -24,54 +24,58 @@
 {% endblock %}
 
 {% block main_content %}
-{% if form.duns_number.errors[0] == 'DUNS number already used' %}
-<div class="validation-masthead" role="group" aria-labelledby="validation-masthead-heading">
-  <h1 class="validation-masthead-heading" id="validation-masthead-heading">
-      A supplier account already exists with that DUNS number
-  </h1>
-  <p class="validation-masthead-description">
-      If you no longer have your account details, or if you think this may be an error, email <a href="mailto:enquiries@digitalmarketplace.service.gov.uk?subject=DUNS%20number%20question" title="Please contact enquiries@digitalmarketplace.service.gov.uk">enquiries@digitalmarketplace.service.gov.uk</a>
-  </p>
-</div>
-{% endif %}
 
-<div class="grid-row">
-  <div class="column-two-thirds">
-    {% with heading = "Enter your DUNS number",
-            smaller = True %}
-      {% include "toolkit/page-heading.html" %}
-    {% endwith %}
-  
+<div class="single-question-page">
+  <div class="grid-row">
+    {% if form.duns_number.errors[0] == 'DUNS number already used' %}
+      {% with lede = "A supplier account already exists with that DUNS number",
+              description = 'If you no longer have your account details, or if you think this may be an error, email <a href="mailto:enquiries@digitalmarketplace.service.gov.uk?subject=DUNS%20number%20question" title="Please contact enquiries@digitalmarketplace.service.gov.uk">enquiries@digitalmarketplace.service.gov.uk</a>'|safe
+      %}
+        {% include 'toolkit/forms/validation.html' %}
+      {% endwith %}
+    {% elif form.duns_number.errors %}
+      {% with errors = [{'question': form.duns_number.label, 'input_name': form.duns_number.name}] %}
+        {% include 'toolkit/forms/validation.html' %}
+      {% endwith %}
+    {% endif %}
     
-    <form method="POST" action="{{ url_for('.submit_duns_number') }}">
-      <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
-      <div class="dmspeak">
-        <p>If you registered your business with Companies House, you will automatically have been given a unique
-          DUNS number.</p>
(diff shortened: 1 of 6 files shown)


app/templates/suppliers/duns_number.html
```diff
@@ -24,54 +24,55 @@
...
-      </div>
-        <p class="lead">You can either:</p>
-        <ul class="list-bullet">
-          <li><a href="https://www.dnb.co.uk/duns-number/lookup.html" rel="external">find your DUNS number</a> on the
-            Dun &amp; Bradstreet website</li>
-          <li><a href="https://www.dnb.co.uk/duns-number/lookup/request-a-duns-number.html" rel="external">apply for
-            a DUNS number</a> if you don&rsquo;t have one</li>
-        </ul>
+      {% if form.duns_number.errors[0] == 'DUNS number already used' %}
+        {% with lede = "A supplier account already exists with that DUNS number",
```
CrystalPea: what's lede?

samuelhwilliams: It means "the opening sentence or paragraph of a news article, summarizing the most important aspects of the story."

It's a parameter that the `validation.html` macro looks for.

CrystalPea: I mean, I see it in validation form macro, but what does the name refer to, do you maybe know?

CrystalPea: Ahh, ok, thanks a lot :3

TheDoubleK:
This looks different from our other banner messages - everywhere else they are full-width I think.
screen shot 2018-02-20 at 12 21 19

Also, the DUNS number one doesn't work for empty input:
screen shot 2018-02-20 at 12 22 34

samuelhwilliams:
Re: full-width/half-width, see notes on PR. Waiting for Laurence to give feedback on which it should be.

Re: duns number validation on no input - the existing template specifically excluded a masthead in this instance. I didn't override the decision in this case - I just converted it to using the validation macro. I guess we could add a masthead for no input, but it feels like there was a conscious decision to exclude this case... for some reason.

Maybe no good reason. 🤔

TheDoubleK:
I've had a bit of a dig round...  surprisingly, we are slightly inconsistent in whether or not the validation masthead is inside or outside the 2/3 column.  But the vast majority (including the base "service edit" template that is used for all the service submission form pages) have it outside.  It's only a couple of pages in the "return your signed framework agreement" flow that have it inside the 2/3 column.

## Outside

https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/c93cb30c3af6e81929f1581590d373cb89bc3bbd/app/templates/services/_base_edit_section_page.html#L29

https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/7cce1d5e694fccc46f39bf32a6e19ae390c3b743/app/templates/frameworks/signature_upload.html#L44

https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/d8e1c22372e17016cf85a0ad9b694b1c546fa8a0/app/templates/frameworks/reuse_declaration.html#L37

https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/5949fb1e6feefcadb7971b64cb61dc3dbd32d203/app/templates/frameworks/contract_variation.html#L31

https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/2b4626cb13d4a0e99dab0f26b7aac3b4602782b6/app/templates/frameworks/edit_declaration_section.html#L35

https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/77919075e66886ba432211c02996aa58091e5a3d/app/templates/frameworks/updates.html#L59

## Inside

https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/7cce1d5e694fccc46f39bf32a6e19ae390c3b743/app/templates/frameworks/signer_details.html#L42

https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/7cce1d5e694fccc46f39bf32a6e19ae390c3b743/app/templates/frameworks/contract_review.html#L45

laurenceberry:
I doubt there was a good reason not to include the masthead, it was possibly because we didn't use to use mastheads on single question pages which we've agreed to do from now on.

Either way, mastheads should be full width.

TheDoubleK:
`the existing template specifically excluded a masthead in this instance` - no, it's the other way round: the template specifically *included* a masthead in the instance that a duplicate clashing DUNS number had been entered.  The page previously followed our pattern of "no masthead if a single question page", but DUNS number was a "special case" because we wanted to direct people to contact support if their DUNS number was already used, so we added a special banner message for that.

So I think if we're now adding banners for all single-question pages then the empty/invalid DUNS number case should also have one.  But we still need the special messaging for if there's a DUNS number clash.

Here's the PR that added the message:
https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/186

samuelhwilliams:
@TheDoubleK comments addressed, tests updated.

PR #799, opened 2018-02-20
Rearrange Supplier Details page elements and change headers

CrystalPea:
Trello ticket: https://trello.com/c/VF5yc1pd/50-supplier-account-1-update-supplier-details-page
- rename headers
- rename 'Edit' to 'Change' link
- add 'Optional' next to summary field
- Move address field to second section of the page and add registration country to it
- remove 'Registration Country' as a separate question
- write a test for address field
- adjust existing tests

Also 2nd Trello ticket: https://trello.com/c/A3ovgHlk/51-remove-address-from-edit-supplier-details-page

- Delete address details from "Edit supplier" page
- Change header and breadcrumb for 'What buyers will see' page
- Move validation errors banner above the page header
-  Rename route/page from 'edit_supplier' to 'edit_what_buyers_will_see' 

ADDITIONALLY:
- unite GET and POST routes for edit_supplier contact information

Screenshot of the updated "Supplier details" page:
screen shot 2018-02-20 at 14 36 03

____

Screenshot of the updated  "What buyers will see"page (content checked with @constancecerf ):
screen shot 2018-02-26 at 14 49 41

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -215,7 +215,6 @@ def edit_what_buyers_will_see():
         abort(e.status_code)
 
     supplier['contact'] = supplier['contactInformation'][0]
-    error = None
     http_status = 200
 
     supplier_form = EditSupplierForm()
@@ -240,7 +239,7 @@ def edit_what_buyers_will_see():
                     current_user.email_address
                 )
             except APIError as e:
-                error = e.message
+                abort(e.status_code)
             else:
                 return redirect(url_for(".supplier_details"))
 
@@ -254,7 +253,6 @@ def edit_what_buyers_will_see():
 
     return render_template(
         "suppliers/edit_what_buyers_will_see.html",
-        error=error,
         supplier_form=supplier_form,
         contact_form=contact_form
     ), http_status
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -204,9 +204,9 @@ def edit_supplier_redirect():
     return redirect(url_for('.edit_supplier'), 302)
 
 
-@main.route('/details/edit', methods=['GET'])
+@main.route('/details/edit', methods=['GET', 'POST'])
 @login_required
-def edit_supplier(supplier_form=None, contact_form=None, error=None):
+def edit_supplier():
     try:
         supplier = data_api_client.get_supplier(
             current_user.supplier_id
@@ -214,60 +214,50 @@ def edit_supplier(supplier_form=None, contact_form=None, error=None):
     except APIError as e:
         abort(e.status_code)
 
-    if supplier_form is None:
-        supplier_form = EditSupplierForm(
-            description=supplier.get('description', None),
-            clients=supplier.get('clients', None)
-        )
-        contact_form = EditContactInformationForm(
-            prefix='contact_',
-            **supplier['contactInformation'][0]
-        )
+    supplier['contact'] = supplier['contactInformation'][0]
+    error = None
+    http_status = 200
+
+    supplier_form = EditSupplierForm()
+    contact_form = EditContactInformationForm()
+
+    if request.method == 'POST':
+        supplier_info_valid = supplier_form.validate_on_submit()
+        contact_info_valid = contact_form.validate_on_submit()
+
+        if supplier_info_valid and contact_info_valid:
+            try:
+                data_api_client.update_supplier(
+                    current_user.supplier_id,
+                    supplier_form.data,
+                    current_user.email_address
+                )
+
+                data_api_client.update_contact_information(
+                    current_user.supplier_id,
+                    supplier['contact']['id'],
+                    contact_form.data,
+                    current_user.email_address
+                )
+            except APIError as e:
+                error = e.message
+            else:
+                return redirect(url_for(".supplier_details"))
+
+        http_status = 400
+
+    else:
+        supplier_form.description.data = supplier.get('description', None)
+        contact_form.contactName.data = supplier['contact'].get('contactName')
+        contact_form.phoneNumber.data = supplier['contact'].get('phoneNumber')
+        contact_form.email.data = supplier['contact'].get('email')
 
     return render_template(
         "suppliers/edit_supplier.html",
         error=error,
         supplier_form=supplier_form,
         contact_form=contact_form
-    ), 200
-
-
-@main.route('/details/edit', methods=['POST'])
-@login_required
-def update_supplier():
-    # FieldList expects post parameter keys to have number suffixes
-    # (eg client-0, client-1 ...), which is incompatible with how
-    # JS list-entry plugin generates input names. So instead of letting
-    # the form search for request keys we pass in the values directly as data
-    supplier_form = EditSupplierForm(
-        formdata=None,
-        description=request.form['description'],
-    )
-
-    contact_form = EditContactInformationForm(prefix='contact_')
-
-    if not (supplier_form.validate_on_submit() and contact_form.validate_on_submit()):
-        return edit_supplier(supplier_form=supplier_form, contact_form=contact_form)
-
-    try:
-        data_api_client.update_supplier(
-            current_user.supplier_id,
-            supplier_form.data,
-            current_user.email_address
-        )
-
-        data_api_client.update_contact_information(
-            current_user.supplier_id,
-            contact_form.id.data,
-            contact_form.data,
-            current_user.email_address
-        )
-    except APIError as e:
-        return edit_supplier(supplier_form=supplier_form,
-                             contact_form=contact_form,
-                             error=e.message)
-
-    return redirect(url_for(".supplier_details"))
+    ), http_status
 
 
 @main.route('/organisation-size/edit', methods=['GET', 'POST'])
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -199,14 +199,14 @@ def edit_supplier_registered_name():
 
 @main.route('/edit', methods=['GET'])
 @login_required
-def edit_supplier_redirect():
+def edit_what_buyers_will_see_redirect():
     # redirect old route for this view
-    return redirect(url_for('.edit_supplier'), 302)
+    return redirect(url_for('.edit_what_buyers_will_see'), 302)
 
 
-@main.route('/details/edit', methods=['GET', 'POST'])
+@main.route('/what-buyers-will-see/edit', methods=['GET', 'POST'])
 @login_required
-def edit_supplier():
+def edit_what_buyers_will_see():
     try:
         supplier = data_api_client.get_supplier(
             current_user.supplier_id
@@ -253,7 +253,7 @@ def edit_supplier():
         contact_form.email.data = supplier['contact'].get('email')
 
     return render_template(
-        "suppliers/edit_supplier.html",
+        "suppliers/edit_what_buyers_will_see.html",
         error=error,
         supplier_form=supplier_form,
         contact_form=contact_form
--- a/app/templates/suppliers/details.html
+++ b/app/templates/suppliers/details.html
@@ -23,57 +23,48 @@
 {% block main_content %}
 {%
   with
-    heading = "Supplier details",
+    heading = "Company details",
     smaller = true
 %}
   {% include "toolkit/page-heading.html" %}
 {% endwith %}
 
-
-{{ summary.top_link('Edit', url_for('.edit_supplier')) }}
+{{ summary.heading("What buyers will see", id="what_buyers_will_see") }}
+{{ summary.top_link('Change', url_for('.edit_supplier')) }}
 {% call(item) summary.mapping_table(
-  caption='Supplier details',
+  caption='What buyers will see',
   field_headings=[
     'Label',
-    'Value'
+    'Value',
+    'Optional'
   ],
   field_headings_visible=False
 ) %}
   {% call summary.row() %}
     {{ summary.field_name('Contact name') }}
     {{ summary.text(supplier.contact.contactName) }}
+    {{ summary.text("") }}
   {% endcall %}
   {% call summary.row() %}
     {{ summary.field_name('Contact email') }}
     {{ summary.text(supplier.contact.email) }}
+    {{ summary.text("") }}
   {% endcall %}
   {% call summary.row() %}
     {{ summary.field_name('Contact phone number') }}
     {{ summary.text(supplier.contact.phoneNumber) }}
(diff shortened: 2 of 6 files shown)


TheDoubleK:
https://imgur.com/gallery/6SkZ6um

app/templates/suppliers/details.html
```diff
@@ -23,57 +23,48 @@
...
-        {% include "toolkit/contact-details.html" %}
-      {% endwith %}
-    {% endcall %}
-  {% endcall %}
-  {% call summary.row() %}
-    {{ summary.field_name('Supplier summary') }}
+    {{ summary.field_name('Summary') }}
     {{ summary.text(supplier.description) }}
+    {% call summary.field(action=True) %}
+      <font color="grey">Optional</font>
```
TheDoubleK: We shouldn't be embedding font styles in our templates. I think if we really want grey text then it should become part of the toolkit pattern with a semantic meaning, e.g. `call summary.field(action=True, hint=True)`. 

For now I'd say go with normal text here, unless @laurenceberry is really strong on it being grey, in which case add the parameter to the toolkit.  Rather than grey it could go in brackets? i.e. `(Optional)` in the final column?

tests/app/main/test_suppliers.py
```diff
@@ -743,9 +743,8 @@ def test_shows_supplier_info(self, data_api_client):
             ):
                 assert document.xpath("//*[normalize-space(string())=$t]", t=property_str), property_str
 
-            # Registration country and registration number not shown if Companies House ID exists
-            for property_str in ("Belize", "BEL153",):
-                assert property_str not in page_html
+            # Registrationnnumber not shown if Companies House ID exists
```
TheDoubleK: Missing a space here

tests/app/main/test_suppliers.py
```diff
@@ -799,6 +798,27 @@ def test_handles_supplier_with_no_registration_country_key(self, data_api_client
...
+
+            address1 = document.xpath("//span[text()='Registered company address']/following::td[1]/span/p/span[1]")
+            town = document.xpath("//span[text()='Registered company address']/following::td[1]/span/p/span[2]")
+            country = document.xpath("//span[text()='Registered company address']/following::td[1]/span/p/span[3]")
+            postcode = document.xpath("//span[text()='Registered company address']/following::td[1]/span/p/span[4]")
+
+            assert "1 Street" in address1[0].text
+            assert "Supplierville" in town[0].text
+            assert "United Kingdom" in country[0].text
+            assert "11 AB" in postcode[0].text
```
TheDoubleK: I think the country should be after the postcode in the address - we'll need to update the frontend toolkit pattern to fix this.

app/templates/suppliers/details.html
laurenceberry: OK, I agree the normal text styling is the best thing to do right now rather than update the frontend toolkit, that is a bit of a nice-to-have.

tests/app/main/test_suppliers.py
CrystalPea: I'm on it. Here is a frontend-toolkit pull request: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/407

I will proceed to update test and toolkit version in supplier-frontend now :).

CrystalPea:
@TheDoubleK all addressed ^_^

app/templates/suppliers/details.html
```diff
@@ -23,57 +23,48 @@
...
-      %}
-        {% include "toolkit/contact-details.html" %}
-      {% endwith %}
-    {% endcall %}
-  {% endcall %}
-  {% call summary.row() %}
-    {{ summary.field_name('Supplier summary') }}
+    {{ summary.field_name('Summary') }}
     {{ summary.text(supplier.description) }}
+    {% call summary.field(action=True) %}
```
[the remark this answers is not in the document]
CrystalPea: Hi @samuelhwilliams I tried changing 'Optional' how we discussed yesterday from field with action to test field, but when I do it, the word moves and looks weird <_<.

screen shot 2018-02-22 at 11 54 08

app/templates/suppliers/edit_what_buyers_will_see.html
```diff
@@ -0,0 +1,106 @@
...
+  </div>
+
+  <form action="{{ url_for('.edit_what_buyers_will_see') }}" method="post" enctype="multipart/form-data">
+
+  <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
+
+  <div class="grid-row">
+    <div class="column-two-thirds">
+
+      {{ forms.question_textbox('contactName', 'Contact name', contact_form.contactName.data, 'Primary contact', errors=contact_form.contactName.errors) }}
```
Wynndow: We don't have the 'Primary contact' hint in the prototype, not sure if we should have it here?

app/templates/suppliers/edit_what_buyers_will_see.html
```diff
@@ -0,0 +1,106 @@
...
+      {{ forms.question_textbox('email', 'Contact email', contact_form.email.data, errors=contact_form.email.errors) }}
```
Wynndow: Prototype has "Contact email address"

app/templates/suppliers/edit_what_buyers_will_see.html
```diff
@@ -0,0 +1,106 @@
...
+
+
+      {{ forms.question_textarea('description', 'Supplier summary', supplier_form.description.data, '50 words max', errors=supplier_form.description.errors, max_length_in_words=50) }}
+    </div>
+  </div>
+
+  {%
+    with
+    type = "save",
+    label = "Save and return"
```
Wynndow: Label on the prototype is "Continue". This sounds better but maybe worth checking?

app/templates/suppliers/edit_what_buyers_will_see.html
```diff
@@ -0,0 +1,106 @@
...
+  %}
+    {% include "toolkit/button.html" %}
+  {% endwith %}
+  <p>
+    <a href="{{ url_for('.supplier_details') }}">Return without saving</a>
```
Wynndow: Again, slightly different to prototype which has "Previous page" instead of "Return without saving"

app/main/views/suppliers.py
```diff
@@ -165,75 +165,65 @@ def edit_registered_address():
...
 @login_required
-def edit_supplier_redirect():
+def edit_what_buyers_will_see_redirect():
     # redirect old route for this view
-    return redirect(url_for('.edit_supplier'), 302)
+    return redirect(url_for('.edit_what_buyers_will_see'), 302)
 
 
-@main.route('/details/edit', methods=['GET'])
+@main.route('/what-buyers-will-see/edit', methods=['GET', 'POST'])
```
Wynndow: Do we need to change this URL? Will it upset Ash with regards to analytics?

app/main/views/suppliers.py
```diff
@@ -165,75 +165,65 @@ def edit_registered_address():
...
 
-    if not (supplier_form.validate_on_submit() and contact_form.validate_on_submit()):
-        return edit_supplier(supplier_form=supplier_form, contact_form=contact_form)
+                data_api_client.update_contact_information(
+                    current_user.supplier_id,
+                    supplier['contact']['id'],
+                    contact_form.data,
+                    current_user.email_address
+                )
+            except APIError as e:
```
Wynndow: I think we decided yesterday that instead of catching API errors like this, we'd abort instead and let the apps error handling deal with it. As we're validating on the frontend we should only ever get an error if there's a network problem I think. You can see how Kev's done it here: https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/47-kk-registered-name-page/app/main/views/suppliers.py#L185

app/main/views/suppliers.py
```diff
@@ -165,75 +165,65 @@ def edit_registered_address():
...
-            current_user.supplier_id,
-            contact_form.id.data,
-            contact_form.data,
-            current_user.email_address
-        )
-    except APIError as e:
-        return edit_supplier(supplier_form=supplier_form,
-                             contact_form=contact_form,
-                             error=e.message)
+    else:
```
Wynndow: I don't think you need the else here?

app/templates/suppliers/details.html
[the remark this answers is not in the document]
Wynndow: I can't see anything in the toolkit for right aligning text. Maybe you could add it? This seems to work though.

app/templates/suppliers/details.html
```diff
@@ -139,5 +137,12 @@
     {% endcall %}
   {% endif %}
 
+  {% if supplier.dunsNumber %}
```
Wynndow: I think VAT number should be moved to just above DUNS as per the prototype

app/templates/suppliers/edit_what_buyers_will_see.html
```diff
@@ -0,0 +1,106 @@
...
+      }
+    ]
+  %}
+    {% include "toolkit/breadcrumb.html" %}
+  {% endwith %}
+{% endblock %}
+
+{% block main_content %}
+
+{% if error %}
```
Wynndow: If we don't handle APIErrors we can get rid of this block.

app/templates/suppliers/edit_what_buyers_will_see.html
```diff
@@ -0,0 +1,106 @@
...
+<div class="validation-masthead" aria-labelledby="validation-masthead-heading">
+  <h3 class="validation-masthead-heading" id="validation-masthead-heading">
+    {{ error }}
+  </h3>
+</div>
+{% endif %}
+{% if supplier_form.errors or contact_form.errors %}
```
Wynndow: You might be able to reduce this down to just using the validation macro. You could build a list of form errors in the view and pass it in? Like this: https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/47-kk-registered-name-page/app/templates/suppliers/edit_registered_name.html#L29

app/templates/suppliers/edit_what_buyers_will_see.html
```diff
@@ -0,0 +1,106 @@
...
+      {% endwith %}
+
+      <div class="dmspeak">
+        <p>This information will be visible on the Digital Marketplace.</p>
+        <p>You can change it at any time.</p>
+      </div>
+    </div>
+  </div>
+
+  <form action="{{ url_for('.edit_what_buyers_will_see') }}" method="post" enctype="multipart/form-data">
```
Wynndow: You don't need the `enctype` attribute here, only if you're uploading a file. David linked me to a good article about it when he was reviewing the country picker. https://stackoverflow.com/questions/4526273/what-does-enctype-multipart-form-data-mean

tests/app/main/test_suppliers.py
```diff
@@ -743,9 +743,8 @@ def test_shows_supplier_info(self, data_api_client):
             ):
                 assert document.xpath("//*[normalize-space(string())=$t]", t=property_str), property_str
 
-            # Registration country and registration number not shown if Companies House ID exists
-            for property_str in ("Belize", "BEL153",):
-                assert property_str not in page_html
+            # Registration nnumber not shown if Companies House ID exists
```
Wynndow: Double n!

app/main/views/suppliers.py
CrystalPea: Good point with analytics, I will ask Ash. Do you think we should leave it as was?

app/templates/suppliers/details.html
CrystalPea: that's where it is now, right?

app/templates/suppliers/edit_what_buyers_will_see.html
CrystalPea: I asked @constancecerf about this and other one, will see what she says :). Thanks for spotting! <3

CrystalPea: So this should be "Save and continue", I'll change it in a minute

app/templates/suppliers/edit_what_buyers_will_see.html
CrystalPea: I asked @constancecerf and she said it should be "Return to company details", so I'll change that in a minute ☺️. Good spot!

app/templates/suppliers/edit_what_buyers_will_see.html
CrystalPea: Good point, I will ask @constancecerf

CrystalPea: Ok, I checked the content for the whole page with @constancecerf now, so hopefully it should be fine 🤞

app/templates/suppliers/edit_what_buyers_will_see.html
CrystalPea: Ok, I will get rid of it 👍

app/main/views/suppliers.py
CrystalPea: Ash says it is ok as long as we give him the new url. @TheDoubleK is it ok to change this route? If we do, do we need to do anything else?

app/main/views/suppliers.py
CrystalPea: ok 👍

app/templates/suppliers/edit_what_buyers_will_see.html
CrystalPea: ok 👍

app/main/views/suppliers.py
CrystalPea: I think we might, to avoid overwriting the data if the forms do not validate on POST request.

app/templates/suppliers/edit_what_buyers_will_see.html
CrystalPea: Hi @Wynndow we looked at it with @TheDoubleK and we decided against moving to macro at this moment for following reasons:
1. Since our validation uses two forms it might be a bit complex to implement.
2. We are short on time

I hope this is fine. Still, we changed the order of the errors, so summary validation shows after other validations, so it matches the order on the page.

app/main/views/suppliers.py
TheDoubleK: I think it should be fine... the functional tests just click through the journey I think, without explicitly checking the URLs.

app/templates/suppliers/details.html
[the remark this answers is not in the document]
CrystalPea: Should we add it to tech debt maybe and leave as-is for now?

PR #804, opened 2018-02-22
Accessibility fixes for summary table Edit links

katstevens:
Trello: https://trello.com/c/sYJYKnK9/58-how-summary-tables-handle-their-links-means-pages-can-contain-multiple-links-with-duplicate-text

Summary tables with multiple 'Edit' or 'Change' links are unhelpful for screenreader users who may not know which section the link refers to. 

v28.1.0 of the FE toolkit adds in a 'hidden_text' param to the `summary.top_link` macro (similar to the existing `summary.edit_link`) so we can make the link text less ambiguous for screen reader users. 

For example:
```
<a href="...">Edit</a>
```
becomes
```
<a href="...">Edit<span class="visually-hidden"> supplier details</span></a>
```

NB: I've guessed at the content for some of the links where there isn't an existing `section.name` to use. These will need a content designer to review (@constancecerf @karlchillmaid ?). 

See also: Brief Responses FE https://github.com/alphagov/digitalmarketplace-brief-responses-frontend/pull/32

### DIFF
--- a/app/templates/frameworks/contract_review.html
+++ b/app/templates/frameworks/contract_review.html
@@ -62,7 +62,7 @@
       {% call summary.field() %}
         <p>{{ agreement.signedAgreementDetails.signerName }}</p><p>{{ agreement.signedAgreementDetails.signerRole }}</p>
       {% endcall %}
-      {{ summary.edit_link("Edit", url_for('.signer_details', framework_slug=framework.slug, agreement_id=agreement.id)) }}
+      {{ summary.edit_link("Edit", url_for('.signer_details', framework_slug=framework.slug, agreement_id=agreement.id), hidden_text="person who signed") }}
     {% endcall %}
     {% call summary.row() %}
       {{ summary.field_name('Signature page') }}
@@ -71,7 +71,7 @@
       {% else %}
         {{ summary.field_name( "Uploaded {}".format(signature_page.last_modified|datetimeformat) if signature_page else None) }}
       {% endif %}
-      {{ summary.edit_link("Change", url_for('.signature_upload', framework_slug=framework.slug, agreement_id=agreement.id)) }}
+      {{ summary.edit_link("Change", url_for('.signature_upload', framework_slug=framework.slug, agreement_id=agreement.id), hidden_text="signature") }}
     {% endcall %}
   {% endcall %}
 </div>
--- a/app/templates/suppliers/company_summary.html
+++ b/app/templates/suppliers/company_summary.html
@@ -49,27 +49,27 @@
   {% call summary.row(complete=session.get("duns_number", None)) %}
   {{ summary.field_name("DUNS number") }}
   {{ summary.text(session.get("duns_number", "You must answer this question.")) }}
-  {{ summary.edit_link("Edit", url_for(".duns_number")) }}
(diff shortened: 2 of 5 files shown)


katstevens:
Crikey good spot @karlchillmaid - it might be some underlying logic as I think that part checks whether that section has already been completed (stored in the session), but I suspect it's just a typo. I'm not super familiar with the supplier flow - can any other devs shed light on what that should be? I'm happy to fix if need be.

app/templates/frameworks/contract_review.html
```diff
@@ -62,7 +62,7 @@
       {% call summary.field() %}
         <p>{{ agreement.signedAgreementDetails.signerName }}</p><p>{{ agreement.signedAgreementDetails.signerRole }}</p>
       {% endcall %}
-      {{ summary.edit_link("Edit", url_for('.signer_details', framework_slug=framework.slug, agreement_id=agreement.id)) }}
+      {{ summary.edit_link("Edit", url_for('.signer_details', framework_slug=framework.slug, agreement_id=agreement.id), hidden_text="Person who signed") }}
```
TheDoubleK: The hidden text probably doesn't capital letters at all, if it's going to end up being shown after the word "Edit" or "Change"?

app/templates/suppliers/company_summary.html
```diff
@@ -49,27 +49,27 @@
...
   {{ summary.field_name("DUNS number") }}
   {{ summary.text(session.get("duns_number", "You must answer this question.")) }}
-  {{ summary.edit_link("Edit", url_for(".duns_number")) }}
+  {{ summary.edit_link("Edit", url_for(".duns_number"), hidden_text="DUNS number") }}
   {% endcall %}
   {% call summary.row(complete=session.get("company_name", None)) %}
   {{ summary.field_name("Company name") }}
   {{ summary.text(session.get("company_name", "You must answer this question.")) }}
-  {{ summary.edit_link("Edit", url_for(".company_name")) }}
+  {{ summary.edit_link("Edit", url_for(".company_name"), hidden_text="Company name") }}
```
TheDoubleK: Same here for capital letters

app/templates/suppliers/company_summary.html
```diff
@@ -49,27 +49,27 @@
...
   {% endcall %}
   {% call summary.row(complete=session.get("contact_name", None)) %}
   {{ summary.field_name("Contact name") }}
   {{ summary.text(session.get("contact_name", "You must answer this question.")) }}
-  {{ summary.edit_link("Edit", url_for(".company_contact_details")) }}
+  {{ summary.edit_link("Edit", url_for(".company_contact_details"), hidden_text="Contact name") }}
   {% endcall %}
   {% call summary.row(complete=session.get("phone_number", None)) %}
```
TheDoubleK: Yeah I think you're right @karlchillmaid - this should be `email_address`.
Nice spot - that's been in there for over 3 years!

app/templates/frameworks/contract_review.html
karlchillmaid: Agree about the caps

katstevens:
OK, this should now be fixed up @TheDoubleK @karlchillmaid.

DilwoarH:
LGTM - probably needs a content person to approve 👍

PR #806, opened 2018-02-23
Update registered address page

Wynndow:
Some minor tweaks to the new registered address page, to bring it inline with how we're doing things on other pages. Mostly spurred on my working out why validation masthead links were broken in some cases.

### Remove JavaScript link from validation masthead
There was an issue with the usual links not working so we used
JavaScript instead. I've worked out what the problem was.

The links are actually powered by some other JavaScript from the
frontend toolkit. They look for an element with an ID that matches the
field name, and then target the first input or text area _within_ that.
So if the input itself has the id, or the id isn't used, the link won't
work.

### Abort on APIErrors when updating
As we're validating on the frontend, all errors should be due to
network issues or the like. So any APIErrors can bubble up as 500's.

### Use validation masthead from toolkit
By processing both forms errors in the view and passing them to the
template we can use validation masthead from the toolkit rather than
something custom.

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -119,6 +119,7 @@ def edit_registered_address():
     http_status = 200
     registered_address_form = EditRegisteredAddressForm()
     registered_country_form = EditRegisteredCountryForm()
+    form_errors = None
 
     if request.method == 'POST':
         address_valid = registered_address_form.validate_on_submit()
@@ -146,6 +147,18 @@ def edit_registered_address():
 
         http_status = 400
 
+        address_form_errors = [{
+            'question': registered_address_form[field].label.text,
+            'input_name': registered_address_form[field].name,
+        } for field in registered_address_form.errors.keys()]
+
+        country_form_errors = [{
+            'question': registered_country_form[field].label.text,
+            'input_name': registered_country_form[field].name,
+        } for field in registered_country_form.errors.keys()]
+
+        form_errors = address_form_errors + country_form_errors
+
     else:
         registered_address_form.address1.data = supplier['contact'].get('address1')
         registered_address_form.city.data = supplier['contact'].get('city')
@@ -159,6 +172,7 @@ def edit_registered_address():
         countries=COUNTRY_TUPLE,
         registered_address_form=registered_address_form,
         registered_country_form=registered_country_form,
+        form_errors=form_errors,
     ), http_status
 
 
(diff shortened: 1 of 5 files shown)


app/main/views/suppliers.py
```diff
@@ -139,13 +139,24 @@ def edit_registered_address():
...
-                error = e.message
+                abort(e.status_code)
 
-            else:
-                return redirect(url_for(".supplier_details"))
+            return redirect(url_for(".supplier_details"))
 
         http_status = 400
 
+        address_form_errors = [{
```
samuelhwilliams: Seems like this would be a nice little helper given how many wtforms we have here now. This repo is fine for now although it also seems like a good candidate for utils.

tests/app/main/test_suppliers.py
```diff
@@ -1285,13 +1285,10 @@ def test_handles_api_errors_when_updating_supplier_or_contact_information(self):
...
 
-        for error_message in ["I'm an error from the API", "So am I!"]:
+        for i in range(2):
             status, response = self.post_supplier_address_edit()
-            doc = html.fromstring(response)
 
-            assert status == 400
-            assert doc.xpath('normalize-space(//h1[@id="validation-masthead-heading"])') == error_message
-            assert "What is your registered office address?" in doc.xpath('//h1')[-1].text
+            assert status == 503
```
samuelhwilliams: I see what this test is doing but it's slightly confusing to read. I think it could benefit from either being separating out (less DRY) or parameterizing (maybe more complex at first glance). You're basically testing two things here and the setup for each is getting wrapped up with the other (e.g. if we remove the call to update_supplier from the code, the first iteration would still fail because it would silently catch the update_ci error).

app/main/views/suppliers.py
Wynndow: Yep, makes sense. Have moved out to a helper.
I've not implemented it in the other views however. Mostly because they're still being worked on by other people / the helper didn't actually add much / added more lines of code by having to add `form_errors` as a variable to the templates.
Let me know what you think; I was on the fence about doing it.

tests/app/main/test_suppliers.py
Wynndow: I see what you mean. I've tried it both ways, and separating into two tests is more clear I think. Parameterising it just added unnecessary complexity.

PR #808, opened 2018-02-23
Add one time vat page

Wynndow:
Trello ticket here: https://trello.com/c/NFRFA2K5

The tests won't go green until the [related PR](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/409) on the frontend-toolkit goes in. It's been approved, it's just waiting for Laurence to give it a thumbs up.

~~The validation messages may also be updated; Constance is reviewing them so they may change.~~
Constance has reviewed the validation messages.

### Add VatNumberForm
We need only validate the VAT number entered if the user has stated that
they are registered. We can't use the `Optional` validator for this in
case the user has started typing something in the input field and then
changed their mind.

The regex used for the VAT number covers 9 and 12 digit VAT codes, both
with the optional 'GB' prefix. It also supports the lesser used but
still valid GD and HA vat numbers which are used for government
departments and health authorities. Again, with the optional 'GB'
prefix.

### Pull in verision 28.2.0 of the frontend toolkit
It includes patterns for conditionally reveling inputs with
radiobuttons.

### Screen capture:

new_vat

### DIFF
--- a/app/main/forms/suppliers.py
+++ b/app/main/forms/suppliers.py
@@ -1,3 +1,4 @@
+from re import IGNORECASE
 from flask_wtf import Form
 from wtforms import RadioField
 from wtforms.validators import AnyOf, InputRequired, Length, Optional, Regexp, StopValidation, ValidationError
@@ -218,11 +219,12 @@ class VatNumberForm(Form):
             raise StopValidation()
 
     vat_registered = RadioField("VAT registered",
-                                validators=[InputRequired(message="You must provide an answer.")],
+                                validators=[InputRequired(message="You need to answer this question.")],
                                 choices=[('Yes', 'Yes'), ('No', 'No')])
     vat_number = StripWhitespaceStringField('VAT number', default="", validators=[
         stop_validation_if_not_registered,
         InputRequired(message="You must provide a VAT number."),
-        Regexp(r'^([GB])*(([1-9]\d{8})|([1-9]\d{11})|((GD|HA)[1-9]\d{2}))$',
-               message="You must provide a valid VAT number.")
+        Regexp(r'^([GB])*((\d{9})|(\d{12})|((GD|HA)\d{3}))$',
+               flags=IGNORECASE,
+               message="You must provide a valid VAT number, they are usually either 9 or 12 digits.")
     ])
--- a/app/main/forms/suppliers.py
+++ b/app/main/forms/suppliers.py
@@ -1,6 +1,6 @@
 from flask_wtf import Form
 from wtforms import RadioField
-from wtforms.validators import AnyOf, InputRequired, Length, Optional, Regexp, ValidationError
+from wtforms.validators import AnyOf, InputRequired, Length, Optional, Regexp, StopValidation, ValidationError
 
 from dmutils.forms import StripWhitespaceStringField, EmailField, EmailValidator
 from ..helpers.suppliers import COUNTRY_TUPLE
@@ -207,3 +207,22 @@ class CompanyTradingStatusForm(Form):
     trading_status = RadioField('Trading status',
                                 validators=[InputRequired(message="You must choose a trading status.")],
                                 choices=[(o['value'], o['label']) for o in OPTIONS])
+
+
+class VatNumberForm(Form):
+    def stop_validation_if_not_registered(form, field):
+        # If a user is not registered for VAT we don't care about validating something they have entered in the
+        # VAT number input field. If a value for `vat_registered` is not sent in the POST request, WTForms sets the
+        # value to `None`, hence its inclusion here
+        if form.vat_registered.data in ('No', 'None'):
+            raise StopValidation()
+
+    vat_registered = RadioField("VAT registered",
+                                validators=[InputRequired(message="You must provide an answer.")],
+                                choices=[('Yes', 'Yes'), ('No', 'No')])
+    vat_number = StripWhitespaceStringField('VAT number', default="", validators=[
+        stop_validation_if_not_registered,
+        InputRequired(message="You must provide a VAT number."),
+        Regexp(r'^([GB])*(([1-9]\d{8})|([1-9]\d{11})|((GD|HA)[1-9]\d{2}))$',
+               message="You must provide a valid VAT number.")
+    ])
--- a/app/main/forms/suppliers.py
+++ b/app/main/forms/suppliers.py
@@ -1,6 +1,6 @@
-from re import IGNORECASE
+import re
 from flask_wtf import Form
-from wtforms import RadioField
+from wtforms import RadioField, StringField
 from wtforms.validators import AnyOf, InputRequired, Length, Optional, Regexp, StopValidation, ValidationError
 
 from dmutils.forms import StripWhitespaceStringField, EmailField, EmailValidator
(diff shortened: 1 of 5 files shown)


Wynndow:
@karlchillmaid Hey Karl, would you be able to sign off on the content on this please? Thanks!

app/main/forms/suppliers.py
```diff
@@ -139,3 +140,23 @@ class CompanyOrganisationSizeForm(Form):
...
+            raise StopValidation()
+
+    vat_registered = RadioField("VAT registered",
+                                validators=[InputRequired(message="You need to answer this question.")],
+                                choices=[('Yes', 'Yes'), ('No', 'No')])
+    vat_number = StripWhitespaceStringField('VAT number', default="", validators=[
+        stop_validation_if_not_registered,
+        InputRequired(message="You must provide a VAT number."),
+        Regexp(r'^([GB])*((\d{9})|(\d{12})|((GD|HA)\d{3}))$',
+               flags=IGNORECASE,
```
CrystalPea: should we make all numbers either uppercase or lowercase before we save them into the database?

app/main/views/suppliers.py
```diff
@@ -263,6 +264,56 @@ def edit_supplier_organisation_size():
...
+    if supplier.get("vatNumber"):
+        return (
+            render_template("suppliers/already_completed.html", completed_data_description="vat number"),
+            200 if request.method == 'GET' else 400
+        )
+
+    form_errors = None
+    if request.method == 'POST':
+        if form.validate_on_submit():
+            maybe_vat_number = form.vat_number.data if form.vat_registered.data == 'Yes' else 'Not VAT registered'
```
CrystalPea: Isn't `maybe_vat_number` an odd variable name? Maybe we could rename it? Something like `vat_number_info` potentially? What do you think?

app/main/views/suppliers.py
```diff
@@ -263,6 +264,56 @@ def edit_supplier_organisation_size():
...
+            try:
+                data_api_client.update_supplier(supplier_id=current_user.supplier_id,
+                                                supplier={"vatNumber": maybe_vat_number},
+                                                user=current_user.email_address)
+            except APIError as e:
+                abort(e.status_code)
+
+            return redirect(url_for('.supplier_details'))
+
+        current_app.logger.warning(
```
CrystalPea: What is this logging about?

app/main/views/suppliers.py
```diff
@@ -263,6 +264,56 @@ def edit_supplier_organisation_size():
...
+def edit_supplier_vat_number():
+    form = VatNumberForm()
+    try:
+        supplier = data_api_client.get_supplier(current_user.supplier_id)['suppliers']
+    except APIError as e:
+        abort(e.status_code)
+
+    if supplier.get("vatNumber"):
+        return (
+            render_template("suppliers/already_completed.html", completed_data_description="vat number"),
```
CrystalPea: Do we want to capitalise 'VAT', like so?

`completed_data_description="VAT number"`

tests/app/main/test_suppliers.py
```diff
@@ -80,7 +80,9 @@ def get_supplier(*args, **kwargs):
         }
     }
     supplier_info.update(kwargs)
-    return {"suppliers": supplier_info}
+    return {"suppliers": dict(
+        filter(lambda x: x[1] is not None, supplier_info.items())
```
CrystalPea: what does this filtering do? :)

tests/app/main/test_suppliers.py
```diff
@@ -2282,3 +2284,152 @@ def test_existing_org_size_sets_current_selection(self, data_api_client, existin
...
+            assert doc.xpath("//h1[normalize-space(string())='Change your vat number']")
+            assert not self.data_api_client.update_supplier.called
+
+    def test_posting_valid_vat_code_updates_supplier(self):
+        with self.app.test_client():
+            res = self.client.post(
+                "/suppliers/vat-number/edit",
+                data={
+                    "vat_registered": "Yes",
+                    "vat_number": "GB123456789",
```
CrystalPea: Since you made this beautiful regex checking for several types of VAT numbers, both with and without prefixes, would it be worth it to parametrise this test and check that they all work? :)

CrystalPea: Never mind, I see you did it below, well done!

tests/app/main/test_suppliers.py
```diff
@@ -2282,3 +2284,152 @@ def test_existing_org_size_sets_current_selection(self, data_api_client, existin
...
+                assert not self.data_api_client.update_supplier.called
+            else:
+                assert res.status_code == 302
+                self.data_api_client.update_supplier.assert_called_once_with(
+                    supplier_id=1234,
+                    supplier={"vatNumber": vat_number},
+                    user="email@email.com",
+                )
+
+    def test_validation_of_vat_registered(self):
```
CrystalPea: the name of this test could be more specific, like `test_validation_against_empty_string_for_vat_registered_suppliers`

tests/app/main/test_suppliers.py
samuelhwilliams: Given a dictionary such as `{'a': True, 'b': False, 'c': None}` it will filter out any key/value pairs where the value is not None, i.e. you'll be left with ``{'a': True, 'b': False}`.

`supplier_info.items()` groups the key/value pairs into tuples like so: `[('a', True), ('b', False), ('c', None)]`. The filter then takes each tuple and checks the second item (and index 1) against `is not None`.

CrystalPea: Ahh, makes a lot of sense,  thanks a lot! :)

app/main/views/suppliers.py
TheDoubleK: Yep I think so, but check with @constancecerf

app/main/forms/suppliers.py
Wynndow: Good call. I've updated to uppercase everything.

app/main/views/suppliers.py
Wynndow: Yeah it's a bit weird. I know we've used similar before though in the search-api. Happy to change it here.

app/main/views/suppliers.py
Wynndow: It's just giving handy feedback on why something might potentially fail. It seems to be a pattern we do for supplier account views, just continuing with it really.

app/main/views/suppliers.py
Wynndow: Yeah, good spot.

tests/app/main/test_suppliers.py
Wynndow: The reason I did this is because when the API serialises a supplier, it purges all keys with a null value.

Adding this filter in should make our tests more realistic

app/main/views/suppliers.py
CrystalPea: Ahh, ok, thank you ^_^

tests/app/main/test_suppliers.py
CrystalPea: Ahh, smart move! :)

tests/app/main/test_suppliers.py
```diff
@@ -2322,3 +2324,188 @@ def test_post_shows_already_entered_page_and_api_not_called_if_data_already_ente
...
+                data={
+                    "vat_registered": "Yes",
+                    "vat_number": vat_number,
+                }
+            )
+            doc = html.fromstring(res.get_data(as_text=True))
+
+            if message:
+                assert res.status_code == 400
+                assert doc.xpath(
```
samuelhwilliams: We have a helper method to validate mastheads now, `assert_single_question_page_validation_errors`. It would be good to use that here.

tests/app/main/test_suppliers.py
```diff
@@ -2322,3 +2324,188 @@ def test_post_shows_already_entered_page_and_api_not_called_if_data_already_ente
...
+                data={
+                    "vat_number": "",
+                }
+            )
+            doc = html.fromstring(res.get_data(as_text=True))
+
+            assert res.status_code == 400
+            assert doc.xpath(
+                "//div[@class='validation-masthead'][normalize-space(string())=$masthead]",
+                masthead=f"There was a problem with your answer to: VAT registered"
```
samuelhwilliams: Doesn't need to be an f-string here.

tests/app/main/test_suppliers.py
```diff
@@ -2322,3 +2324,188 @@ def test_post_shows_already_entered_page_and_api_not_called_if_data_already_ente
...
+    @pytest.mark.parametrize("vat_number, message, masthead", (
+        ("GB123456789", None, None),
+        ("GB123456789012", None, None),
+        ("GBGD123", None, None),
+        ("GBHA456", None, None),
+        ("123456789", None, None),
+        ("123456789012", None, None),
+        ("GD123", None, None),
+        ("HA456", None, None),
+        ("NOTVALID", "You must provide a valid VAT number, they are usually either 9 or 12 digits.", "VAT number"),
```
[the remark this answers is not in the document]
Wynndow: I checked this with @constancecerf, happy to revisit though?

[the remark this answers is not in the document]
karlchillmaid: FWIW this looks fine to me.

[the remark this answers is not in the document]
samuelhwilliams: 🤷‍♂️ Alright.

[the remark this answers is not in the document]
karlchillmaid: Style guide says: Don’t use semicolons as they are often mis-read. Long sentences using semicolons should be broken up into separate sentences instead.

I think it's right to question the comma here, but breaking this into 2 sentences feels unnecessary. Thanks for your content vigilance! 🙂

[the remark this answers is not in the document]
Wynndow: @karlchillmaid Constance has suggested using a dash instead of a comma. Would that be ok?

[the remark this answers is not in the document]
karlchillmaid: Yep, a dash will do nicely.

PR #813, opened 2018-02-28
Fix links on "What buyers will see" validation banner

CrystalPea:
Also make "Optional" only appear when the summary is empty.

### DIFF
--- a/app/templates/suppliers/details.html
+++ b/app/templates/suppliers/details.html
@@ -59,7 +59,9 @@
     {{ summary.field_name('Summary') }}
     {{ summary.text(supplier.description) }}
     {% call summary.field(action=True) %}
-      Optional
+      {% if not supplier.description %}
+        Optional
+      {% endif %}
     {% endcall %}
   {% endcall %}
 {% endcall %}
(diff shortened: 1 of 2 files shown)


app/templates/suppliers/details.html
```diff
@@ -58,9 +58,11 @@
...
     {{ summary.field_name('Summary') }}
     {{ summary.text(supplier.description) }}
-    {% call summary.field(action=True) %}
-      Optional
-    {% endcall %}
+    {%  if not supplier.description %}
+      {% call summary.field(action=True) %}
+        Optional
+      {% endcall %}
+    {% endif %}
```
samuelhwilliams: I think we need to still generate an empty cell here or we'll get a line that terminates early.

CrystalPea: The line doesn't end early now, but it might if we had another "Optional" field. Do you think I should add empty field for consistency?


screen shot 2018-02-28 at 15 14 28

samuelhwilliams: Yep! Could catch someone out if we try to add another row in the future. Good to prevent that if we can.

PR #816, opened 2018-03-01
Add your company registration number page

TheDoubleK:
For this ticket: https://trello.com/c/nToxBmYI/48-new-one-time-page-build-registered-with-companies-house-page

## The page
screen shot 2018-03-01 at 16 41 32

## The reveal
screen shot 2018-03-01 at 16 42 16

screen shot 2018-03-01 at 16 42 24

## Error messaging
screen shot 2018-03-01 at 16 42 58

screen shot 2018-03-01 at 16 42 46

screen shot 2018-03-01 at 16 42 38


app/main/forms/suppliers.py
```diff
@@ -66,20 +66,41 @@ class AddCompanyRegisteredNameForm(Form):
...
-               )
-    ])
+class AddCompanyRegistrationNumberForm(Form):
+    has_companies_house_number = RadioField(
+        "Are you registered with Companies House?",
+        validators=[InputRequired(message="You need to answer this question.")],
+        choices=[('Yes', 'Yes'), ('No', 'No')]
+    )
+    companies_house_number = StripWhitespaceStringField(
+        'Companies house number',
```
samuelhwilliams: Should this be `House`?

app/templates/suppliers/edit_company_registration_number.html
```diff
@@ -0,0 +1,89 @@
...
+  {% endwith %}
+{% endblock %}
+
+{% block main_content %}
+
+<div class="single-question-page">
+  <div class="grid-row">
+    <div class="column-one-whole">
+      {% if form.errors %}
+        {% for field in form.errors %}
```
samuelhwilliams: This feels weird because it looks like you could end up with multiple validation boxes at the top of the page. It would be nice if we don't have a for loop around this as we anticipate only ever displaying one (and it would be weirdly broken to show more).

I think it should be a list comprehension inside the `{% with ... %}` block.

app/main/forms/suppliers.py
TheDoubleK: I wasn't sure... but maybe

app/templates/suppliers/edit_company_registration_number.html
TheDoubleK: You can't do list comprehensions in Jinja, but I'll see if I can find a nicer way.

samuelhwilliams: Oh wow, yeah, totally forgot that. Maybe can build the structure Python side. 🤷‍♂️

app/main/forms/suppliers.py
TheDoubleK: Of course it should be, fixing now.

laurenceberry:
Looking good to me 👍

PR #818, opened 2018-03-05
Remove registration date

samuelhwilliams:
## Summary
Registration date is a field we are no longer interested in, so we should remove all references from it as we will be dropping it from the API shortly.

## Ticket
https://trello.com/c/M2jPJ3kZ/58-purge-registration-date-from-logic-and-data


samuelhwilliams:
Good spot 👍

PR #821, opened 2018-03-05
Add links to all the new detail edit pages

CrystalPea:
Trello ticket: https://trello.com/c/x5mzD7bS/49-supplier-add-links-to-all-the-new-change-pages

screen shot 2018-03-05 at 16 31 09

### DIFF
--- a/app/templates/suppliers/details.html
+++ b/app/templates/suppliers/details.html
@@ -80,9 +80,9 @@
   {{ summary.field_name('Registered company name') }}
   {% if supplier.registeredName %}
     {{ summary.text(supplier.registeredName) }}
-    {{ summary.edit_link(label="Change", link=url_for(".edit_supplier_registered_name")) }}
+    {{ summary.edit_link(label="Change", link=url_for(".edit_supplier_registered_name"), hidden_text="Registered company name") }}
   {% else %}
-    {{ summary.link(link_title="Answer required", link=url_for(".edit_supplier_registered_name")) }}
+    {{ summary.link(link_title="Answer required", link=url_for(".edit_supplier_registered_name"), hidden_text="for registered company name") }}
     {{ summary.text("") }}
   {% endif %}
 {% endcall %}
@@ -106,9 +106,9 @@
         {% include "toolkit/contact-details.html" %}
       {% endwith %}
     {% endcall %}
-    {{ summary.edit_link(label="Change", link=url_for(".edit_registered_address")) }}
+    {{ summary.edit_link(label="Change", link=url_for(".edit_registered_address"), hidden_text="Registered company address") }}
   {% else %}
-    {{ summary.link(link_title="Answer required", link=url_for(".edit_registered_address")) }}
+    {{ summary.link(link_title="Answer required", link=url_for(".edit_registered_address"), hidden_text=" for registered company address") }}
     {{ summary.text("") }}
   {% endif %}
 {% endcall %}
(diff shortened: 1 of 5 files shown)


app/templates/suppliers/details.html
```diff
@@ -71,23 +71,28 @@
...
-  {% call summary.row() %}
-    {{ summary.field_name('Registered company name') }}
-      {{ summary.text(supplier.registeredName) }}
-  {% endcall %}
-{% endif %}
+{% call summary.row() %}
+  {{ summary.field_name('Registered company name') }}
+  {% if supplier.registeredName %}
+    {{ summary.text(supplier.registeredName) }}
+    {{ summary.edit_link(label="Change", link=url_for(".edit_supplier_registered_name")) }}
```
katstevens: Can we use the hidden_text field for these `edit_links` to provide context for screen reader users? For example:

```
summary.edit_link(label="Change", link=url_for(".edit_supplier_registered_name"), hidden_text="Registered company name")
```

See here for more context: https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/804

app/templates/suppliers/details.html
```diff
@@ -101,48 +106,65 @@
...
-      {{ summary.text(supplier.companiesHouseNumber) }}
-    {% endcall %}
-  {% endif %}
-
-  {% if supplier.otherCompanyRegistrationNumber and not supplier.companiesHouseNumber and country_name != 'United Kingdom' %}
-    {% call summary.row() %}
-      {{ summary.field_name('Registration number') }}
-      {{ summary.text(supplier.otherCompanyRegistrationNumber) }}
-    {% endcall %}
+    {{ summary.edit_link(label="Change", link=url_for(".edit_registered_address")) }}
```
katstevens: Same here for adding `hidden_text="Registration number"`

laurenceberry:
Looks good from my perspective.

app/templates/suppliers/details.html
CrystalPea: Yes, great idea :)

samuelhwilliams:
Took me way too long to spot this, but your `hidden_text` attributes are nested inside the `url_for` call, so you're generating weird URLs rather than adding the hidden text.

Also I don't think we want to have the verb (e.g. 'change', 'add') included as part of the hidden text, as a screenreader will read the visible text followed by the hidden text, which will end up with e.g. 'Change change registered company name'. This ignores our internationalisation issue for now, but is consistent with other uses of the hidden_text attribute (e.g. as in @katstevens's example).

CrystalPea:
@samuelhwilliams good spot with the hidden url, I probably should not have added those just before going home 😅 . And good point with the verbs, I will get rid of them.

Also, do you think we need hidden_text attributes on both "Change" and "Answer Required" links?
I am asking as Answer Required link is just next to the row title...

katstevens:
The `hidden_text` is needed where there's potential ambiguity. So if there are multiple 'Change' links on the page, the screen reader needs to be able to differentiate between them.

CrystalPea:
@katstevens good idea, I will check with @constancecerf 

As to the change from "or" to "and": yes, it's intentional. In earlier iteration of the page we did not have "Answer required" link so it made sense to show any information that is there. Now, all this information will need to be submitted for a supplier to be able to proceed with their framework application, so if any is missing, an "Answer required" link should be shown.

TheDoubleK:
The links have moved out of line:
screen shot 2018-03-06 at 17 51 24

CrystalPea:
@TheDoubleK thanks for spotting, fixed :) (somehow I am least cautious when I am introducing changes as a part of code review, I need to work on that 😬 )
screen shot 2018-03-07 at 15 38 34

CrystalPea:
@TheDoubleK good suggestion, I changed the copy of these hidden_text attributes now 👍. Merging when Travis is done ☺️

PR #825, opened 2018-03-06
Set GA custom dimension for supplier organisation size

katstevens:
Trello: https://trello.com/c/PZyZdMtG/337-custom-dimension-to-capture-size-of-supplier

- Adds in extra context for the custom dimensions template to set the organisation size for a supplier.
- Updates test helpers to include the supplier organisation size in the test user JSON.

### DIFF
--- a/tests/app/main/test_suppliers.py
+++ b/tests/app/main/test_suppliers.py
@@ -955,6 +955,33 @@ class TestSupplierDashboardLogin(BaseApplicationTest):
             assert len(doc.xpath('//meta[@data-value="supplier"]')) == 1
             assert len(doc.xpath('//meta[@data-value="small"]')) == 1
 
+    @mock.patch("app.main.views.suppliers.data_api_client")
+    @mock.patch("app.main.views.suppliers.get_current_suppliers_users")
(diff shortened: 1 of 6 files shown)


katstevens:
Figured out the issue with the tests - the `test-utils` example API response had the wrong key name.

tests/app/main/test_suppliers.py
```diff
@@ -936,6 +937,24 @@ def test_should_redirect_to_login_if_not_logged_in(self):
...
+
+        with self.app.test_client():
+            self.login()
+            res = self.client.get('/suppliers')
+
+            doc = html.fromstring(res.get_data(as_text=True))
+            # The default supplier details from self.login() should available as attributes of current_user
+            assert len(doc.xpath('//meta[@data-value="supplier"]')) == 1
+            assert len(doc.xpath('//meta[@data-value="small"]')) == 1
+
```
brenetic: Might be useful to also test that when a user is not logged in that organisation size isn't set.

katstevens: Now done - also added a test for logged in suppliers without an organisation size.

PR #828, opened 2018-03-09
 Show a back to framework application link if framework application info in session

CrystalPea:
Trello ticket: https://trello.com/c/l2xh4cR3/63-confirm-supplier-details-4-return-to-your-framework-application-link

The link and hint text:
screen shot 2018-03-09 at 16 49 03

### DIFF
--- a/app/templates/suppliers/details.html
+++ b/app/templates/suppliers/details.html
@@ -169,4 +169,17 @@
   {% endif %}
 
 {% endcall %}
+
+{% if currently_applying_to %}
+  <p>You must complete your company details to make an application.</p>
+  {%
+    with
+    url = url_for(".framework_dashboard", framework_slug=currently_applying_to.slug),
+    text = "Return to your {} application".format(currently_applying_to.name),
+    bigger = false
+  %}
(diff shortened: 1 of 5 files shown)


laurenceberry:
Looks good to me

app/templates/suppliers/details.html
```diff
@@ -169,4 +169,17 @@
   {% endif %}
 
 {% endcall %}
+
+{% if currently_applying_to %}
+  You must complete your company details to make an application.
```
Wynndow: Does this line need to be here? I see that it's in the screenshot from the prototype but I'm not sure it's part of this ticket? It feels weird to conditionally display it depending on the framework status. Worth checking with @laurenceberry?

Other than that, this looks good to me!

constancecerf: This is fine for now but it's a sentence that will change once we build the green confirm button.

CrystalPea:
@Wynndow the content has been confirmed as correct, are you ok to approve or any changes needed? ☺️

Wynndow:
👍

PR #832, opened 2018-03-13
add banner to account page

DilwoarH:
PRE REQUISITE:  https://github.com/alphagov/digitalmarketplace-user-frontend/pull/43

Add the user research banner to the account page using the existing GOV.uk pattern. Allow for links to 'sign up to mailing list' page

As a user I should be presented with the banner to sign up to the user research mailing list when I log in if: I have not already closed it and I have not already signed up for this.
This is the first point of contact for getting users to sign up to the mailing list.

The close link should set a cookie value of 'seen_this_banner_thing' to true so we don't show it again
The sign up link should link to the new sign up page.

https://trello.com/c/OJ9lFk1M/129-add-banner-to-account-page

screen shot 2018-03-13 at 14 43 21


DilwoarH:
@benvand - additional tests added

PR #833, opened 2018-03-13
Allow editing until confirmation

samuelhwilliams:
## Summary
Allow suppliers to edit some company details (vat number, registration
number, registration name) until they specifically confirm they are all
correct. In the theoretically impossible situation where a supplier has
confirmed their details but one/more of these pieces of information are
missing, they will still be able to provide it as a one-off.

## Related PRs
* https://github.com/alphagov/digitalmarketplace-api/pull/762

## Example: Company details not confirmed
unconfirmed-account

## Example: Company details confirmed
confirmed-account

## Ticket
https://trello.com/c/NaNjMOHc/61-confirm-supplier-details-2-frontend-logic-changes

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -174,7 +174,7 @@ def edit_registered_address():
 def edit_supplier_registered_name():
     form = AddCompanyRegisteredNameForm()
     supplier = data_api_client.get_supplier(current_user.supplier_id)['suppliers']
-    if supplier.get("registeredName"):
+    if supplier.get("registeredName") and supplier.get('companyDetailsConfirmed'):
         return (
             render_template("suppliers/already_completed.html", completed_data_description="registered company name"),
             200 if request.method == 'GET' else 400
@@ -199,6 +199,9 @@ def edit_supplier_registered_name():
 
             return render_template("suppliers/edit_registered_name.html", form=form), 400
 
+    else:
+        form.registered_company_name.data = supplier.get("registeredName")
+
     return render_template('suppliers/edit_registered_name.html', form=form)
 
 
@@ -207,7 +210,8 @@ def edit_supplier_registered_name():
 def edit_supplier_registration_number():
     form = AddCompanyRegistrationNumberForm()
     supplier = data_api_client.get_supplier(current_user.supplier_id)['suppliers']
-    if supplier.get("companiesHouseNumber") or supplier.get("otherCompanyRegistrationNumber"):
+    if (supplier.get("companiesHouseNumber") or supplier.get("otherCompanyRegistrationNumber")) \
+            and supplier.get('companyDetailsConfirmed'):
         return (
             render_template(
                 "suppliers/already_completed.html",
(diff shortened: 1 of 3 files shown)


tests/app/main/test_suppliers.py
```diff
@@ -2662,7 +2682,26 @@ def test_fails_if_api_update_fails(self, data_api_client):
             res = self.client.post("/suppliers/registration-number/edit", data=valid_post_data)
             assert res.status_code == 504
 
-    def test_get_shows_already_entered_page_and_api_not_called_update_if_data_already_entered(self, data_api_client):
+    @pytest.mark.parametrize('overwrite_supplier_data',
+                             (
+                                 {'companyDetailsConfirmed': False},
+                                 {'companiesHouseNumber': None, 'otherCompanyRegistrationNumber': None},
+                             ))
+    def test_get_shows_form_on_page_if_supplier_data_not_complete_and_confirmed(self, data_api_client,
```
CrystalPea: Are test on line 2441 and this test essentially the same test for two different pages? If yes, maybe it would be good to combine them / do a more robust parametrise to test behaviour of all affected pages? What do you think?

app/main/views/suppliers.py
```diff
@@ -222,13 +226,15 @@ def edit_supplier_registration_number():
                 if form.has_companies_house_number.data == "Yes":
                     data_api_client.update_supplier(
                         supplier_id=current_user.supplier_id,
-                        supplier={"companiesHouseNumber": form.companies_house_number.data.upper()},
+                        supplier={"companiesHouseNumber": form.companies_house_number.data.upper(),
+                                  "otherCompanyRegistrationNumber": None},
```
Wynndow: OK, this does break it. Using `None` here kicks and API error.
`dmapiclient.errors.HTTPError: JSON was not a valid format. None is not of type 'string' (status: 400)`

Think you need to give it empty string.

Wynndow:
I've noticed that suppliers can't edit their DUNS number. Is this because we collect it during supplier creation?

app/main/views/suppliers.py
samuelhwilliams: This is what the corresponding api pr is for - to allow setting back to none/null. You'll need to check that out as well to test.

samuelhwilliams:
Yep, duns number is basically our PK for suppliers (I think...)  so they can't change it after signing up.

CrystalPea:
@Wynndow That's correct re: DUNS number. It is the way CCS identifies the suppliers I think

tests/app/main/test_suppliers.py
samuelhwilliams: They are basically the same yeah. I decided not to combine them/parameterize them so that the test code stays close to the rest of the context for the page. There might be more I could do. I'll take another look and see if there's any sensible options to de-duplicate.

CrystalPea: I see, yeah, it does make sense, but then shouldn't we have one for each page?

app/main/views/suppliers.py
Wynndow: aha!

Wynndow: Yep, totally works with the api branch

tests/app/main/test_suppliers.py
samuelhwilliams: We do have one of these tests for each of the three pages I think?

PR #834, opened 2018-03-13
Change how open framework banner is displayed on dashboard

CrystalPea:
Trello ticket: https://trello.com/c/e1Gk2eTF/12-fix-styling-for-framework-banners-on-account

Everything done as per the screenshot I think except the headers are `h3`, not `h2`,  as changing them would add a layer of complexity (The header element lives in the macro and is used for many different temporary messages across digital marketplace).

### Two started applications:
screen shot 2018-03-13 at 16 34 10

____

### Started and not started application:
screen shot 2018-03-13 at 16 34 38

____

### Mobile view:
screen shot 2018-03-13 at 16 37 03

____

### One framework open:
screen shot 2018-03-13 at 17 29 44

### DIFF
--- a/app/templates/suppliers/_frameworks_open.html
+++ b/app/templates/suppliers/_frameworks_open.html
@@ -1,41 +1,31 @@
- <div class='grid-row'> 
-  <div class='column-two-thirds'>
+
     {% for framework in frameworks.open %}
       {% if framework.registered_interest %}
-        {%
-          with
-          items = [{
-            "title": "Continue your %s application" | format(framework.name),
-            "link": url_for(".framework_dashboard", framework_slug=framework.slug),
-            "body": framework.deadline,
-          }]
+        {% with
+          headingId = "closed-notice",
+          grid = "column-one-third",
+          main = true,
+          messages = [framework.deadline],
+          link = { "label": "Continue your application", "url": url_for(".framework_dashboard", framework_slug=framework.slug) },
+          heading = "Your {} application".format(framework.name)
         %}
-          {% include "toolkit/browse-list.html" %}
+          {% include "toolkit/temporary-message.html" %}
         {% endwith %}
       {% else %}
         <form action="{{ url_for('.framework_dashboard', framework_slug=framework.slug) }}" method="POST">
-            <div class="summary-item-lede">
-              <h2 class="summary-item-heading">
-                Apply to {{ framework.name }} 
-              </h2>
-              <p>
-                Deadline: {{ framework.dates.framework_close_date }}
-              </p>
-              {%
-              with
-                  type = "save",
-                  label = "Start application"
-              %}
-                  {% include "toolkit/button.html" %}
-              {% endwith %}
-
-              <p>
-                Starting your application means you’ll receive <br>{{ framework.name }} email updates.
-              </p>
-            </div>
-          <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
+          <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
+          {% with
+            headingId = "closed-notice",
+            grid = "column-one-third",
+            main = true,
+            messages = [
+              "Starting your application means you’ll receive {} email updates.".format(framework.name)
+            ],
+            button = {"label": "Apply", "type": "save"},
+            heading = "{} is open for applications".format(framework.name)
+          %}
+            {% include "toolkit/temporary-message.html" %}
+          {% endwith %}
         </form>
       {% endif %}
     {% endfor %}
-  </div>
-</div>
--- a/app/templates/suppliers/_frameworks_open.html
+++ b/app/templates/suppliers/_frameworks_open.html
@@ -3,6 +3,7 @@
       {% if framework.registered_interest %}
         {% with
           main = true,
+          header_level = "h2",
           messages = [framework.deadline],
           link = { "label": "Continue your application", "url": url_for(".framework_dashboard", framework_slug=framework.slug) },
           heading = "Your {} application".format(framework.name)
@@ -14,6 +15,7 @@
           <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
           {% with
             main = true,
+            header_level = "h2",
             messages = [
               "Starting your application means you’ll receive {} email updates.".format(framework.name)
             ],
--- a/app/templates/suppliers/_frameworks_open.html
+++ b/app/templates/suppliers/_frameworks_open.html
@@ -2,8 +2,6 @@
     {% for framework in frameworks.open %}
       {% if framework.registered_interest %}
         {% with
-          headingId = "closed-notice",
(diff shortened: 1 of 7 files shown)


app/templates/suppliers/_frameworks_open.html
```diff
@@ -1,41 +1,30 @@
...
-              %}
-                  {% include "toolkit/button.html" %}
-              {% endwith %}
-
-              <p>
-                Starting your application means you’ll receive <br>{{ framework.name }} email updates.
-              </p>
-            </div>
-          <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
+          {% with
```
Wynndow: Need to add the csrf token here. I think we found it yesterday but never commited it. `<input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>`

CrystalPea: Ahh, that's right, doing that now, good spot :)

laurenceberry:
This looks good to me.

samuelhwilliams:
Looks good @CrystalPea, although I think it would be worth putting some extra work into this page to get the headings at the right level in the hierarchy. At the moment it's a bit confusing. The main heading is h1, then have three h3's down the left-hand side that visually look very different from each other, and then on the right we have a single h2 that looks the same as the bottom h3. This doesn't feel very accessible and I suspect it would give a very confusing experience to people using e.g. a screenreader.

I think the best thing to do is have all of the current h3's become h2's. I'd argue that the blue box headers should be the same font size as the other subheadings on the page, although that's more one for @laurenceberry.

Here's what it would look like with equal font sizing:
screen shot 2018-03-15 at 08 39 24

app/templates/suppliers/_frameworks_open.html
```diff
@@ -1,41 +1,31 @@
...
       {% if framework.registered_interest %}
-        {%
-          with
-          items = [{
-            "title": "Continue your %s application" | format(framework.name),
-            "link": url_for(".framework_dashboard", framework_slug=framework.slug),
-            "body": framework.deadline,
-          }]
+        {% with
+          headingId = "closed-notice",
```
samuelhwilliams: I can't work out what this line does?

TheDoubleK:
If we're going to change the heading size then perhaps this needs a new macro altogether in the toolkit for this, to avoid messing with how these look elsewhere.  There are already a number of variations on these "temporary messages" and making all the things infinitely configurable takes us away from the idea of fixed building block patterns.

I do quite like the look of Sam's mock-up though.

And 100% agree about somehow getting the heading levels right.

app/templates/suppliers/_frameworks_open.html
TheDoubleK: This looks to me like an old thing that is no longer used in the toolkit anywhere, but that still exists in the example yml files.  I can't see that it ever gets processed anywhere or does anything.

You'd expect it to be included by the macro as the `id` of the heading element generated.  But I'm 99% sure it doesn't actually ever get read.

TheDoubleK: So... I've done a bit of digging.

Those lines were added here: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/212/files
Along with code in the temporary-message template that aded the ID to the heading in the macro.

But then it got removed from the template here: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/227

But the examples in the yml files didn't have the `headingId` removed from them.

So we can safely delete `headingId` from all temporary message includes here, and also remove them from the examples in the toolkit.

app/templates/suppliers/_frameworks_open.html
```diff
@@ -1,41 +1,31 @@
- <div class='grid-row'>
```
TheDoubleK: I think we still probably want these in a grid row and column?  Even if it's a `column-one-whole`?  Or are these partials already inside a grid row element when they get included?

app/templates/suppliers/_frameworks_open.html
CrystalPea: I checked and I am afraid it does not do anything at all. I took it wholesale from pages builder.  I will get rid of it. I will also talk to Chris about taking the line out of pages builder.

CrystalPea: @Wynndow I will remove this line from supplier-frontend, are you ok to remove it from frontend-toolkit?

CrystalPea: The same with  `grid = "column-one-third",` line :/

CrystalPea:
You have a good point @samuelhwilliams that the headings should be h2, and I also agree with @TheDoubleK that that calls for a new pattern then, since we change so much in the old one. I guess with @Wynndow we avoided this route since it was meant to be a small story, but happy to go for it :).

Wynndow:
Would it make any sense to have the headings in the temporary messages be h2's and then the headings for the live framework and the side bar ("Your company") be h3's? The temporary nature of temporary messages could make them more important than the existing stuff?

samuelhwilliams:
That would leave us with no `h2`s when there are no temporary messages. I think they're all of equal importance regardless of whether they're ephemeral or not.

laurenceberry:
Sorry for the late response.

The live frameworks are H3s but beneath a visually hidden H2, this makes sense. 
The heading in the sidebar should should also remain as a H2.

Yeah, the headings for the banners should be H2s.

I don't think the font size needs increasing for the banners but yeah they should be H2s.

CrystalPea:
@laurenceberry @samuelhwilliams @TheDoubleK it is h2 now, behold:
ejdztu

DilwoarH:
HI @CrystalPea @TheDoubleK 

Is this intended?

With Temp message:
screen shot 2018-03-21 at 15 04 29

Without Temp message:
screen shot 2018-03-21 at 15 04 51

---
I would suggest doing the following:
screen shot 2018-03-21 at 15 10 53

```
  {% include 'suppliers/_frameworks_coming.html' %}
  {% include 'suppliers/_frameworks_open.html' %}
  {% include 'suppliers/_frameworks_pending.html' %}
  {% include 'suppliers/_frameworks_standstill.html' %}

  <h2 class="visually-hidden">Your services</h2>
  <div class="dmspeak">
    {% include 'suppliers/_frameworks_live.html' %}
  </div>
</div>
```

TheDoubleK:
Yeah those links should always be on the right.

TheDoubleK:
Not sure what changed to make it move.

app/templates/suppliers/dashboard.html
```diff
@@ -60,4 +57,11 @@ <h2 class="heading-xmedium">Your company</h2>
       </p>
     </div>
   </div>
+
+  <h2 class="visually-hidden">Your services</h2>
+  <div class="grid-row">
```
DilwoarH: @TheDoubleK - because it's in a new row, it doesn't keep the alignment

PR #835, opened 2018-03-14
Fix supplier forms validation

Wynndow:
For this tech debt ticket: https://trello.com/c/StDVGuoI
Also, this PR on utils: https://github.com/alphagov/digitalmarketplace-utils/pull/372

### Update validation on ContactInformationForm
Contact name wasn't being validated for length.

The email field can use the EmailValidator we have in utils, with an
updated message.

### Remove unused forms
We no longer use the LoginForm or the ChangePasswordForm. This
functionality has moved to the user app.

Also, the EmailAddressFrom can be refactored to use the common
`EmailField` field from utils. This comes with a free regex validator.

### DIFF
--- a/app/main/forms/auth_forms.py
+++ b/app/main/forms/auth_forms.py
@@ -1,38 +1,10 @@
 from flask_wtf import Form
-from wtforms import PasswordField
-from wtforms.validators import DataRequired, Regexp, EqualTo, Length
+from wtforms.validators import InputRequired
 
-from dmutils.forms import StripWhitespaceStringField
-
-
-class LoginForm(Form):
-    email_address = StripWhitespaceStringField('Email address', validators=[
-        DataRequired(message="Email address must be provided"),
-        Regexp("^[^@^\s]+@[^@^\.^\s]+(\.[^@^\.^\s]+)+$",
-               message="Please enter a valid email address")
-    ])
-    password = PasswordField('Password', validators=[
-        DataRequired(message="Please enter your password")
-    ])
+from dmutils.forms import EmailField
 
 
 class EmailAddressForm(Form):
-    email_address = StripWhitespaceStringField('Email address', validators=[
-        DataRequired(message="Email address must be provided"),
-        Regexp("^[^@^\s]+@[^@^\.^\s]+(\.[^@^\.^\s]+)+$",
-               message="Please enter a valid email address")
-    ])
-
-
-class ChangePasswordForm(Form):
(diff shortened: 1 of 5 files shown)


app/main/forms/auth_forms.py
```diff
@@ -1,38 +1,10 @@
 from flask_wtf import Form
-from wtforms import PasswordField
-from wtforms.validators import DataRequired, Regexp, EqualTo, Length
+from wtforms.validators import DataRequired
```
samuelhwilliams: By and large we use `InputRequired` now, so it might be good to switch over to using that everywhere so it's consistent across this app. Should be a simple find/replace, no other changes needed.

The reason for this is that DataRequired does some extra conversions on any submitted data and checks that it's Truth-y, but as part of doing this is can sometimes throw an error message which isn't overridden properly by the message you pass into the constructor. With InputRequired this doesn't happen.

app/main/forms/auth_forms.py
```diff
@@ -1,38 +1,10 @@
...
-        Length(min=10,
-               max=50,
-               message="Passwords must be between 10 and 50 characters"
-               )
-    ])
-    confirm_password = PasswordField('Confirm password', validators=[
-        DataRequired(message="Please confirm your new password"),
-        EqualTo('password', message="The passwords you entered do not match")
+    email_address = EmailField('Email address', validators=[
+        DataRequired(message="Email address must be provided")
```
samuelhwilliams: I think part of this tech debt ticket was around validating length in more places. It feels like maybe we should add some length validation on email addresses (256 is probably fine, but we could be more generous at 512 if there feels like any good reason...). Not sure if it's better to just add it here or in dmutils though.

app/main/forms/auth_forms.py
Wynndow: Nice

app/main/forms/auth_forms.py
Wynndow: Yeah, I guess I was thinking the email regex would handle it, but I guess it doesn't stop people chucking really long but still valid email addresses in.

This suggests email addresses should have a limit of 254 chars. I'm not sure this includes if you use the '+...@' trick though, as I just sent an email with a 324 character long address to myself.

Either way, I think 512 should be fine. And I do think it should go in the utils. Will add their and link the PR here.

Wynndow: https://github.com/alphagov/digitalmarketplace-utils/pull/372

PR #837, opened 2018-03-14
Edit submit button content for registered address

Wynndow:
All other supplier details questions used 'Save and continue', so updating here to be
consistent.


Wynndow:
Tests are fine, and the func tests that test it are the ones I'm writing at the moment! That's how i noticed it.

PR #838, opened 2018-03-15
Confirm supplier details big green button

TheDoubleK:
For this ticket: https://trello.com/c/gjl4LXzu/60-confirm-supplier-details-3-the-green-button

Adds in button to confirm supplier details.  The button is shown once all details have been provided, and disappears after it's been clicked once to confirm the details are correct.

It looks like this:
screen shot 2018-03-15 at 12 36 53

I've also reworked the logic of deciding if supplier details are complete - previously we had it contingent on whether country is UK or not whether you need Companies House number or some other registration number, but actually UK suppliers could have an "other" registration number.  So we just require one or the other number, not dependent on the country.

### DIFF
--- a/app/main/helpers/suppliers.py
+++ b/app/main/helpers/suppliers.py
@@ -31,14 +31,12 @@ COUNTRY_TUPLE = load_countries()
 
 def supplier_company_details_are_complete(supplier_data):
     supplier_required_fields = ['dunsNumber', 'name', 'registeredName', 'registrationCountry', 'organisationSize',
-                                'tradingStatus']
+                                'tradingStatus', 'vatNumber']
     contact_required_fields = ['address1', 'city', 'postcode']
 
-    registration_country = supplier_data['registrationCountry'] if 'registrationCountry' in supplier_data else None
-    if registration_country == 'gb' or registration_country == 'country:GB':
+    # We require one of either 'companiesHouseNumber' or 'otherCompanyRegistrationNumber'
+    if 'companiesHouseNumber' in supplier_data:
         supplier_required_fields.append('companiesHouseNumber')
-        supplier_required_fields.append('vatNumber')
-
     else:
         supplier_required_fields.append('otherCompanyRegistrationNumber')
 
--- a/app/templates/suppliers/details.html
+++ b/app/templates/suppliers/details.html
@@ -21,8 +21,8 @@
 
 
 {% block main_content %}
-<div class='grid-row'>
-  <div class='column-one-whole padding-bottom-small'>
+  <div class='grid-row'>
+    <div class='column-one-whole padding-bottom-small'>
 {%
   with
     heading = "Company details",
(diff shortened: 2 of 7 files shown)


laurenceberry:
Looks good to me from the screenshots

app/main/helpers/suppliers.py
```diff
@@ -31,14 +31,12 @@ def get_country_name_from_country_code(country_code):
 
 def supplier_company_details_are_complete(supplier_data):
     supplier_required_fields = ['dunsNumber', 'name', 'registeredName', 'registrationCountry', 'organisationSize',
```
Wynndow: Not really related to this, but we have a small handful of suppliers in the db that don't have DUNS numbers. Not sure how. And on the supplier details page, we don't show the DUNS number row if the supplier doesn't have one, so they'd never be able to add one, meaning they could never complete their details.
Do you know why we have suppliers without DUNS? And why the supplier details page conditionally shows the DUNS row if we expect all suppliers to have one?

app/templates/suppliers/details.html
```diff
@@ -21,6 +21,8 @@
 
 
 {% block main_content %}
+<div class='grid-row'>
+  <div class='column-one-whole padding-bottom-small'>
 {%
```
Wynndow: Indentation is a bit skew here, and a the close of the column div.

app/templates/suppliers/details.html
```diff
@@ -170,16 +172,41 @@
...
-    text = "Return to your {} application".format(currently_applying_to.name),
-    bigger = false
-  %}
-    {% include "toolkit/secondary-action-link.html" %}
-  {% endwith %}
-{% endif %}
+  </div>
+
+  {% if not supplier.companyDetailsConfirmed %}
+    <div class='column-two-thirds'>
```
Wynndow: Should separate `column-<some-width>` divs go in their own grid row if they're going to appear below other divs, like the one above? Asking cos I don't know.

tests/app/main/test_suppliers.py
```diff
@@ -955,6 +955,65 @@ def test_back_to_application_link_not_visible_if_currently_applying_to_not_in_se
...
+    def test_post_green_button_redirects_to_the_correct_place(self, data_api_client, current_fwk, expected_destination):
+        data_api_client.get_supplier.return_value = get_supplier()
+        with self.app.test_client():
+            self.login()
+            if current_fwk:
+                with self.client.session_transaction() as session:
+                    session["currently_applying_to"] = current_fwk
+            response = self.client.post("/suppliers/details")
+            assert response.status_code == 302
+            assert response.location == f"http://localhost{expected_destination}"
```
Wynndow: Might be worth adding some tests around the behaviour of a supplier only requiring one of `companiesHouseNumber` or `otherCompanyRegistrationNumber`. All the suppliers returned by `get_supplier` have both.

app/main/helpers/suppliers.py
TheDoubleK: Why...?  "Historical reasons", in that I think they are accounts that got ported from the old system that for some unknown reason were missing DUNS.  But good spot that we need to not break for these.

You're right about conditionally hiding it - should always just be there.  Will fix... somehow.

tests/app/main/test_suppliers.py
TheDoubleK: I thought I had that already, will double check.

app/templates/suppliers/details.html
TheDoubleK: Good question that I don't really know the answer to - the frontend toolkit only has examples for single rows 🤔.   Feels like logically yes they probably should, though I have a feeling it "works" without, and we might have lots of places where they don't.  I'm going to make a new grid-row here though now you've mentioned it.

TheDoubleK:
@Wynndow I think this is all fixed up now.

PR #839, opened 2018-03-15
edit_supplier_registration_number view: fix warning log message

risicle:
https://trello.com/c/qLOqQYy0/226-unable-to-format-log-message-log-messages-not-particularly-useful

Was referring to a parameter `errors` but providing a parameter `rnumber_errors` instead, causing a confusing error when this is issued.

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -243,7 +243,7 @@ def edit_supplier_registration_number():
         else:
             current_app.logger.warning(
                 "supplieredit.fail: has-companies-house-number:{hasnum}, companies-house-number:{chnum}, "
-                "other-registered-company-number:{rnumber}, errors:{errors}",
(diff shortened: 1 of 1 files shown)


risicle:
Whaddya mean wasn't caught by the except block? It was. But that except block, once its emitted its extra error message, doesn't then abort execution of the function - it continues to the point that it returns `log_record`. However of course this `log_record` hasn't had its `message` entry overwritten by the `format()`ted version of itself - that statement didn't succeed. So we end up with the message "template" being sent back out as the log message.

samuelhwilliams:
Oh right I see. It just looks weird in the logs. Maybe as you said, it would be nice to make the extra error message that gets spat out a bit clearer.

risicle:
Yes ... although ... if you read `dmutils/logging.py` a little further up there is a block comment warning about trying to log log failures which I think should be heeded.

However I think it _should_ still be possible to attempt to recover from this particular simple case of formatting entries not existing.

PR #840, opened 2018-03-16
 Standardise button text across supplier app

Wynndow:
### Update button content for copying service
When copying a draft service, after clicking the 'copy' button, the
first question you saw had a 'Save and return' button. Clicking that
though took you to the service summary, instead of returning you 
to the service list, which is inconsistent. 'Save and continue' seemed 
a better choice.

This required the `force_continue_button` argument to be passed in the
redirect from the copy view and accepted in the `edit_service_submission`
view.

Support for passing the `return_to_summary` argument to the
`edit_service_submission` view has also been removed, as it was not
being used anywhere in the app.

### Remove `company_name` template.
It's been replaced in the new supplier creation flow and is no longer
used.

### Update button content
Make sure buttons that return you to previous page you say return. And buttons 
that continue you in a flow say continue.


app/main/views/services.py
```diff
@@ -361,7 +361,7 @@ def copy_draft_service(framework_slug, lot_slug, service_id):
                             service_id=draft_copy['id'],
                             section_id=section_id_to_edit,
                             question_slug=question_slug_to_edit,
-                            return_to_summary=1
+                            force_continue_button=1
```
TheDoubleK: The existing behaviour is exactly as was intended when the "copy draft service" flow was designed and built.

The intention here - when copying a service - is that the supplier hits "copy" and is taken to the page where they have to edit the name of the copied service.  Once the service name has been edited they then "return to summary" to see the overview of the copied service, which they can then edit sections of individually as they choose.

Wynndow: Yeah, it feels a bit weird to me though. Coming from the page where you hit copy, and then seeing 'Save and return to service summary'. You've never been to the service summary for that new service, so you can't return there. 'Return' is back to the page you were on, the service summary list. If you think the original behaviour is ok then I'm happy to just do the button content here, have included a couple of gifs to show the existing and potential new button though:

#### Existing:

save_and_return_to_service_summary

#### Potential:

save_and_continue

TheDoubleK: Agreed that "Save and continue" better the first time after a copy.

TheDoubleK:
Nice, yeah this looks good to me.  Well done for spotting the template I left hanging round recently 👀

TheDoubleK:
Needs a rebase :trollface:

PR #845, opened 2018-03-22
flash messages: port to centralized implementation from frontend toolkit

risicle:
Trello https://trello.com/c/H7upcsyQ/303-flash-messages-suppliers-app

Depends on https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/425/

I *think* I properly converted all of the messages to use the equivalent content that would have been triggered using the previous symbolic messages, though of course it's sometimes hard to know when it comes to inter-app messages.

Added what I think is my preferred way of testing flash messages - `get_flash_messages()` - simple and allows a lot of flexibility for predicates. I like it for similar reasons I like using `Mock`'s `call_args_list`.

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -47,6 +47,19 @@ from ..forms.frameworks import SignerDetailsForm, ContractReviewForm, AcceptAgre
 CLARIFICATION_QUESTION_NAME = 'clarification_question'
 
 
+MESSAGE_SENT_QS_OPEN_MESSAGE = (
+    "Your clarification question has been sent. Answers to all clarification questions will be published on this page."
+)
+MESSAGE_SENT_QS_CLOSED_MESSAGE = (
+    "Your question has been sent. You’ll get a reply from the Crown Commercial Service soon."
+)
+
+
+AGREEMENT_RETURNED_MESSAGE = (
+    "Your framework agreement has been returned to the Crown Commercial Service to be countersigned."
+)
+
+
 @main.route('/frameworks/<framework_slug>', methods=['GET', 'POST'])
 @login_required
 def framework_dashboard(framework_slug):
@@ -480,9 +493,8 @@ def framework_supplier_declaration_submit(framework_slug):
     except APIError as e:
         abort(e.status_code)
 
-    # follow existing flash message passing pattern
     flash_key = "{}/declaration_complete".format(url_for('.framework_dashboard', framework_slug=framework['slug']))
-    flash(flash_key, 'declaration_complete')
+    flash(flash_key, "track-page-view")
     return redirect(url_for('.framework_dashboard', framework_slug=framework['slug']))
 
 
@@ -745,7 +757,10 @@ def framework_updates_email_clarification_question(framework_slug):
         object_id=current_user.supplier_id,
         data={"question": clarification_question, 'framework': framework['slug']})
 
-    flash('message_sent', 'success')
+    flash(
+        MESSAGE_SENT_QS_OPEN_MESSAGE if framework["clarificationQuestionsOpen"] else MESSAGE_SENT_QS_CLOSED_MESSAGE,
+        'success',
+    )
     return framework_updates(framework['slug'])
 
 
@@ -1089,10 +1104,7 @@ def contract_review(framework_slug, agreement_id):
 
             session.pop('signature_page', None)
 
-            flash(
-                'Your framework agreement has been returned to the Crown Commercial Service to be countersigned.',
-                'success'
-            )
+            flash(AGREEMENT_RETURNED_MESSAGE, "success")
 
             if feature.is_active('CONTRACT_VARIATION'):
                 # Redirect to contract variation if it has not been signed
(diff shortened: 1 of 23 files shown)


app/main/views/frameworks.py
```diff
@@ -47,6 +47,19 @@
 CLARIFICATION_QUESTION_NAME = 'clarification_question'
 
 
+MESSAGE_SENT_QS_OPEN_MESSAGE = (
+    "Your clarification question has been sent. Answers to all clarification questions will be published on this page."
+)
+MESSAGE_SENT_QS_CLOSED_MESSAGE = (
+    "Your question has been sent. You'll get a reply from the Crown Commercial Service soon."
```
katstevens: Can we have a curly apostrophe on `You’ll`? 👼

app/main/views/services.py
```diff
@@ -633,6 +640,25 @@ def remove_subsection(framework_slug, lot_slug, service_id, section_id, question
...
     fields_to_remove = question_to_remove.form_fields
 
+    # simply defining this as an inner function so we can avoid having to manually pass around the umpteen variables
+    # it depends on instead pulling them from the outer scope
+    def set_remove_last_subsection_error_message():
+        flash(REMOVE_LAST_SUBSECTION_ERROR_MESSAGE.format(
+            section_name=containing_section.name.lower(),
+            service_name=(draft.get("serviceName") or draft.get("lotName")).lower(),
+        ), "error")
+        flash("/".join((
```
katstevens: Not sure if I like `"/".join()` better than the previous `"{}/{}".format()`, but only a minor quibble.

app/templates/suppliers/dashboard.html
```diff
@@ -15,23 +15,8 @@
...
-        message = message,
-        type = "destructive" if category == 'error' else "success"
-      %}
-        {% include "toolkit/notification-banner.html" %}
-      {% endwith %}
-    {% endfor %}
-  {% endwith %}
 
-  {% if 'account-created' in get_flashed_messages(category_filter=["flag"]) %}
-  <div class="grid-row" data-analytics="trackPageView" data-url="/suppliers?account-created=true">
```
katstevens: As per the toolkit PR, we'll have to work around this one so that the equivalent buyer FE VPV doesn't break.

app/main/views/users.py
```diff
@@ -6,6 +6,9 @@
 from ... import data_api_client
 
 
+DEACTIVATED_USER_MESSAGE = "‘{user_name}’ ({user_email_address}) has been removed as a contributor."
```
katstevens: I don't think the old version had quotes around the `{user_name}` (however I'm not averse to them being put in!)

app/templates/_base_page.html
```diff
@@ -29,6 +29,7 @@
   {% block breadcrumb %}{% endblock %}
   <div id="wrapper">
     <main id="content" role="main">
+      {% include "toolkit/flash_messages.html" %}
```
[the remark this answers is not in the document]
risicle: That's a good question - where is best to put this? I'm not sure I gave it a huge amount of thought - there's no hugely intentional reason for it going specifically *here*.

app/main/views/login.py
```diff
@@ -10,6 +10,9 @@
 from ... import data_api_client
 
 
+USER_INVITED_FLASH_MESSAGE = "User invited"
```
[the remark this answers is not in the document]
risicle: Hmmmmmmmmmm ... wonder where I picked up that piece of content from...

app/main/views/users.py
risicle: Yeah I originally added quotes in most of the content strings that were missing them but then removed them as I didn't want the scope of this to creep, I noticed I missed this one as I was pushing and of course the tests don't catch it because previously most the time they just checked for the symbolic flash message.

app/templates/_base_page.html
```diff
@@ -29,6 +29,42 @@
...
+            {% if category == 'track-page-view' %}
+              <div data-analytics="trackPageView" data-url="{{ message }}"></div>
+            {% elif category == 'flag' %}
+              {# these were originally messages with a function similar to the above "track-page-view" ones (analytics) but #}
+              {# they did not know about their actual virtual page view url, expecting the receiving page to know the       #}
+              {# corresponding url for a particular symbolic "message".                                                     #}
+              {# TODO these can be got rid of as soon as the app generating each flag flash message is converted to using   #}
+              {# new "track-page-view" mechanism                                                                            #}
+              {% if message == "account-created" %}
+                <div data-analytics="trackPageView" data-url="/suppliers?account-created=true"></div>
```
katstevens: Looking at the User FE: we have the redirect destination (used here in `data-url`) at the point where we trigger the flash message. So we should just be able to do:

```
redirect_path = '/suppliers' if current_user.role == 'supplier' else .... # etc
flash("{}?account-created=true".format(redirect_path), category="track-page-view")
return redirect(redirect_path)
```
And the path will be slotted as the `{{ message }}` in the `trackPageView` above.

risicle: Well ... if we were certain we could get the that user FE change merged (and released!) before this, we could just forget all the backwards compat...

katstevens: 🎉 https://github.com/alphagov/digitalmarketplace-user-frontend/pull/50

risicle: Cool, but as discussed I think I still need to merge this ugliness to allow a clean transition...

katstevens: Agreed. We shouldn't forget to clean up the equivalent Buyer FE one as well afterwards.

PR #854, opened 2018-04-12
[WIP] Cpw copy current services combined

Wynndow:


app/main/__init__.py
```diff
@@ -36,6 +36,7 @@
 
 content_loader.load_manifest('g-cloud-10', 'services', 'edit_service')
 content_loader.load_manifest('g-cloud-10', 'services', 'edit_submission')
+content_loader.load_manifest('g-cloud-10', 'services', 'copy_services')
```
samuelhwilliams: This will be `content_loader.load_metadata('g-cloud-10', ['copying-services'])` now.

app/main/views/frameworks.py
```diff
@@ -260,6 +260,17 @@ def framework_submission_services(framework_slug, lot_slug):
     if framework['status'] == 'pending' and declaration_status != 'complete':
         abort(404)
 
+    previous_services = data_api_client.find_services(
+        supplier_id=current_user.supplier_id,
+        framework='g-cloud-9',
```
samuelhwilliams: Should be `content_loader.get_metadata('g-cloud-10', 'copying-services', 'framework')` now - although will be best to hit the API to get these framework details for later use.

samuelhwilliams:
Got a few flake8 errors going on here. Let me know when this is uptodate and I'll review.

Wynndow:
@samuelhwilliams All done for now, comment away!

app/main/views/frameworks.py
```diff
@@ -260,6 +260,26 @@ def framework_submission_services(framework_slug, lot_slug):
...
+        previous_framework = get_framework_or_404(data_api_client, previous_framework_slug)
+        previous_services = data_api_client.find_services(
+            supplier_id=current_user.supplier_id,
+            framework=previous_framework_slug,
+            lot=lot_slug,
+            status='published',
+        )["services"]
+
+    previous_services_still_to_copy = len([
+        service for service in previous_services if not service['copiedToFollowingFramework']
```
samuelhwilliams: A `filter` here would be slightly more Pythonic, I think. But no big deal either way.

`previous_services_still_to_copy = len(filter(lambda x: x['copiedToFollowingFramework'] is False, services))`

app/templates/services/previous_services.html
```diff
@@ -0,0 +1,89 @@
...
+
+  {% include "partials/service_warning.html" %}
+
+  {% if copy_all %}
+    <form action="{{ url_for('.copy_all_previous_services', framework_slug=framework.slug, lot_slug=lot.slug) }}" method="POST">
+      <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
+      {%
+        with
+        message = "Are you sure you want to add all your {} services?".format(lot.name|lower),
+        type = "destructive",
```
samuelhwilliams: I think this should be a green flash message. I recognise that it's a strong action to take, but it's definitely not destructive and I think red sends the wrong signal.

If we have some kind of 'warning' type (orange maybe), that would be fine. But I think we want to keep destructive/red messages to things that actually do deletions.

tests/app/main/test_frameworks.py
```diff
@@ -4095,6 +4106,58 @@ def test_banner_on_service_pages_shows_links_to_company_details_and_declaration(
...
+        data_api_client.get_framework.return_value = self.framework(status='open')
+        self.login()
+
+        res = self.client.get('/suppliers/frameworks/g-cloud-7/submissions/scs')
+        doc = html.fromstring(res.get_data(as_text=True))
+        link = doc.xpath(
+            "//*[@id='content']/p[1]/a[normalize-space(string())='View and add your services from G-Cloud\xa07']"
+        )
+
+        data_api_client.find_services.assert_called_once_with(
```
samuelhwilliams: I think we're trying to move away from using things like `assert_called_once_with` on mock objects, because it's easy to typo that method call and then have it get swallowed by the mock.

I think a better test is something like `assert data_api_client.find_services.call_args_list == [mock.call(supplier_id=1234, framework='g-cloud-6', lot='scs', status='published')]`

tests/app/main/test_frameworks.py
```diff
@@ -4095,6 +4106,58 @@ def test_banner_on_service_pages_shows_links_to_company_details_and_declaration(
...
+            supplier_id=1234,
+            framework='g-cloud-6',
+            lot='scs',
+            status='published',
+        )
+
+        if link_shown:
+            assert link
+            assert 'View and add your services from G-Cloud\xa07\n' in link[0].text
```
samuelhwilliams: You've already asserted this in the xpath.

tests/app/main/test_services.py
```diff
@@ -2275,3 +2275,317 @@ class TestParseDocumentUploadTime(BaseApplicationTest):
...
+                {'serviceName': 'Service two', 'copiedToFollowingFramework': False},
+                {'serviceName': 'Service three', 'copiedToFollowingFramework': False},
+            ],
+        }
+        get_metadata.return_value = 'g-cloud-9'
+
+        res = self.client.get('/suppliers/frameworks/g-cloud-10/submissions/cloud-hosting/previous-services')
+        assert res.status_code == 200
+
+        data_api_client.find_services.assert_called_once_with(
```
samuelhwilliams: Another `assert_called_once_with` to avoid.

tests/app/main/test_services.py
```diff
@@ -2275,3 +2275,317 @@ class TestParseDocumentUploadTime(BaseApplicationTest):
...
+            '/suppliers/frameworks/g-cloud-10/submissions/cloud-hosting/copy-previous-framework-service/2000000000'
+        )
+
+        assert res.status_code == 302
+        assert res.location == (
+            "http://localhost/suppliers/frameworks/"
+            "g-cloud-10/submissions/cloud-hosting/previous-services"
+        )
+
+        self.data_api_client.copy_draft_service_from_existing_service.assert_called_once_with(
```
samuelhwilliams: Another one.

tests/app/main/test_services.py
```diff
@@ -2275,3 +2275,317 @@ class TestParseDocumentUploadTime(BaseApplicationTest):
...
+class TestCopyAllPreviousServices(CopyingPreviousServicesSetup):
+    def test_copies_all_services(self):
+        res = self.client.post(
+            '/suppliers/frameworks/g-cloud-10/submissions/cloud-hosting/copy-all-previous-framework-services'
+        )
+
+        assert res.status_code == 302
+        assert res.location == 'http://localhost/suppliers/frameworks/g-cloud-10/submissions/cloud-hosting'
+
+        self.data_api_client.copy_published_from_framework.assert_called_once_with(
```
samuelhwilliams: ☝️

app/templates/services/previous_services.html
Wynndow: Constance and I spend a bit of time looking at this. We decided that whilst not destructive, the implications of copying all your services when you didn't mean to and then having to delete them all from your list of drafts, merited a red message.

There _is_ and orange box, but then it looks really weird with the red button. If Constance is about, have a chat with her about it and see what she thinks.

tests/app/main/test_frameworks.py
Wynndow: Really? I find doing it that way really ugly! Having to use `mock.call` just gets all cluttered. If you use `autospec=True` with the mock (which I haven't here) it'll catch the error no?

If we're moving away from it as a whole, I'm happy to change here though.

app/main/views/frameworks.py
Wynndow: I'm gonna keep as it actually. `filter` objects don't have `len()`, so it becomes `len(list(filter(....)))` which is a bit naf.

samuelhwilliams: 🤦‍♂️ 

I thought it would just iterate over the filtered list. Sometimes Python 3 feels a bit like it's made things a bit weird.

Wynndow: Yeah, i find `filter` and `map` objects really unintuitive.

app/templates/services/previous_services.html
samuelhwilliams: Copying services is a positive thing. It's going to help the supplier. I think it feels like we're warning them against doing it and that's going to put the supplier in a weird place mentally.

It's a big action, and that's why we're making the user confirm they want to take it. But it's not a _bad_ action.

I'll chat to Constance if I see her.

samuelhwilliams: P.S. I probably shouldn't have mentioned the orange box as I think @laurenceberry wanted us to get rid of it (and maybe we already did).

app/templates/services/previous_services.html
```diff
@@ -0,0 +1,89 @@
...
+{% block main_content %}
+
+  {% include "partials/service_warning.html" %}
+
+  {% if copy_all %}
+    <form action="{{ url_for('.copy_all_previous_services', framework_slug=framework.slug, lot_slug=lot.slug) }}" method="POST">
+      <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
+      {%
+        with
+        message = "Are you sure you want to add all your {} services?".format(lot.name|lower),
```
samuelhwilliams: Can we show this above the `partials/service_warning` so that both green boxes (the request for confirmation and the flash message) appear above the blue box. Otherwise it's a bit weird - the confirmation is below and the flash is above.

app/main/views/services.py
```diff
@@ -702,3 +709,106 @@ def set_remove_last_subsection_error_message():
...
+
+
+@main.route('/frameworks/<framework_slug>/submissions/<lot_slug>/previous-services', methods=['GET'])
+@login_required
+def list_previous_services(framework_slug, lot_slug):
+    framework, lot = get_framework_and_lot_or_404(data_api_client, framework_slug, lot_slug)
+    if framework['status'] != 'open':
+        abort(404)
+
+    copy_all = request.args.get('copy_all', None)
```
samuelhwilliams: Have the query param part of this be hyphenated would be more consistent with the rest of the URL. Not a big deal if you want to leave this as is, but IMO it'd be nicer as `copy-all`. Your call.

Wynndow: Weirdness about having a hyphen in the key for the parameter to `url_for()` (causes a Jinja2 syntax error) so gonna leave for now.

tests/app/main/test_frameworks.py
katstevens: 👋 

A quick summary (as sent to Sam on Slack)
1) easier to avoid typos that pass silently (especially if the mock itself hasn't been set up correctly). Although `mock` now catches typos for anything starting `assert` or `assret`,  if you absent mindedly miss that `assert` prefix off the method, e.g.  `data_api_client.called_once_with('wrong')`, it will magically pass. Better to have the `assert` up front IMHO.
2) it explicitly states ordering of the calls (and any failed assertions will show if the right args are there but in the wrong order - we can always use a set if we don't care) 
3) I find it easier to visually compare with other test cases where the code may result in more than one function call (or no calls).

Agreed it's a bit verbose, but as mocks are often misunderstood (especially for less experienced folks) I think that's a good thing.

Wynndow: Thanks Kat, those are all solid points. I will endeavour to follow them in future!

PR #856, opened 2018-04-16
Refactor application started email

samuelhwilliams:
## Summary
Refactors the emails which get sent out when a supplier user starts an
application with a framework. Instead of having a template for each
framework iteration, we can have one for each framework family and
inject the required dynamic content using the content-loader. This
should mean we don't need to update this template every time we run a
new framework.

In theory it might be possible to refactor this further and have a
single 'application started' template for all frameworks - but it seems
reasonable to allow differences in the email between frameworks in case
guidance needs to be materially different.

Alternative PR to https://github.com/alphagov/digitalmarketplace-apiclient/pull/133

## Ticket
https://trello.com/c/y3NrHfy5/39-content-requiring-final-dates-1-you-have-started-an-application-email


tests/app/main/test_frameworks.py
```diff
@@ -182,7 +182,7 @@ def test_email_sent_when_interest_registered_in_framework(self, send_email, data
             'You started a G-Cloud 7 application',
             'do-not-reply@digitalmarketplace.service.gov.uk',
             'Digital Marketplace Admin',
-            ['digital-outcomes-and-specialists-application-started']
+            ['g-cloud-7-application-started']
```
Wynndow: Is is also worth checking that `render_template` got called with the correct arguments when the email is templated?

samuelhwilliams: So I tried to mock out `render_template` and _everything_ broke. I spent about 5 minutes looking into it and gave up. I could put some more effort in as it does feel like a sensible thing to test. 🤔

app/templates/emails/g-cloud_application_started.html
```diff
@@ -0,0 +1,37 @@
...
+      <li>complete your company details</li>
+      <li>make the supplier declaration</li>
+      <li>provide information about the services you offer, and mark each service as complete</li>
+    </ul>
+    <br/><br/>
+    To understand if your services are suitable for {{ framework.name }}, read this guidance:
+    <br/>
+    <a href="https://www.gov.uk/guidance/g-cloud-suppliers-guide">https://www.gov.uk/guidance/g-cloud-suppliers-guide</a>
+    <br/><br/>
+    The {{ framework.name }} application section of your Digital Marketplace account helps you:
```
Wynndow: The content linked in the Trello ticket shows that this should have a non breaking space. I don't think the name from the framework does. Could use the nbsp filter. 

Having said that, it's not likely that it's ever going to be an issue here so you may not need to bother.

app/templates/emails/g-cloud_application_started.html
```diff
@@ -0,0 +1,37 @@
...
+    {{ framework_dates.clarifications_close_date }}.
+    <br/><br/>
+    To complete your application, you must:
+    <ul>
+      <li>complete your company details</li>
+      <li>make the supplier declaration</li>
+      <li>provide information about the services you offer, and mark each service as complete</li>
+    </ul>
+    <br/><br/>
+    To understand if your services are suitable for {{ framework.name }}, read this guidance:
```
Wynndow: Same here as the nbsp comment below

tests/app/main/test_frameworks.py
Wynndow: _Everything_ broke? That is a lot....

app/templates/emails/g-cloud_application_started.html
samuelhwilliams: Good spot. Added.

app/templates/emails/g-cloud_application_started.html
samuelhwilliams: As above - added. 👍

tests/app/main/test_frameworks.py
Wynndow: There is this `wraps` thing that I bumped into a the weekend. I've not used it, but I think it allows you to mock an object without affecting it's functionality, so you can effectively spy on it. Might be useful here? https://docs.python.org/3/library/unittest.mock.html

samuelhwilliams:
Should all be addressed.

Wynndow:
Nice. 👍

PR #862, opened 2018-04-17
Check contact details and declaration status for warning banner

Wynndow:
Without these variables, the banner would show regardless of the
suppliers state. That is bad.

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -16,7 +16,6 @@ from ..helpers.frameworks import (
     get_supplier_framework_info,
     get_framework_or_404,
 )
-from ..helpers.suppliers import supplier_company_details_are_complete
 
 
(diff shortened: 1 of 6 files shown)


app/main/views/services.py
```diff
@@ -735,13 +736,17 @@ def list_previous_services(framework_slug, lot_slug):
...
+
     return render_template(
         "services/previous_services.html",
         framework=framework,
         lot=lot,
         source_framework=source_framework,
         previous_services=previous_services_still_to_copy,
         copy_all=copy_all,
+        declaration_status=get_declaration_status(data_api_client, framework_slug),
+        company_details_complete=supplier_company_details_are_complete(supplier),
```
samuelhwilliams: This should be `supplier['companyDetailsConfirmed']`, as per my latest commit changes.

PR #863, opened 2018-04-17
Don't copy questions that depend on other questions

Wynndow:
See this PR on the frameworks repo: https://github.com/alphagov/digitalmarketplace-frameworks/pull/519


samuelhwilliams:
When tests pass...

PR #871, opened 2018-04-24
Send clarification email through Notify

samuelhwilliams:
## Summary
When a supplier submits a clarification question, the user who submitted
it will receive email confirmation that it has been submitted
successfully. This is currently sent through Mandrill, but we want to
try to move more of our notification emails over to Notify.

## Ticket
https://trello.com/c/7dRVK9Yt/68-send-asked-clarification-question-to-supplier-emails-through-notify

## Question
One thing to think about is if this is how we want to handle instantiating the Notify client or if we should do it in a similar way to the API clients. If the latter, we'll need to re-work how the Notify client is constructed (i.e. give it an `init_app` thing and allow it to take the tokens outside of the initializer). I suspect probably not. I started looking into https://github.com/alecthomas/flask_injector as a way of injecting dependencies into views, but haven't dug down into this much yet.

## TODO
 * [x] Do not merge until functional tests have been checked/updated.


tests/app/main/test_frameworks.py
```diff
@@ -3646,11 +3645,9 @@ def _send_email(self, clarification_question):
             data={'clarification_question': clarification_question}
         )
 
-    def _assert_clarification_email(self, send_email, is_called=True, succeeds=True):
-
-        if succeeds:
-            assert send_email.call_count == 2
```
galund: Why did it used to get called more than once??

app/main/views/frameworks.py
```diff
@@ -741,33 +743,32 @@ def framework_updates_email_clarification_question(framework_slug):
     if framework['clarificationQuestionsOpen']:
         # Send confirmation email to the user who submitted the question
         # No need to fail if this email does not send
-        subject = current_app.config['CLARIFICATION_EMAIL_SUBJECT']
-        tags = ["clarification-question-confirm"]
         audit_type = AuditTypes.send_clarification_question
-        email_body = render_template(
-            "emails/clarification_question_submitted.html",
```
galund: Is there a template that should now be deleted?

tests/app/main/test_frameworks.py
samuelhwilliams: It sends two emails - one to admins notifying of a new clarification question, and then a confirmation email to the supplier.

We have only converted one of the two at this point, so there's still one email getting sent with mandrill.

app/main/views/frameworks.py
samuelhwilliams: Yes it is - gone.

tests/app/main/test_frameworks.py
galund: Can we maybe rename `send_email` to `mandrill_send_email`, if I'm understanding correctly what that means then. But yep.

samuelhwilliams: Ah yep sure, will do that.

samuelhwilliams:
TODO: Do not merge until functional tests have been checked/updated.

PR #872, opened 2018-04-24
[wip] Shw dates from api

samuelhwilliams:
Sits on top of https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/871.

WIP. Do not review.


tests/app/helpers.py
```diff
@@ -312,51 +313,44 @@ def services():
...
-                'clarificationQuestionsOpen': clarification_questions_open,
-                'name': name,
-                'slug': slug,
-                'lots': lots,
-                'frameworkAgreementVersion': framework_agreement_version,
-                'framework': metaframework,
-            }
-        }
+        elif 'digital-outcomes-and-specialists' in slug:
+            lots = [api_stubs.lot(lot_id=1, slug='digital-specialists', name='Digital specialists',
```
Wynndow: I know this is a straight port over to using the stubs, but maybe the other DOS lots should go in here too.

tests/app/main/test_frameworks.py
```diff
@@ -390,7 +391,7 @@ def test_downloads_shown_open_framework(self, data_api_client, s3):
         )
         assert len(doc.xpath(
             "//main//p[contains(normalize-space(string()), $a)]",
-            a="until 5pm BST, 22 September 2015",
+            a="until 5pm BST, Tuesday 22 September 2015",
```
Wynndow: Strange that the content has changed here. Was this just a oddity?

samuelhwilliams: Yeah, in G-Cloud 7 dates, we didn't show the name of the day. In later (current) frameworks, we do. Now that we're storing it as timestamps and formatting it consistently, all frameworks will show the day of the month. Looks weird, but it means our test cases more accurately represent the current content, so I think it's all good 👍

samuelhwilliams: https://github.com/alphagov/digitalmarketplace-frameworks/blob/e1f8e5c4a1a98d817e1bc9ff90ace85c1a0762c6/frameworks/g-cloud-7/messages/dates.yml

vs

https://github.com/alphagov/digitalmarketplace-frameworks/blob/e1f8e5c4a1a98d817e1bc9ff90ace85c1a0762c6/frameworks/g-cloud-10/messages/dates.yml

tests/app/main/test_suppliers.py
```diff
@@ -16,38 +18,34 @@
...
+    return {
+        'applications_close_at': datetime(2000, month, 3),
+        'intention_to_award_at': datetime(2000, month, 4),
+        'clarifications_close_at': datetime(2000, month, 1),
+        'clarifications_publish_at': datetime(2000, month, 2),
+        'framework_live_at': datetime(2000, month, 5),
+    }
+
+
+FIND_FRAMEWORKS_RETURN_VALUE = {
```
Wynndow: Would a global like this normally go above the function definition?

tests/app/main/test_suppliers.py
```diff
@@ -16,38 +18,34 @@
...
+        framework_stub(status='expired', slug='h-cloud-88', name='H-Cloud 88', framework_family='g-cloud',
+                       **framework_stub_dates(1))['frameworks'],
+
+        framework_stub(status='live', slug='g-cloud-6',
+                       **framework_stub_dates(2))['frameworks'],
+
+        framework_stub(status='open', slug='digital-outcomes-and-specialists',
+                       **framework_stub_dates(4))['frameworks'],
+
+        framework_stub(status='live', slug='digital-rhymes-and-reasons', name='Digital Rhymes and Reasons',
```
Wynndow: This framework is the best framework.

tests/app/main/test_suppliers.py
```diff
@@ -419,13 +389,7 @@ def test_shows_gcloud_7_in_standstill_application_passed_with_message(
...
         }
         data_api_client.find_frameworks.return_value = {
-            "frameworks": [
-                {
-                    'status': 'standstill',
-                    'slug': 'g-cloud-7',
-                    'name': 'G-Cloud 7'
-                }
-            ]
+            'frameworks': [self.framework(status='standstill', slug='g-cloud-7')['frameworks']]
```
Wynndow: Why the use of `self.framework` from the BaseApplicationTest, which is an api_stub anyway, instead of using the stub directly?

Wynndow:
Few small comments but otherwise looks really good.

tests/app/main/test_suppliers.py
samuelhwilliams: I think we tend to define globals at the top of the file. Either way, this isn't a new global that I've added - I just renamed it to all caps so it's more obviously a global.

tests/app/main/test_suppliers.py
samuelhwilliams: No real reason I suppose. `self.framework` adds the lots, which I guess made it feel more complete. But they're irrelevant here. 🤔

I'll change it to use the stub as you recommend.

samuelhwilliams: I take it back. Going to leave it as-is. The `self.framework` establishes common dates that are used across a set of tests, including this one, which gets asserted on later. Better to just have these dates in one place.

PR #873, opened 2018-04-27
Password change link for logged in suppliers

katstevens:
Trello: https://trello.com/c/TieWrBG5/44-password-reset-link-from-logged-in-session

change-your-password-supplier

Adds a dashboard link to the new 'Change your password' form, for logged in supplier users.

To be merged after the equivalent User FE branch.


tests/app/main/test_suppliers.py
```diff
@@ -1150,6 +1150,10 @@ def test_should_show_supplier_dashboard_logged_in(
             assert self.strip_all_whitespace("email@email.com") in \
                 self.strip_all_whitespace(res.get_data(as_text=True))
 
+            assert self.strip_all_whitespace("Change your password") in \
+                self.strip_all_whitespace(res.get_data(as_text=True))
+            assert "href=\"/user/change-password\"" in res.get_data(as_text=True)
+
```
risicle: I'd... personally... be a little more stringent here...? But I guess it's not your job to convert all test assertions you find to xpath assertions.

katstevens: Yeah, we copied the format from a similar test elsewhere. Happy to change if there's a strong feeling (guessing not...)

samuelhwilliams:
Slightly different padding here between the two groups of links. Any way to get consistent?

screen shot 2018-05-01 at 15 43 31

katstevens:
Interesting - I guess that Account details section has only ever had one link in it before! I don't generally like `<br />`s when `<p>`s will do the job, so I'll change the existing one.

samuelhwilliams:
That looks better (to me).

tests/app/main/test_suppliers.py
samuelhwilliams: Might be nicer to use single-quotes on the outside as well, so you don't need to escape the double quotes in the HTML.

I agree with Rob that it feels weird to test the kink is present using this check. Xpath would probably be more sensible.

PR #882, opened 2018-05-25
Update content for become a supplier messages

Wynndow:
For this trello ticket: [https://trello.com/c/7P1cRFqt](https://trello.com/c/7P1cRFqt)

### Pull in updated content for `become a supplier` page

The content was wrong and needed changing. This has been done in a separate PR on the frameworks repo and is already [merged](https://github.com/alphagov/digitalmarketplace-frameworks/pull/522)

### Fix framework ordering bug on `become_a_supplier`

We were ordering frameworks by their slug and then reversing. This was
so that they were ordered by family, newest first. This broke down
when we got to g-cloud-10 as lexicographically before g-cloud-9.

Ordering first by framework, then by id has the desired effect.

Screenshot of new content

screen shot 2018-05-25 at 12 40 37

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -507,7 +507,7 @@ def become_a_supplier():
     try:
         frameworks = sorted(
             data_api_client.find_frameworks().get('frameworks'),
-            key=lambda framework: framework['slug'],
+            key=lambda framework: framework['id'],
             reverse=True
         )
         displayed_frameworks = get_frameworks_closed_and_open_for_applications(frameworks)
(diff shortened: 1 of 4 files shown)


app/main/views/suppliers.py
```diff
@@ -507,7 +508,7 @@ def become_a_supplier():
     try:
         frameworks = sorted(
             data_api_client.find_frameworks().get('frameworks'),
-            key=lambda framework: framework['slug'],
+            key=operator.itemgetter('framework', 'id'),
```
galund: I'm surprised this works, why is the first argument quoted?

galund: Actually, take that back, the quoting would make sense, except I don't see why the first argument is there -at all-. Probably me being daft tho at this point :)

Wynndow: So it's so the list of frameworks being sorted are first sorted by framework, and then by id within that first sorting. Keeps the family members grouped.

Wynndow: But yeah, I see what you mean now. Is keeping them grouped actually necessary. Probably not.

galund: Oh, framework-framework - right yep. So this is a change, somehow I assumed it was just intended to be the same ordering but written differently. If there's any reason to order at all then go for it, don't mind me :)

Wynndow: @galund Have updated to drop to just sorting by ID

PR #883, opened 2018-05-26
Rename framework_framework variable

samuelhwilliams:
## Summary
`framework_framework` is an unnecessarily obtuse variable name and is
not overly descriptive of what it actually means. When used, this
describes the 'family' of framework being referenced. For example,
G-Cloud 10 is in the 'G-Cloud' family of frameworks. Digital Outcomes
and Specialists 2 is in the 'Digital Outcomes and Specialists' family.
Let's change all the references of `framework_framework` to
`framework_family`

 ## Ticket
https://trello.com/c/iqFcVKpd/14

### DIFF
--- a/app/main/helpers/frameworks.py
+++ b/app/main/helpers/frameworks.py
@@ -343,7 +343,7 @@ def check_agreement_is_related_to_supplier_framework_or_abort(agreement, supplie
 
 
 def get_frameworks_closed_and_open_for_applications(frameworks):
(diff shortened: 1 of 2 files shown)


app/main/helpers/frameworks.py
```diff
@@ -354,7 +354,7 @@ def status_priority(status):
```
Wynndow: Line 346 above has a framework-framework

PR #886, opened 2018-05-29
Digital Outcomes and Specialists - Copying services

Wynndow:
For [this trello ticket.](https://trello.com/c/QhfXDVjt)

Adds the logic to allow suppliers to copy previous services for single service lots. Uses a similar pattern to declaration copying, with a single page boolean question.

I'm not convinced by the content in this PR as it stands. Need to go over with a content person.

### DIFF
--- a/app/main/forms/frameworks.py
+++ b/app/main/forms/frameworks.py
@@ -50,3 +50,16 @@ class ReuseDeclarationForm(FlaskForm):
         validators=[InputRequired(message='You must answer this question.')]
     )
     old_framework_slug = HiddenField()
+
+
+class OneServiceLimitCopyServiceForm(FlaskForm):
+
+    def __init__(self, lot_name, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.copy.label.text = f"Do you want to reuse your previous {lot_name} service?"
+
+    copy = BooleanField(
+        false_values={'False', 'false', ''},
+        validators=[InputRequired(message='You must answer this question.')]
+    )
+    service_id = HiddenField()
--- a/app/main/forms/frameworks.py
+++ b/app/main/forms/frameworks.py
@@ -56,9 +56,9 @@ class OneServiceLimitCopyServiceForm(FlaskForm):
 
     def __init__(self, lot_name, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.copy.label.text = f"Do you want to reuse your previous {lot_name} service?"
+        self.copy_service.label.text = f"Do you want to reuse your previous {lot_name} service?"
 
-    copy = BooleanField(
+    copy_service = BooleanField(
         false_values={'False', 'false', ''},
         validators=[InputRequired(message='You must answer this question.')]
     )
--- a/app/main/forms/frameworks.py
+++ b/app/main/forms/frameworks.py
@@ -59,7 +59,7 @@ class OneServiceLimitCopyServiceForm(FlaskForm):
         self.copy_service.label.text = f"Do you want to reuse your previous {lot_name} service?"
 
     copy_service = BooleanField(
+        'Do you want to reuse your previous service?',
         false_values={'False', 'false', ''},
         validators=[InputRequired(message='You must answer this question.')]
     )
-    service_id = HiddenField()
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -271,7 +271,15 @@ def framework_submission_services(framework_slug, lot_slug):
         previous_framework = {}
 
     if previous_framework_slug:
-        previous_framework = get_framework_or_404(data_api_client, previous_framework_slug)
+        try:
+            previous_framework = data_api_client.get_framework(previous_framework_slug)['frameworks']
+        except HTTPError as e:
+            current_app.logger.error(
+                "Failed to find previous framework. Error: {error}, framework_slug: {framework_slug}",
+                extra={'error': str(e), 'framework_slug': previous_framework_slug}
+            )
+            abort(500, f'Previous framework not found: {previous_framework_slug}')
+
         previous_services = data_api_client.find_services(
             supplier_id=current_user.supplier_id,
             framework=previous_framework_slug,
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -287,7 +287,7 @@ def framework_submission_services(framework_slug, lot_slug):
         draft = next(iter(drafts + complete_drafts), None)
         if not draft and previous_services_still_to_copy:
             return redirect(
-                url_for('.list_previous_services', framework_slug=framework_slug, lot_slug=lot_slug)
+                url_for('.previous_services', framework_slug=framework_slug, lot_slug=lot_slug)
             )
 
         elif not draft:
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -285,7 +285,12 @@ def framework_submission_services(framework_slug, lot_slug):
 
     if lot['oneServiceLimit']:
         draft = next(iter(drafts + complete_drafts), None)
-        if not draft:
+        if not draft and previous_services_still_to_copy:
+            return redirect(
+                url_for('.list_previous_services', framework_slug=framework_slug, lot_slug=lot_slug)
+            )
+
+        elif not draft:
             draft = data_api_client.create_new_draft_service(
                 framework_slug, lot_slug, current_user.supplier_id, {}, current_user.email_address,
             )['services']
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -32,7 +32,7 @@ from ..helpers import login_required
 from ..helpers.frameworks import (
     get_declaration_status, get_last_modified_from_first_matching_file, register_interest_in_framework,
     get_supplier_on_framework_from_info, get_declaration_status_from_info, get_supplier_framework_info,
-    get_framework_or_404, get_framework_and_lot_or_404, count_drafts_by_lot, get_statuses_for_lot,
+    get_framework_or_404, get_framework_and_lot_or_404, get_framework_or_500, count_drafts_by_lot, get_statuses_for_lot,
     return_supplier_framework_info_if_on_framework_or_abort, returned_agreement_email_recipients,
     check_agreement_is_related_to_supplier_framework_or_abort, get_framework_for_reuse,
     get_supplier_registered_name_from_declaration
@@ -271,15 +271,7 @@ def framework_submission_services(framework_slug, lot_slug):
         previous_framework = {}
 
     if previous_framework_slug:
-        try:
-            previous_framework = data_api_client.get_framework(previous_framework_slug)['frameworks']
-        except HTTPError as e:
-            current_app.logger.error(
-                "Failed to find previous framework. Error: {error}, framework_slug: {framework_slug}",
-                extra={'error': str(e), 'framework_slug': previous_framework_slug}
-            )
-            abort(500, f'Previous framework not found: {previous_framework_slug}')
-
+        previous_framework = get_framework_or_500(data_api_client, previous_framework_slug, logger=current_app.logger)
         previous_services = data_api_client.find_services(
             supplier_id=current_user.supplier_id,
             framework=previous_framework_slug,
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -10,7 +10,13 @@ from dmutils.documents import upload_service_documents
 from ... import data_api_client, flask_featureflags
 from ...main import main, content_loader
 from ..helpers import login_required
-from ..helpers.services import is_service_associated_with_supplier, get_signed_document_url, count_unanswered_questions
+from ..helpers.services import (
+    copy_service_from_previous_framework,
+    count_unanswered_questions,
+    get_lot_drafts,
+    get_signed_document_url,
+    is_service_associated_with_supplier,
+)
 from ..helpers.frameworks import (
     get_framework_and_lot_or_404,
     get_declaration_status,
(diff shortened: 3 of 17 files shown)


app/main/forms/frameworks.py
```diff
@@ -50,3 +50,16 @@ class ReuseDeclarationForm(FlaskForm):
         validators=[InputRequired(message='You must answer this question.')]
     )
     old_framework_slug = HiddenField()
+
+
+class OneServiceLimitCopyServiceForm(FlaskForm):
+
+    def __init__(self, lot_name, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.copy.label.text = f"Do you want to reuse your previous {lot_name} service?"
```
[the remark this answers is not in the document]
galund: Oh, duuuuh I see :-)

app/main/views/services.py
```diff
@@ -715,12 +722,35 @@ def set_remove_last_subsection_error_message():
...
 
 
-@main.route('/frameworks/<framework_slug>/submissions/<lot_slug>/previous-services', methods=['GET'])
+@main.route('/frameworks/<framework_slug>/submissions/<lot_slug>/previous-services', methods=['GET', 'POST'])
 @login_required
 def list_previous_services(framework_slug, lot_slug):
     framework, lot = get_framework_and_lot_or_404(data_api_client, framework_slug, lot_slug)
     if framework['status'] != 'open':
         abort(404)
+    form = OneServiceLimitCopyServiceForm(lot['name'].lower())
```
galund: There's something quite confusing here about having this view be both for listing services and for showing this special form. It feels like really they ought to be separate views, if at all possible - and it looks like the template is really two templates with a massively important `if` to switch between these two cases, as well!

Wynndow: My thinking about keeping them together, and perhaps the naming of the view doesn't help, is that there's one view that deals with presenting the supplier with the options for copying their previous services - regardless of the type of lot.

I thought about splitting the template out into partials. Happy to split it into two separate templates though if you think it'd be better?

app/main/forms/frameworks.py
```diff
@@ -50,3 +50,16 @@ class ReuseDeclarationForm(FlaskForm):
...
+
+    copy = BooleanField(
```
samuelhwilliams: Might be nice to call this `copy_service`. `copy` is fairly generic, particularly if it's going to be an element ID.

app/main/views/services.py
galund: Yep. Maybe only set `form` if `lot.get('oneServiceLimit')`? I know it's not strictly necessary, and constructing that thing won't really impact performance - but it might make it more clear what's going on? You'd have to be careful about how you're passing the errors to the template if you did this.

I'm happy with one template if you're sure you want to stick with one view.

tests/app/main/test_services.py
```diff
@@ -2327,7 +2328,37 @@ def test_lists_correct_services_for_previous_framework_and_lot(self, get_metadat
...
+            'services': [{'serviceName': 'Service one', 'copiedToFollowingFramework': False}],
+        }
+
+        res = self.client.get(
+            f'/suppliers/frameworks/digital-outcomes-and-specialists-3/submissions/{lot_slug}/previous-services'
+        )
+        doc = html.fromstring(res.get_data(as_text=True))
+
+        assert res.status_code == 200
+        assert doc.xpath(f"//h1[normalize-space()='Do you want to reuse your previous {lot_name} service?']")
```
galund: Do we really have to have all our content duplicated into all our tests? No big deal, I know we do it everywhere. But I'd prefer to test "does the `h1` include `lot_name` or something like that.

tests/app/main/test_services.py
```diff
@@ -2460,6 +2491,113 @@ def test_shows_service_warning_in_correct_conditions(
...
+            'submissions/digital-specialists/previous-services',
+            data={'service_id': '0001'},
+        )
+        doc = html.fromstring(res.get_data(as_text=True))
+
+        assert res.status_code == 200
+        assert "Do you want to reuse your previous digital specialists service?" == doc.xpath(
+            "//a[@class='validation-masthead-link']/text()"
+        )[0].strip()
+        assert "You must answer this question." == doc.xpath(
```
galund: If "you must answer this question" is a constant somewhere, then it'd be better to reference that constant.

tests/app/main/test_services.py
Wynndow: Yeah, I guess it might be a bit redundant. Just following what we do elsewhere.

app/main/views/services.py
Wynndow: Cool, have updated to only load the form if necessary.

tests/app/main/test_services.py
Wynndow: It's not really, it's defined in the form as the validation message for that particular field. Not sure how easy it would be to import.

app/main/forms/frameworks.py
Wynndow: Yep, done.

app/main/forms/frameworks.py
```diff
@@ -50,3 +50,16 @@ class ReuseDeclarationForm(FlaskForm):
...
+
+class OneServiceLimitCopyServiceForm(FlaskForm):
+
+    def __init__(self, lot_name, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.copy_service.label.text = f"Do you want to reuse your previous {lot_name} service?"
+
+    copy_service = BooleanField(
+        false_values={'False', 'false', ''},
+        validators=[InputRequired(message='You must answer this question.')]
```
samuelhwilliams: Can we add a default label here that will get overridden later? Maybe just `Do you want to reuse your previous service?`

Wynndow: For sure

tests/app/main/helpers/test_services.py
```diff
@@ -0,0 +1,110 @@
...
+        current_user.name = 'Mr Sausages'
+        current_user.supplier_id = 1234
+
+        self.api_client_mock.get_service.return_value = self.previous_service()
+
+        copy_service_from_previous_framework(
+            self.api_client_mock, self.content_loader_mock, 'dos-cloud-X', 'digital-sausages', 4444244
+        )
+
+        assert self.content_loader_mock.get_metadata.call_args_list == [
```
samuelhwilliams: I guess this is technically an implementation detail, so it probably doesn't need asserting. As long as we end up calling the api client with the correct args, it should be fine, I think?

tests/app/main/helpers/test_services.py
```diff
@@ -0,0 +1,110 @@
...
+            mock.call('dos-cloud-X', 'copy_services', 'questions_to_copy'),
+            mock.call('dos-cloud-X', 'copy_services', 'source_framework'),
+        ]
+
+        assert self.api_client_mock.get_service.call_args_list == [mock.call(4444244)]
```
samuelhwilliams: Hmm, not sure I understand why we need to assert that this call happened?

tests/app/main/test_services.py
```diff
@@ -2400,7 +2431,7 @@ def test_shows_confirmation_banner_and_button_when_copying_all(self, get_metadat
             "[normalize-space()='Are you sure you want to add all your cloud hosting services?']"
         )
         assert doc.xpath(
-            "//button[@class='button-save banner-action'][normalize-space()='Yes, add all services']"
+            "//button[@class='button-save banner-action'][normalize-space()='Yes, add']"
```
samuelhwilliams: `Yes, add` feels like weird content... Could it just be `Yes`?

app/main/views/services.py
```diff
@@ -715,14 +722,14 @@ def set_remove_last_subsection_error_message():
...
-def list_previous_services(framework_slug, lot_slug):
+def previous_services(framework_slug, lot_slug):
     framework, lot = get_framework_and_lot_or_404(data_api_client, framework_slug, lot_slug)
     if framework['status'] != 'open':
         abort(404)
 
-    copy_all = request.args.get('copy_all', None)
+    form = OneServiceLimitCopyServiceForm(lot['name'].lower()) if lot.get('oneServiceLimit') else None
     source_framework_slug = content_loader.get_metadata(framework['slug'], 'copy_services', 'source_framework')
     source_framework = get_framework_or_404(data_api_client, source_framework_slug)
```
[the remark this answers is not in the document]
Wynndow: Good point. Have updated, here and also in the `framework_submission_services` view.

tests/app/main/helpers/test_services.py
Wynndow: Yep, have removed, and the assertation on the api client `get_service` call.

tests/app/main/helpers/test_services.py
Wynndow: We do not. I've removed.

tests/app/main/test_services.py
Wynndow: Going to talk to Jane about this as well as some other content stuff this afternoon.

samuelhwilliams: Cool. Drop me a line when that's all updated and I'll 👍

app/main/views/frameworks.py
```diff
@@ -271,7 +271,15 @@ def framework_submission_services(framework_slug, lot_slug):
         previous_framework = {}
 
     if previous_framework_slug:
-        previous_framework = get_framework_or_404(data_api_client, previous_framework_slug)
+        try:
+            previous_framework = data_api_client.get_framework(previous_framework_slug)['frameworks']
+        except HTTPError as e:
+            current_app.logger.error(
```
samuelhwilliams: Unlikely as it is, presumably this could include some kind of transport errors or other failures - so a check that this error is a 404 would be robust.

app/main/views/services.py
```diff
@@ -715,41 +722,80 @@ def set_remove_last_subsection_error_message():
...
         abort(404)
 
-    copy_all = request.args.get('copy_all', None)
+    form = OneServiceLimitCopyServiceForm(lot['name'].lower()) if lot.get('oneServiceLimit') else None
     source_framework_slug = content_loader.get_metadata(framework['slug'], 'copy_services', 'source_framework')
-    source_framework = get_framework_or_404(data_api_client, source_framework_slug)
 
-    previous_services = data_api_client.find_services(
+    try:
+        source_framework = data_api_client.get_framework(source_framework_slug)['frameworks']
```
samuelhwilliams: Maybe a `get_framework_or_500`?

Would feel like a weird method name but could be explained by a docstring and would remove this duplication.

app/templates/services/previous_services.html
```diff
@@ -35,62 +35,124 @@
...
+      <form action="{{ url_for('.copy_all_previous_services', framework_slug=framework.slug, lot_slug=lot.slug) }}" method="POST">
+        <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
+        {%
+          with
+          message = "Are you sure you want to add all your {} services?".format(lot.name|lower),
+          type = "success",
+          action = '<button type="submit" class="button-save banner-action" data-analytics="trackEvent"
+            data-analytics-category="Copy services"
+            data-analytics-action="Copy all"
+            data-analytics-label="Count: {}">Yes, add all services'.format(previous_services_still_to_copy|length)|safe
```
samuelhwilliams: Should this content match whatever we change the other label to?

tests/app/main/helpers/test_services.py
```diff
@@ -0,0 +1,104 @@
...
+
+class TestCopyServiceFromPreviousFramework():
+    def setup_method(self):
+        self.abort_patch = mock.patch('app.main.helpers.services.abort')
+        self.abort = self.abort_patch.start()
+        self.abort.side_effect = CustomAbortException()
+
+        self.get_supplier_framework_info_patch = mock.patch('app.main.helpers.services.get_supplier_framework_info')
+        self.get_supplier_framework_info = self.get_supplier_framework_info_patch.start()
+        self.get_supplier_framework_info.return_value = True
```
samuelhwilliams: Returns True?

tests/app/main/helpers/test_services.py
```diff
@@ -0,0 +1,104 @@
...
+                {
+                    'targetFramework': 'dos-cloud-X',
+                    'status': 'not-submitted',
+                    'questionsToCopy': ['list', 'of', 'questions']
+                },
+            )
+        ]
+
+    def test_404s_if_no_supplier_framework_info(self):
+        self.get_supplier_framework_info.return_value = False
```
samuelhwilliams: Feels like these values could be more realistic.

tests/app/main/test_services.py
samuelhwilliams: You could define the message as a class variable for the form and then use that here easily enough with something like `OneServiceLimitCopyServiceForm.ANSWER_REQUIRED_MESSAGE`.

app/templates/services/previous_services.html
Wynndow: The link text you'd click to show this banner? That'll say 'Add all your services?'

tests/app/main/helpers/test_services.py
Wynndow: Yeah, I'm just mocking out the guard clause. It just checks the truthyness off `get_supplier_framework_info`, doesn't actually do anything with the result. Do you think I should be returning an actual supplier framework?

tests/app/main/helpers/test_services.py
Wynndow: Again, as above. Do you think there's value in providing a supplier record? Or not providing one in this case.

Wynndow:
I dunno, feels a bit weird to me to add a class variable just to test it in that way?

app/templates/services/previous_services.html
samuelhwilliams: I specifically meant the analytics label being consistent with the displayed value. But maybe on second thought it should stay the same, to remain backwards compatible with past data.

tests/app/main/helpers/test_services.py
samuelhwilliams: I think we should keep it as realistic as possible, yeah. If we don't have an api stub for a supplier framework and there's no easy thing to use from the supplier-fe test fixtures already, feel free to add a stub for it. ;)

tests/app/main/helpers/test_services.py
samuelhwilliams: Yeah I think we should mock things out realistically as much as possible.

samuelhwilliams:
I don't think it's unreasonable for most text to be defined and used as variables rather than injected straight into things. Particularly for generic error messages like this that will clearly be used for validation across multiple fields/forms.

PR #887, opened 2018-06-04
Changes supplier frontend to use generic templates

DilwoarH:


tests/app/main/test_notifications.py
```diff
@@ -12,24 +12,6 @@ def teardown_method(self, method):
...
-        cookie_value = self.get_cookie_by_name(res, 'seen_user_research_message')
-        assert cookie_value is None
-
-    def test_should_not_see_banner_if_subscribed(self):
-        self.login(user_research_opted_in=True)
-        res = self.client.get('/suppliers')
-        assert res.status_code == 200
-        assert 'Help us improve the Digital Marketplace' not in res.get_data(as_text=True)
-        assert 'Sign up to be a potential user research participant' not in res.get_data(as_text=True)
-        assert 'class="user-research-banner-close-btn"' not in res.get_data(as_text=True)
```
katstevens: Shouldn't the new base page contain these links?

lfdebrux: It does, which is why I took these tests out, as I felt the behaviour should be tested in the frontend toolkit and functional tests, not here.

lfdebrux: https://github.com/alphagov/digitalmarketplace-functional-tests/blob/master/features/user/user_research.feature

katstevens: Makes sense 👍

PR #888, opened 2018-06-05
Delete PERMANENT_SESSION_LIFETIME variable from config

CrystalPea:
We delete PERMANENT_SESSION_LIFETIME variable from config, as the one in user-frontend is enough for all frontends. The session lifetime in user-frontend has been reduced from 4 hours to 1 hour.

Trello ticket: https://trello.com/c/KvdDwvnk/400-inadequate-session-timeout


CrystalPea:
@katstevens pull request updated ☺️

PR #889, opened 2018-06-06
Add error page for 400s

Wynndow:
Unlike the buyer or user frontends, the supplier frontend didn't have a
custom error page, or even any error handling at all, for 400s. This
meant that any time `abort(400)` was being called the user would get
a janky horrible error page.

This allows a custom error message to also be passed through which wil
be displayed to the user.

### Before
screen shot 2018-06-06 at 14 32 15

### After
screen shot 2018-06-06 at 14 29 21

### DIFF
--- a/app/templates/errors/400.html
+++ b/app/templates/errors/400.html
@@ -1,0 +1,20 @@
+{% extends "_base_page.html" %}
+
+{% block page_title %}Bad request - Digital Marketplace{% endblock %}
+
+{% block main_content %}
+
+<div class="error-page">
+    <header class="page-heading-smaller">
+      <h1>Sorry, there was a problem with your request</h1>
+    </header>
+    <p>
+      {% if error_message %}
+        {{ error_message }}
+      {% else %}
+        Please do not attempt the same request again.
+      {% endif %}
+    </p>
+</div>
+
+{% endblock %}
--- a/tests/app/main/test_services.py
+++ b/tests/app/main/test_services.py
@@ -11,6 +11,7 @@ import mock
 import pytest
 
 from dmapiclient import HTTPError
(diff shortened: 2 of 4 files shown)


tests/app/main/test_services.py
```diff
@@ -2557,20 +2558,41 @@ def test_400s_if_lot_does_not_have_one_service_limit(self):
...
+        (
+            ('Digital outcomes', 'digital-outcomes'),
+            ('Digital Specialists', 'digital-specialists'),
+            ('User research participants', 'user-research-participants'),
+        ),
+    )
+    def test_400s_if_draft_already_exists(self, lot_name, lot_slug):
+        self.data_api_client.get_framework.return_value = framework(
+            slug='digital-outcomes-and-specialists-3',
+            lots=[{
```
samuelhwilliams: Might as well use a `lot` stub from `api_stubs` here as well.

app/templates/errors/400.html
```diff
@@ -0,0 +1,20 @@
...
+
+<div class="error-page">
+    <header class="page-heading-smaller">
+      <h1>Sorry, there was a problem with your request</h1>
+    </header>
+    <p>
+      {% if error_message %}
+        {{ error_message }}
+      {% else %}
+        Please try again.
```
samuelhwilliams: This is probably bad content (I know/suspect you copied this from user-frontend, so 🤷‍♂️ ).

The buyer-frontend 400 template says "Please do not attempt the same request again.", which is more sensible - as a bad request shouldn't be repeated without changing something in the request.

Really should centralise these error templates one day...

Wynndow: Done and done. Will stick a PR in for the user-frontend too.

PR #894, opened 2018-06-15
Refactor Supplier FE tests (part 1)

katstevens:
Part of the ongoing work to refactor the front end tests.

- Remove spurious `with self.app.app_context` context managers
- Move `data_api_client` patching to test setup methods (and shared return values for some of its methods, where sensible), to be consistent with our other test classes
- Minor whitespace tidying

Part 1 covers:
- `test_frameworks`
- `test_login`
- `test_notifications`
- `test_users`

Part 2 will cover:
- `test_services`
- `test_suppliers`
- Helper/validation functions.

Recommend ticking the 'hide whitespace changes' when reviewing!

### DIFF
--- a/tests/app/main/test_frameworks.py
+++ b/tests/app/main/test_frameworks.py
@@ -3686,7 +3686,9 @@ class TestFrameworkUpdatesPage(BaseApplicationTest):
         assert response.status_code == 200
         assert u'Ask a question about your G-Cloud 7 application' not in data
 
-
+@mock.patch('dmutils.s3.S3')
+@mock.patch('app.main.views.frameworks.DMNotifyClient.send_email', autospec=True)
+@mock.patch('app.main.views.frameworks.mandrill_send_email')
 class TestSendClarificationQuestionEmail(BaseApplicationTest):
 
     def setup_method(self, method):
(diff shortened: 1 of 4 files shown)


tests/app/main/test_frameworks.py
```diff
@@ -4986,15 +5049,15 @@ def test_framework_agreement_returned_having_signed_contract_variation_redirects
     @mock.patch('dmutils.s3.S3')
```
benvand: Looks like these 2 patches have already been added to the class, hence the failing test

tests/app/main/test_frameworks.py
```diff
@@ -1823,11 +1855,10 @@ def test_page_returns_404_if_framework_in_wrong_state(self, data_api_client, man
 
         assert res.status_code == 404
 
-    def test_page_returns_404_if_supplier_not_on_framework(self, data_api_client, mandrill_send_email, s3):
+    def test_page_returns_404_if_supplier_not_on_framework(self, mandrill_send_email, s3):
         self.login()
 
-        data_api_client.get_framework.return_value = self.framework(status='standstill')
```
benvand: Nice catch that this was in every test on this class

tests/app/main/test_frameworks.py
```diff
@@ -2675,16 +2732,8 @@ def _setup_data_api_client(self, data_api_client, framework_status, framework_sl
...
     )
-    def test_display_open(
-        self,
-        data_api_client,
-        framework_slug,
-        declaration,
-        decl_valid,
-        prefill_fw_slug,
-        expected_sections,
-    ):
```
benvand: Another nice catch, do you have a code cleanup plugin for this or just caught it by site?

tests/app/main/test_frameworks.py
katstevens: Oops - I think that's leftover from a rebase conflict.

katstevens:
Agreed that the setup/teardown introduces a fair bit of duplication at present - the next step is to put all of this into the `BaseApplicationTest` class in a sensible way.

tests/app/main/test_frameworks.py
katstevens: Now fixed (fingers crossed!)

benvand:
Seen the changes, have another 👍

PR #902, opened 2018-06-26
Confirm company details every application

samuelhwilliams:
## Summary
Suppliers need to confirm that the company details they've declared in
their account remain accurate over time as they apply to new iterations
of frameworks on the Digital Marketplace.

When suppliers first set up their account, they confirm that some
immutable attributes are correct: registered name, company registration
number, VAT number and DUNS number. Once these have been confirmed as
correct, they are locked down in the account and cannot be edited
without talking to support. Generally, these will only be updated by
support if there has been a mistake in entering the details. If the
details have legitimately changed, the supplier will need to create a
new supplier account on the Marketplace to complete their application.

After confirming their details are correct at the account level, they
must also confirm the details are correct at the framework application
level. Until they do this, they cannot progress their application
(complete the declaration or add services). This is to prevent the
supplier from putting time into these labour-intensive tasks before
realising that their supplier account no longer accurately represents
their company.

 ## Notes
It might be easiest to review this on a per-commit basis, as some of them are refactoring/small tweaks which dilute the intention of the ticket.

 ## Ticket
https://trello.com/c/V0FSdOCJ/188

### DIFF
--- a/app/main/helpers/frameworks.py
+++ b/app/main/helpers/frameworks.py
@@ -1,10 +1,11 @@
 # -*- coding: utf-8 -*-
 from datetime import datetime
+from functools import wraps
 from itertools import chain, islice, groupby
 import re
 
 from flask import abort
-from flask_login import current_user
+from flask_login import current_user, current_app
 
 from dmapiclient import APIError, HTTPError
 from dmutils.formats import DATETIME_FORMAT
@@ -367,3 +368,60 @@ def get_supplier_registered_name_from_declaration(declaration):
         declaration.get('supplierRegisteredName')  # G-Cloud 10 and later declaration key
         or declaration.get('nameOfOrganisation')  # G-Cloud 9 and earlier declaration key
     )
+
+
+def get_unconfirmed_open_supplier_frameworks(data_api_client, supplier_id):
+    frameworks = data_api_client.find_frameworks().get('frameworks')
+    open_framework_slugs = [framework['slug'] for framework in frameworks if framework['status'] == 'open']
+    unconfirmed_open_supplier_frameworks = [
+        sf for sf in
+        data_api_client.get_supplier_frameworks(supplier_id=supplier_id)['frameworkInterest']
+        if sf['frameworkSlug'] in open_framework_slugs and sf['applicationCompanyDetailsConfirmed'] is False
+    ]
+
+    framework_slug_map = {framework['slug']: framework for framework in frameworks}
+    for fw in unconfirmed_open_supplier_frameworks:
+        if 'frameworkName' not in fw:
+            fw['frameworkName'] = framework_slug_map[fw['frameworkSlug']]['name']
+
+    return unconfirmed_open_supplier_frameworks
+
+
+class EnsureApplicationCompanyDetailsHaveBeenConfirmed:
+    """A decorator for framework application views that should not be accessible before company details have
+    been confirmed for that application.
+
+    This is a class-based decorator primarily so that it's possible to mock out the validator for the majority of tests
+    which aren't actively testing that the view requires suppliers to be in a certain application state."""
+
+    def __init__(self, data_api_client):
+        self.data_api_client = data_api_client
+
+    def __call__(self, func):
+        @wraps(func)
+        def decorated_view(*args, **kwargs):
+            if self.validator(*args, **kwargs):
+                return func(*args, **kwargs)
+
+            # Shouldn't be able to get here
+            abort(500, "There was a problem accessing this page of your application. Please try again later.")
+
+        return decorated_view
+
+    def validator(self, *args, **kwargs):
+        """Performs the actual validation that ensures the logged-in supplier has confirmed their company details
+        are correct for this application"""
+        if 'framework_slug' not in kwargs:
+            current_app.logger.error("Required parameter `framework_slug` is undefined for the calling view.")
+            abort(500, "There was a problem accessing this page of your application. Please try again later.")
+
+        if current_user.is_authenticated() and current_user.supplier_id:
+            supplier_framework = self.data_api_client.get_supplier_framework_info(
+                current_user.supplier_id, kwargs['framework_slug']
+            )['frameworkInterest']
+
+            if supplier_framework['applicationCompanyDetailsConfirmed'] is False:
+                return abort(400, "You cannot access this part of your application until you have confirmed your "
+                                  "company details.")
+
+        return True
--- a/app/main/helpers/frameworks.py
+++ b/app/main/helpers/frameworks.py
@@ -215,7 +215,6 @@ def get_statuses_for_lot(
     unit,
     unit_plural
 ):
-
     if not drafts_count and not complete_drafts_count:
         return []
 
@@ -283,7 +282,6 @@ def get_status_for_one_service_lot(
 def get_status_for_multi_service_lot_and_service_type(
     count, services_status, framework_is_open, declaration_complete, unit, unit_plural
 ):
-
     singular = (1 == count)
     description_of_services = u'{} {} {}'.format(
         count, services_status, unit if singular else unit_plural
@@ -365,7 +363,7 @@ def get_frameworks_closed_and_open_for_applications(frameworks):
 
 
 def get_supplier_registered_name_from_declaration(declaration):
-    return(
+    return (
         declaration.get('supplierRegisteredName')  # G-Cloud 10 and later declaration key
         or declaration.get('nameOfOrganisation')  # G-Cloud 9 and earlier declaration key
     )
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -30,12 +30,22 @@ from ... import data_api_client
 from ...main import main, content_loader
 from ..helpers import login_required
 from ..helpers.frameworks import (
-    get_declaration_status, get_last_modified_from_first_matching_file, register_interest_in_framework,
-    get_supplier_on_framework_from_info, get_declaration_status_from_info, get_supplier_framework_info,
-    get_framework_or_404, get_framework_and_lot_or_404, get_framework_or_500, count_drafts_by_lot, get_statuses_for_lot,
-    return_supplier_framework_info_if_on_framework_or_abort, returned_agreement_email_recipients,
-    check_agreement_is_related_to_supplier_framework_or_abort, get_framework_for_reuse,
-    get_supplier_registered_name_from_declaration
+    check_agreement_is_related_to_supplier_framework_or_abort,
+    count_drafts_by_lot,
+    get_declaration_status,
+    get_declaration_status_from_info,
+    get_framework_and_lot_or_404,
+    get_framework_for_reuse,
+    get_framework_or_404,
+    get_framework_or_500,
+    get_last_modified_from_first_matching_file,
+    get_statuses_for_lot,
+    get_supplier_framework_info,
+    get_supplier_on_framework_from_info,
+    get_supplier_registered_name_from_declaration,
+    register_interest_in_framework,
+    return_supplier_framework_info_if_on_framework_or_abort,
+    returned_agreement_email_recipients,
 )
 from ..helpers.services import (
     count_unanswered_questions,
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -32,6 +32,7 @@ from ..helpers import login_required
 from ..helpers.frameworks import (
     check_agreement_is_related_to_supplier_framework_or_abort,
     count_drafts_by_lot,
+    EnsureApplicationCompanyDetailsHaveBeenConfirmed,
     get_declaration_status,
     get_declaration_status_from_info,
     get_framework_and_lot_or_404,
@@ -53,6 +54,7 @@ from ..helpers.services import (
     get_lot_drafts,
     get_signed_document_url,
 )
+from ..helpers.suppliers import supplier_company_details_are_complete
 from ..helpers.validation import get_validator
 from ..forms.frameworks import SignerDetailsForm, ContractReviewForm, AcceptAgreementVariationForm, ReuseDeclarationForm
 
@@ -117,7 +119,16 @@ def framework_dashboard(framework_slug):
     if declaration_status == 'unstarted' and framework['status'] == 'live':
         abort(404)
 
-    application_made = supplier_is_on_framework or (len(complete_drafts) > 0 and declaration_status == 'complete')
+    application_company_details_confirmed = (
+        supplier_framework_info and supplier_framework_info['applicationCompanyDetailsConfirmed']
+    )
+    application_made = (
+        supplier_is_on_framework or (
+            len(complete_drafts) > 0
+            and declaration_status == 'complete'
+            and application_company_details_confirmed
+        )
+    )
     lots_with_completed_drafts = [lot for lot in framework['lots'] if count_drafts_by_lot(complete_drafts, lot['slug'])]
 
     try:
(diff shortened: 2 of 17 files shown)


samuelhwilliams:
Still need to write at least these tests:
* confirm_supplier_details
    * all/only open framework applications are confirmed
* supplier dashboard:
    *  `session.currently_applying_to` is cleared on visit
*  supplier_details:
    * supplier unconfirmed + application unconfirmed get confirmed at the same time
    * supplier confirmed + open unconfirmed application shows save button

But this is mostly done now, so please feel free to start reviewing

app/templates/suppliers/details.html
```diff
@@ -182,12 +178,13 @@
...
   </div>
 
-  {% if not supplier.companyDetailsConfirmed %}
+  {% if supplier_company_details_confirmed == False %}
     {% if company_details_complete  %}
     <div class='grid-row'>
       <div class='column-two-thirds'>
         <form method="POST" action="{{ url_for('.confirm_supplier_details') }}">
           <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
+          {% if supplier_company_details_confirmed == False %}
```
Wynndow: This seems extra?

Wynndow: Bah. I'm only looking at one commit. Ignore.

app/main/helpers/frameworks.py
```diff
@@ -365,7 +364,47 @@ def status_priority(status):
...
+
+            # Shouldn't be able to get here
+            abort(500, "There was a problem accessing this page of your application. Please try again later.")
+
+        return decorated_view
+
+    def validator(self, *args, **kwargs):
+        """Performs the actual validation that ensures the logged-in supplier has confirmed their company details
+        are correct for this application"""
+        if 'framework_slug' not in kwargs:
```
Wynndow: Is it just luck that all the views we need to decorate take `framework_slug`? This couldn't be used on a view that didn't take it (if we needed to in future?)

app/main/helpers/frameworks.py
```diff
@@ -365,7 +364,47 @@ def status_priority(status):
...
+            current_app.logger.error("Required parameter `framework_slug` is undefined for the calling view.")
+            abort(500, "There was a problem accessing this page of your application. Please try again later.")
+
+        if current_user.is_authenticated() and current_user.supplier_id:
```
Wynndow: Do we need to check they're authenticated? If they aren't then the validation will pass anyway? And I'm guessing this needs to be used on routes that have the `@login_required` decorator?

app/main/helpers/frameworks.py
samuelhwilliams: It's not luck, per se - the decorator is about locking them out of a part of their application to a framework, so it seems logical to me that we need to know _which_ framework they're applying to - so we should always have a reference to the framework slug.

app/main/helpers/frameworks.py
samuelhwilliams: Hmm, maybe we don't need to check if they're authenticated. I didn't really check how the user object works - I thought we might not be able to access supplier_id without an authenticated users, and the decorator theoretically could be called before login_required.

Wynndow: Yeah, I just checked and supplier_id throws an attribute error on an anonymous user. so the `is_authenticated` check is necessary in case this decorator is called before `login_required`.

app/main/views/frameworks.py
```diff
@@ -30,19 +30,31 @@
...
+    get_framework_for_reuse,
+    get_framework_or_404,
+    get_framework_or_500,
+    get_last_modified_from_first_matching_file,
+    get_statuses_for_lot,
+    get_supplier_framework_info,
+    get_supplier_on_framework_from_info,
+    get_supplier_registered_name_from_declaration,
+    register_interest_in_framework,
+    EnsureApplicationCompanyDetailsHaveBeenConfirmed,
```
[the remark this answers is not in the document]
samuelhwilliams: Case sensitive line sorting is a mistake :(

app/main/views/suppliers.py
```diff
@@ -82,6 +82,9 @@ def dashboard():
             )
         })
 
+    if "currently_applying_to" in session:
```
Wynndow: Why remove this here?

samuelhwilliams: If the supplier is on the account dashboard and goes to their company details, they aren't coming "from" an application - so they shouldn't get a link on the page which says "go back to the application".

app/main/views/suppliers.py
```diff
@@ -107,16 +110,36 @@ def supplier_details():
...
+    open_framework_slugs = [framework['slug'] for framework in frameworks if framework['status'] == 'open']
+
+    unconfirmed_open_supplier_frameworks, unconfirmed_open_supplier_framework_names = [], []
+    if open_framework_slugs:
+        unconfirmed_open_supplier_frameworks = list(filter(
+            lambda sf: sf['frameworkSlug'] in open_framework_slugs and not sf['applicationCompanyDetailsConfirmed'],
+            data_api_client.get_supplier_frameworks(supplier_id=current_user.supplier_id)['frameworkInterest']
+        ))
+        unconfirmed_open_supplier_framework_names = [
+            framework_slug_map[sf['frameworkSlug']]['name'] for sf in unconfirmed_open_supplier_frameworks
```
Wynndow: I may be being thick, but could you not just do `[sf['name'] for sf in unconfirmed_open_supplier_frameworks]` rather than going through the `framework_slug_map `?

app/main/views/suppliers.py
```diff
@@ -107,16 +110,36 @@ def supplier_details():
...
+        ]
+
+    currently_applying_to_framework = None
+    if "currently_applying_to" in session and session["currently_applying_to"] in framework_slug_map:
```
Wynndow: Is the second conditional here to make sure the user hasn't snuck a dodgy framework slug into their session somehow?

app/main/views/suppliers.py
```diff
@@ -130,6 +153,13 @@ def confirm_supplier_details():
     except APIError as e:
         abort(e.status_code)
 
+    frameworks = data_api_client.find_frameworks().get('frameworks')
+    open_framework_slugs = [framework['slug'] for framework in frameworks if framework['status'] == 'open']
+    unconfirmed_open_supplier_frameworks = filter(
```
Wynndow: worth making this a helper?

app/main/views/suppliers.py
samuelhwilliams: That should be impossible (without discovering our secret key). It's more just a belt and braces approach for if we somehow set an invalid framework slug. It's 99% unnecessary, but given how simple a check it is, still vaguely worth doing.

app/main/views/suppliers.py
samuelhwilliams: I really wanted to do that, but `unconfirmed_open_supplier_frameworks` is a list of SupplierFramework objects, which don't serialize the framework name back out - just the slug and family. So it's not available there sadly!

app/main/views/suppliers.py
samuelhwilliams: Maybe. It's used twice so I get what you mean. Do you think it's worth it?

app/templates/frameworks/_framework_actions.html
```diff
@@ -27,17 +27,11 @@
...
-  <div class="browse-list-item-status-angry">
-    <p class="browse-list-item-status-title">
-    <strong>No services will be submitted because you haven’t completed your company details</strong>
-    </p>
-  </div>
-  {% elif not company_details_complete %}
+  {% if not application_company_details_confirmed %}
   <div class="browse-list-item-status-quiet">
-    <p class="browse-list-item-status-title">You must add your organisation’s details</p>
+    <p class="browse-list-item-status-title">You must {% if not supplier_company_details_complete %}add{% else %}confirm{% endif %} your organisation’s details</p>
```
Wynndow: This might be clearer as `{% if supplier_company_details_complete %}confirm{% else %}add{% endif %}`

app/main/views/suppliers.py
Wynndow: Yep, that makes sense.

app/main/views/suppliers.py
Wynndow: Yeah, i think so. Having the filter and the lambda with two conditionals in the view took me a little while to parse. I think it would be cleaner to shift them out. Also one returns a filter object, the other is cast to a list and I spent a bit of time wondering why.

app/templates/frameworks/_framework_actions.html
samuelhwilliams: Yep that's fair enough 👍

app/main/views/suppliers.py
samuelhwilliams: Ah fair enough 👍

app/templates/frameworks/_framework_actions.html
```diff
@@ -27,17 +27,11 @@
 <li class="browse-list-item">
   <a class="browse-list-item-link"
      href="{{ url_for('.supplier_details') }}">
-    <span>Enter your company details</span>
+    <h2>1. Enter your company details</h2>
```
Wynndow: The indentations here seems a bit off. For example lines 27, 37, 38...

samuelhwilliams: Our indentations are generally screwy within the templates. Turn off whitespace diffs might help.

Wynndow: It's the same in the browser. It's not that straightforward to fix up though...

app/main/views/suppliers.py
samuelhwilliams: Maybe.

Maybe they're different enough that it doesn't make sense.

app/templates/suppliers/details.html
```diff
@@ -22,11 +22,7 @@
 
 {% block main_content %}
   <div class='grid-row'>
-  {% if currently_applying_to or (company_details_complete and not supplier.companyDetailsConfirmed) %}
-    <div class='column-one-whole padding-bottom-small'>
-  {% else %}
-    <div class='column-one-whole'>
-  {% endif %}
+    <div class='column-one-whole{% if currently_applying_to or (supplier_company_details_complete and not supplier_company_details_confirmed) %} padding-bottom-small{% endif %}'>
```
Wynndow: 👍

application.py
```diff
@@ -13,6 +16,9 @@
     ]
 }
 
+# Sadly can't enable this yet, but we should work towards it...
+# application.jinja_env.undefined = jinja2.StrictUndefined
```
Wynndow: What's preventing this?

app/templates/frameworks/_framework_actions.html
samuelhwilliams: Oh hmm. I'll take another look at it in the browser. I thought it looked fine when I did it 🤔  Maybe I didn't look closely enough.

samuelhwilliams: Do you mean visually something on the page looks off, or in the source code the elements are indented a different amount? If it's the latter, that's not something we really seem to care about (certainly from what I've seen all over the rest of the code base)

application.py
samuelhwilliams: Ah I made a tech debt ticket for this.

Basically our frontend toolkit doesn't use Jinja's `is defined` check, it instead just does `{% if some_var %}` (rather than eg `{% if some_var is defined %}`). Without strict undefined, the first test compiles but evaluates to False. With strict undefined, the first won't compile, so we need to change all of the checks like that to use the defined test.

I think it's worth doing but it's way too big a thing to do for now.

app/templates/frameworks/_framework_actions.html
Wynndow: Sorry, wrong word. I mean in the editor (as in turning off whitespace won't help). Don't worry about it, it's not a big deal.

tests/app/main/helpers/test_frameworks.py
```diff
@@ -638,3 +641,131 @@ def test_calls_logger_if_provided(self, abort):
...
+
+
+class TestEnsureApplicationCompanyDetailsHaveBeenConfirmed(BaseApplicationTest):
+    def setup_method(self, method):
+        super().setup_method(method)
+
+        from app import data_api_client
+        self.data_api_client_mock = mock.Mock(spec_set=data_api_client)
+        self.decorator = EnsureApplicationCompanyDetailsHaveBeenConfirmed(self.data_api_client_mock)
+        # self.login()
```
Wynndow: Can this go?

Wynndow: Also, I don't see `self.decorator` being used anywhere else. Does it need to be part of the setup like this?

tests/app/main/helpers/test_frameworks.py
```diff
@@ -638,3 +641,131 @@ def test_calls_logger_if_provided(self, abort):
...
+            application_company_details_confirmed=True
+        )
+
+        with mock.patch('app.main.helpers.frameworks.current_user') as current_user_patch:
+            current_user_patch.is_authenticated.return_value = True
+            current_user_patch.supplier_id.return_value = 1
+
+            assert decorator.validator(framework_slug='g-cloud-10') is True
+
+    def test_validator_can_be_reassigned(self):
```
Wynndow: This feels like you're just testing the mock library.

tests/app/main/helpers/test_frameworks.py
```diff
@@ -638,3 +641,131 @@ def test_calls_logger_if_provided(self, abort):
...
+        def raise_successful_override():
+            return 'overriden'
+
+        decorator = EnsureApplicationCompanyDetailsHaveBeenConfirmed(self.data_api_client_mock)
+
+        with mock.patch.object(decorator, 'validator', raise_successful_override):
+            assert decorator.validator() == 'overriden'
+
+    def test_only_known_decorated_views_are_protected(self):
```
Wynndow: I like this. My only concern was the overhead of having to call every view. But with (currently) 85 views it doesn't seem to be bad at all.

tests/app/main/test_frameworks.py
```diff
@@ -1554,48 +1558,37 @@ def test_make_or_edit_supplier_declaration_shows_correct_page(
...
+            application_company_details_confirmed=True,
+            **supplier_framework_kwargs,
         )
 
         response = self.client.get('/suppliers/frameworks/g-cloud-7')
         document = html.fromstring(response.get_data(as_text=True))
 
         assert (
-            document.xpath("//a[normalize-space(string())=$link_label]/@href", link_label=link_label)[0]
+            document.xpath("//a[contains(normalize-space(string()), $link_label)]/@href", link_label=link_label)[0]
```
Wynndow: What does adding `contains` give you?

tests/app/main/helpers/test_frameworks.py
samuelhwilliams: yeah these are probably both hangovers from intermediate work, will remove

tests/app/main/helpers/test_frameworks.py
samuelhwilliams: mmmm yeah it's possibly a slightly more useless test than I anticipated. I'll remove it.

tests/app/main/helpers/test_frameworks.py
samuelhwilliams: 👍

tests/app/main/test_suppliers.py
```diff
@@ -819,18 +869,43 @@ def test_back_to_application_link_not_visible_if_currently_applying_to_not_in_se
...
+            # Details complete and already confirmed
+            (get_supplier(companyDetailsConfirmed=True), False, False),
+            # Details not complete or confirmed
+            (get_supplier(companyDetailsConfirmed=False, vatNumber=None), False, False),
         ]
     )
-    def test_green_button_is_shown_only_when_details_are_complete_but_not_confirmed(
-        self, supplier_details, button_should_be_shown
+    def test_green_button_is_shown_when_details_are_complete_but_not_confirmed(
+        self, supplier_details, open_application, button_should_be_shown
```
Wynndow: The open_application param doesn't seem to be used in the test.

tests/app/main/test_frameworks.py
samuelhwilliams: The labels now start with a tasklist item number, which I'm not really interested in asserting are a specific value here.

tests/app/main/helpers/test_frameworks.py
samuelhwilliams: I guess a slight thought is whether this test should be inverted to list the views which _shouldn't_ be protected explicitly, as that works on the principle that things are protected by default, which is generally more secure. There would be slightly more overhead in that every view which is added to this app (that isn't part of the application process, which is probably 99% of new views) would need to update this test.

Thoughts..?

Wynndow: Interesting. I guess it would guard against an application view being added and not protected. It'd mean maintaining a longer list here in the test (65ish views).

Application stuff is all behind (I think) `/suppliers/frameworks/<framework_slug>`. Could you add some logic to the test that checks that any views behind that URL structure are protected? That way you could go some way to mitigate an application view being added and accidentally not protected, without having to list all the views here?

PR #921, opened 2018-08-07
Include user email in follow up question email

Wynndow:
For [this trello card.](https://trello.com/c/7lr17Qxo)

CCS's software strips the 'from' email address from follow up questions.
They need to be able to reply to them. We can include the users email
address in the body of the template to make sure they have it.

Rendered template before:
screen shot 2018-08-07 at 12 10 46

Rendered template after:
screen shot 2018-08-07 at 12 10 35


risicle:
:gdpr:

PR #922, opened 2018-08-16
Supplier FE error handling

katstevens:
Trello: https://trello.com/c/3iwy5gAy/115-pull-in-error-handling-to-fe-apps

Now with bonus refactoring!

- Pull in utils/toolkit
- Fix backward incompatibilities from utils v42 (`DMStripWhiteSpaceField` etc)
- Remove old templates
- Use new dmutils error handlers and add tests for CSRF handling 
- Handle 400 errors with custom error messages by rendering the error page directly (there's a tech debt to sort this out once we upgrade Flask, see https://github.com/alphagov/digitalmarketplace-user-frontend/pull/82 for further details)
- Remove redundant try/except for `APIError` aborts. APIErrors are handled by the app.errors.api_error_handler. There's no value in calling abort(e) when we could just let the handler
do its job. There are a few remaining try..excepts that actually do some custom behaviour so those can stay.


lfdebrux:
Awesome work @katstevens

app/main/errors.py
```diff
@@ -1,67 +1,21 @@
...
-def bad_request(e):
-    return _render_error_page(400, e.description or None)
-
-
-@main.app_errorhandler(404)
-def page_not_found(e):
-    return _render_error_page(404)
-
-
-@main.app_errorhandler(500)
```
Wynndow: If we're removing these handlers, how does the app handle a 500 now?

katstevens: In `dmutils.flask_init`. We register handlers for a bunch of common status codes.

Wynndow: ah, nice!

app/main/views/services.py
```diff
@@ -766,7 +767,10 @@ def previous_services(framework_slug, lot_slug):
             # Don't copy a service if the lot has a one service limit and the supplier already has a draft for that lot
             drafts, complete_drafts = get_lot_drafts(data_api_client, framework_slug, lot_slug)
             if drafts or complete_drafts:
-                abort(400, f"You already have a draft {lot['name'].lower()} service.")
+                return render_error_page(
```
Wynndow: My understanding of how the error handling works is limited. Why do you need to call render_error_page here? If you left it as it was, wouldn't the error handler registered in flask_init pick this up? You might need to add a `error_message` kwarg to the csrf handler which could be passed through to `render_error_page` there?

katstevens: Yes, exactly - 400 handling is less than ideal as there's a bug in Flask 0.10.x which means CSRFErrors are treated by error handlers as ordinary 400s. To get round this, in `dmutils.errors.csrf_handler` we listen for *all* 400s then send any that aren't CSRFs to the usual 400 template (no error message there yet, unfortunately). 

 We have a tech debt ticket to sort this out once we've upgraded Flask: https://trello.com/c/wb3WqzEY/545-handle-csrferror-exceptions-separately-from-400-errors

katstevens: If we call `abort(400)` the error message will be lost.

PR #923, opened 2018-08-20
Protect application editing routes with decorator

Wynndow:
For [this trello ticket.](https://trello.com/c/Msvn4HAz)

When applications for a framework close some views are no longer
available. This decorator is applied to those routes to provide users
with meaningful feedback and some extra logging for us.

The apiclient that's passed to the outer wrapper of the decorator is a
callable. This is because the decorator is evaluated at module import
time, which makes mocking the apiclient difficult. By providing it as a
callable it's evaluated at runtime and a mock can be patched in.

The custom error template includes content that is in the frameworks
repo. Its content that describes the expected following framework. If
that hasn't been defined, we show a reduced error message.

### Screenshot! 

applications_closed_decorator

### DIFF
--- a/app/main/__init__.py
+++ b/app/main/__init__.py
@@ -45,7 +45,7 @@ content_loader.load_manifest('digital-outcomes-and-specialists-3', 'services', '
 content_loader.load_manifest('digital-outcomes-and-specialists-3', 'services', 'edit_service')
 content_loader.load_manifest('digital-outcomes-and-specialists-3', 'briefs', 'edit_brief')
 content_loader.load_messages('digital-outcomes-and-specialists-3', ['urls'])
-content_loader.load_metadata('digital-outcomes-and-specialists-3', ['copy_services'])
+content_loader.load_metadata('digital-outcomes-and-specialists-3', ['copy_services', 'following_framework'])
 
 
 @main.after_request
--- a/app/main/helpers/frameworks.py
+++ b/app/main/helpers/frameworks.py
@@ -4,17 +4,20 @@ from functools import wraps
 from itertools import chain, islice, groupby
 import re
 
-from flask import abort
+from flask import abort, render_template, request
 from flask_login import current_user, current_app
 
 from dmapiclient import APIError, HTTPError
+from dmutils.dates import update_framework_with_formatted_dates
 from dmutils.formats import DATETIME_FORMAT
+from dmcontent.errors import ContentNotFoundError
+
+from ...main import content_loader
 
 
 def get_framework_or_404(client, framework_slug, allowed_statuses=None):
     if allowed_statuses is None:
         allowed_statuses = ['open', 'pending', 'standstill', 'live']
-
     framework = client.get_framework(framework_slug)['frameworks']
 
     if allowed_statuses and framework['status'] not in allowed_statuses:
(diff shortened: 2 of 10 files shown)


lfdebrux:
@Wynndow what's the advantage of using a blueprint instead of a decorator?

Wynndow:
@lfdebrux that is what I was attempt to find out!

app/main/__init__.py
```diff
@@ -45,14 +51,58 @@
...
 
 @main.after_request
 def add_cache_control(response):
     response.cache_control.no_cache = True
     return response
 
+@protected_applications.after_request
+def add_cache_control(response):
+    response.cache_control.no_cache = True
+    return response
```
katstevens: I think we can stack the decorators on one function definition?

Wynndow:
Personally I prefer the decorator route rather than the blueprint route. It seems a lot cleaner to me.
I think it would be better if the two blueprints were separate areas of the app, but there's a lot of crossover between them during normal user flow. This leads to a lot of confusion and mess when using  `url_for` as you have to know which blueprint the template/view using it is registered to.
Also the preprocessing stuff seems clunky.
What do people thing?

katstevens:
Yeah... looking at all the `url_for` changes I think any advantage (e.g. simplification of the data_api_client use) is outweighed by the confusion over why some views are in one and not the other. Thanks for indulging my curiosity!

Wynndow:
No problem! It was interesting. I'd never used 'g' before. Why did they give it such a silly name though.

app/main/helpers/frameworks.py
```diff
@@ -425,3 +428,46 @@ def validator(self, *args, **kwargs):
...
+
+
+def return_404_if_applications_closed(data_api_client_callable):
+    def real_decorator(func):
+        @wraps(func)
+        def decorated_view(*args, **kwargs):
+            data_api_client = data_api_client_callable()
+            if 'framework_slug' not in kwargs:
+                current_app.logger.error("Required parameter `framework_slug` is undefined for the calling view.")
+                abort(500, "There was a problem accessing this page of your application. Please try again later.")
```
risicle: Are we going to start writing custom error messages for 500s now? If so, has this gone past content?

app/main/helpers/frameworks.py
```diff
@@ -425,3 +428,46 @@ def validator(self, *args, **kwargs):
...
+
+            framework_slug = kwargs['framework_slug']
+            framework = get_framework_or_404(data_api_client, framework_slug)
+
+            if not framework['status'] == 'open':
+                current_app.logger.info(
+                    'Supplier {supplierId} requested "{method} {url}" after {frameworkSlug} applications closed.',
+                    extra={
+                        'supplierId': current_user.supplier_id,
```
risicle: I *think* most people don't bother camel-casing `extra` parameters from python (though I understand why you are) - so actually these *could* end up mis-fitting with other places where people will use e.g. `supplier_id`. But it's not like it's important.

app/main/views/frameworks.py
```diff
@@ -347,6 +348,7 @@ def framework_submission_services(framework_slug, lot_slug):
 @main.route('/frameworks/<framework_slug>/declaration/start', methods=['GET'])
 @login_required
 @EnsureApplicationCompanyDetailsHaveBeenConfirmed(data_api_client)
```
risicle: hmmmmmmmm interesting how the decorator above *doesn't* use the lambda trick. wonder how it handles mocking in tests...

app/main/helpers/frameworks.py
Wynndow: That's a good point. I might just drop the customness. Seems unnecessary/doesn't add anything.

app/main/views/frameworks.py
Wynndow: It mocks out it's `validator` method, which bypasses having to use the apiclient at all.

tests/app/main/helpers/test_frameworks.py
```diff
@@ -758,3 +760,176 @@ def test_only_known_decorated_views_are_protected(self):
...
+
+        response = view_function(framework_slug='g-cloud-7')
+
+        assert response == 'g-cloud-7 - OK!'
+
+    @mock.patch('app.main.views.frameworks.data_api_client')
+    @mock.patch('app.main.views.services.data_api_client')
+    @mock.patch('app.main.helpers.require_login')
+    @mock.patch('app.main.helpers.frameworks.current_user')
+    def test_only_known_decorated_views_are_protected_applicatins(
```
risicle: Typo, but I do slightly prefer calling them "applicatin's"

app/main/helpers/frameworks.py
Wynndow: Just looked, and it's snake case everywhere else. Will change.

tests/app/main/helpers/test_frameworks.py
Wynndow: Ha, I tagged that on the end to differentiate between a different test with the same name and forgot to remote/update it. I'll fix it.

app/main/views/frameworks.py
risicle: Weird. We're really going to have to sort out these apiclients and put a single one on each request context at some point. Or rather the `g` object.

risicle:
Yeah I concur about decorators vs blueprints - the main disadvantage of blueprints seems to be that, of course, they're not composable.

risicle:
I mean that you can't just throw a bunch of blueprints together to gain all their features. Whereas decorators more or less are composable.

It's a bit like the difference between standard inheritance and mixins.

PR #935, opened 2018-09-25
Use Notify to send framework agreement reviewed email

lfdebrux:
This PR closes https://trello.com/c/1hW3ZdfA

As well as changing the email service from Mandrill to Notify, it also changes the previous behaviour where an email failure would result in a 503.

An error will still be logged by the email client, however the user will not be interrupted.

The new template is on Notify, called ["Framework agreement signature page"](https://www.notifications.service.gov.uk/services/95316ff0-e555-462d-a6e7-95d26fbfd091/templates/08929c93-f9e7-4b40-b75f-418659308324), and differs from the old template in one small but significant way; we no longer include the text about framework live date, which was never populated in the Mandrill template.


katstevens:
Just reading through the template again. It kind of implies that they can start selling straight away. Should include a line saying that we'll email them when the framework goes live (we do email them about that, right? 😸)

lfdebrux:
@katstevens I know what you mean, Patrick is going to get the email content reviewed though, so I'm not too worried about it. I might add a line in the old place saying "You should be able to sell services once the framework has opened."

app/main/views/frameworks.py
```diff
@@ -1124,29 +1124,20 @@ def contract_review(framework_slug, agreement_id):
...
+        email_client = DMNotifyClient()
+        for email_address in returned_agreement_email_recipients(supplier_framework):
+            try:
+                email_client.send_email(
+                    to_email_address=email_address,
+                    template_name_or_id="framework_agreement_signature_page",
+                    personalisation={
+                        "framework_name": framework['name'],
+                        "framework_slug": framework['slug'],
+                        "framework_updates_url": url_for(".framework_updates", framework_slug=framework["slug"]),
```
katstevens: We need the fully qualified URL here - otherwise the template will be populated with `/suppliers/...`. There should be a utils function to grab this for whatever environment you happen to be in.

app/main/views/frameworks.py
```diff
@@ -1124,29 +1124,20 @@ def contract_review(framework_slug, agreement_id):
...
+                    }
+                )
+            except EmailError:
+                pass
```
katstevens: Agree that we shouldn't `abort(503)` here, but it feels like we should still have an `current_app.logger.error()` here as well as the Notify client logging. What do others think?

app/main/views/frameworks.py
```diff
@@ -1124,29 +1125,23 @@ def contract_review(framework_slug, agreement_id):
...
+                        "framework_name": framework['name'],
+                        "framework_slug": framework['slug'],
+                        "framework_updates_url": (
+                            get_web_url_from_stage(current_app.config["DM_ENVIRONMENT"])
+                            + url_for(".framework_updates", framework_slug=framework["slug"])
+                        ),
+                    }
+                )
+            except EmailError:
+                pass
```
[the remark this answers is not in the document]
risicle: What severity level is the in-notify log?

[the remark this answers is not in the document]
lfdebrux: It's an error

https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/email/dm_notify.py#L175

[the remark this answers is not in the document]
risicle: Ahh it's probably alright, given that we have the automagic app backtrace follower thingy. Might be nice to leave a tiny comment like `# error log should have already been emitted by client` to stop the next person's spidey sense tingling when they get to this line.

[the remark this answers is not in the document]
lfdebrux: Done.

app/main/views/frameworks.py
```diff
@@ -1124,29 +1125,25 @@ def contract_review(framework_slug, agreement_id):
...
+                        "framework_slug": framework['slug'],
+                        "framework_updates_url": (
+                            get_web_url_from_stage(current_app.config["DM_ENVIRONMENT"])
+                            + url_for(".framework_updates", framework_slug=framework["slug"])
+                        ),
+                    }
+                )
+            except EmailError:
+                # We don't need to handle this as the email is only informational,
+                # and failures are already logged by DMNotifyClient
```
katstevens: Yep, this looks good to me

tests/app/main/test_frameworks.py
```diff
@@ -4890,15 +4893,14 @@ def test_valid_framework_agreement_returned_updates_api_and_sends_confirmation_e
...
 
         # Delcaration primaryContactEmail and current_user.email_address are different so expect two recipients
-        mandrill_send_email.assert_called_once_with(
-            mock.ANY,  # self
-            to_email_addresses=['email2@email.com', 'email@email.com'],
-            email_body=mock.ANY,
-            subject='Your G-Cloud 8 signature page has been received',
-            from_email_address='enquiries@digitalmarketplace.service.gov.uk',
-            from_name='Digital Marketplace Admin',
-            tags=['g-cloud-8-framework-agreement']
```
katstevens: It'd be great if we could assert the notify call args on at least one of these tests as well as the recipients, that way we could make sure we're sending the right personalisations (including the full url).

lfdebrux: Asserting the full url is a pain, because it depends on the stage, which is why I did the funny business in [these lines](https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/935/files#diff-860e899ae62cbab98fee8fa871bcc53bR4900)

lfdebrux: Testing the personalisation in more detail can be done though

tests/app/main/test_frameworks.py
```diff
@@ -4890,15 +4893,16 @@ def test_valid_framework_agreement_returned_updates_api_and_sends_confirmation_e
...
-        )
+        assert self.notify_send_email.call_count == 2
+        assert self.notify_send_email.call_args_list[0][1]["to_email_address"] == "email2@email.com"
+        assert self.notify_send_email.call_args_list[1][1]["to_email_address"] == "email@email.com"
+
+        assert self.notify_send_email.call_args[1]["personalisation"]["framework_name"] == "G-Cloud 8"
+
+        # Check that email url is valid
+        framework_updates_url = self.notify_send_email.call_args[1]["personalisation"]["framework_updates_url"]
+        assert framework_updates_url.endswith("/suppliers/frameworks/g-cloud-8/updates")
```
katstevens: Yep, this looks fine 👍

PR #956, opened 2019-01-11
Remove mandrill

jonodrew:
The aim of this PR is to remove mandrill from `supplier-frontend` entirely. Maintaining multiple clients complicates the codebase and takes up brain space for developers. [Ticket](https://trello.com/c/w6nXOqSR/317-replace-mandrill-with-notify-in-supplier-frontend)
In addition, the mandrill client is now deprecated

1. The first commit removes a route that wasn't used any more. It also removes the associated email template
2. The second commit brings in [new functionality from utils](https://github.com/alphagov/digitalmarketplace-utils/commit/0977ac95c26c73ea87e22a825d97d9b94539930a)
3. The third commit replaces mandrill with notify in the framework_dashboard 
4. The fourth commit replaces mandrill with notify in the `view_contract_variation` route
5. The fifth commit removes mandrill from a number of tests where it was passed in but not used. I think this clears the tests up a little and makes it clearer what is being tested.


tests/app/main/test_frameworks.py
```diff
@@ -1706,72 +1706,6 @@ def teardown_method(self, method):
         self.data_api_client_patch.stop()
         super().teardown_method(method)
 
-    def test_page_renders_if_all_ok(self):
```
katstevens: We don't want to remove the tests for the GET view, only the POST.

app/main/views/frameworks.py
```diff
@@ -1121,24 +1124,21 @@ def view_contract_variation(framework_slug, variation_slug):
         )
 
         # Send email confirming accepted
-        mandrill_client = DMMandrillClient()
-        email_body = render_template(
-            'emails/{}_variation_{}_agreed.html'.format(framework_slug, variation_slug)
+        notify_client = DMNotifyClient(
+            current_app.config['DM_NOTIFY_API_KEY'],
```
katstevens: As before, don't need the config to be supplied here

app/main/views/frameworks.py
```diff
@@ -1121,24 +1124,21 @@ def view_contract_variation(framework_slug, variation_slug):
...
+            templates=dict()
```
katstevens: `DMNotifyClient` looks for a dict of templates in the app's config during `__init__`, just like it does for the api key. We should be keeping any template IDs we end up creating in the app config - this line would overwrite that config, and we'd never find our framework's template ID in the try/except below.

app/main/views/frameworks.py
```diff
@@ -1087,6 +1087,9 @@ def contract_review(framework_slug, agreement_id):
 @main.route('/frameworks/<framework_slug>/contract-variation/<variation_slug>', methods=['GET', 'POST'])
 @login_required
 def view_contract_variation(framework_slug, variation_slug):
+    """
+    This view asks suppliers to agree to a framework variation and then generates a confirmation email  when they do.
```
katstevens: I'd add a massive `TODO: create a variation template in Notify before adding a variation to a framework in the API` here.

tests/app/main/test_frameworks.py
```diff
@@ -4945,10 +4945,12 @@ def test_shows_updated_heading_and_countersigner_details_but_no_form_if_counters
         assert len(doc.xpath('//label[contains(text(), "I accept these proposed changes")]')) == 0
         assert len(doc.xpath('//input[@value="I accept"]')) == 0
 
-    def test_api_is_called_to_agree(self):
+    @mock.patch('app.main.views.frameworks.DMNotifyClient', autospec=True)
+    def test_api_is_called_to_agree(self, mocked_notify_class):
```
katstevens: I have no idea what this test name means 😹 I know it was there before but can we take this opportunity to make it more descriptive? Either that or combine it with `test_email_is_sent_to_correct_users` below (same scenario, each thing being asserted happens in both tests)

tests/app/main/test_frameworks.py
```diff
@@ -4960,47 +4962,43 @@ def test_api_is_called_to_agree(self):
...
         )
 
-    def test_success_message_is_displayed_on_success(self):
+    @mock.patch('app.main.views.frameworks.DMNotifyClient', autospec=True)
+    def test_success_message_is_displayed_on_success(self, mocked_notify_class):
+        mocked_notify_client = mocked_notify_class.return_value
+        mocked_notify_client.templates = {'g-cloud-8_variation_1': 123456789}
         self.data_api_client.get_framework.return_value = self.g8_framework
         self.data_api_client.get_supplier_framework_info.return_value = self.good_supplier_framework
         res = self.client.post(
```
katstevens: The test below this one (`test_api_is_not_called_and_no_email_sent_for_subsequent_posts`) still has a `DMMandrillClient` patch - it needs replacing with the Notify client so we can assert the email isn't sent.

app/main/views/frameworks.py
```diff
@@ -91,24 +91,25 @@ def framework_dashboard(framework_slug):
...
-                "Application started email failed to send: {error}, supplier_id: {supplier_id}",
-                extra={'error': str(e), 'supplier_id': current_user.supplier_id}
+
+        for address in [user['emailAddress'] for user in supplier_users if user['active']]:
+            # this has no try block as any error is caught and handled by the DMNotifyClient
+            notify_client.send_email(
+                to_email_address=address,
+                template_name_or_id=notify_client.templates[framework['family']],
+                personalisation={
+                    'framework_name': framework['name'],
```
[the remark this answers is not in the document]
jonodrew: If we have two separate templates, no, and if the content is miles apart then it's probably worth taking that approach.

requirements.txt
```diff
@@ -31,22 +31,22 @@ idna==2.8
 inflection==0.3.1
 Jinja2==2.10
 jmespath==0.9.3
-mailchimp3==2.0.18
+mailchimp3==3.0.6
```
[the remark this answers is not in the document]
jonodrew: It passes all the tests - how else can I make sure it's behaving itself?

app/main/views/frameworks.py
jonodrew: This is a mistake on my part. I need to write the template, but your comment speaks to a wider thing - should the templates be part of the config rather than inserted on initialisation?

katstevens: Yes - that's the pattern we use across the apps.

app/main/views/frameworks.py
```diff
@@ -91,24 +91,18 @@ def framework_dashboard(framework_slug):
...
+        for address in [user['emailAddress'] for user in supplier_users if user['active']]:
+            # this has no try block as any error is caught and handled by the DMNotifyClient
+            notify_client.send_email(
+                to_email_address=address,
+                template_name_or_id=notify_client.templates[framework['family']],
+                personalisation={
+                    'framework_applications_close_date': framework['applicationsCloseAt'],
+                    'framework_clarification_questions_close_date': framework['clarificationsCloseAt'],
+                },
+                reply_to_address_id='24908180-b64e-513d-ab48-fdca677cec52'  # this is enquiries@...
```
benvand: I'd expect this UUID to be a config value. If I went looking for it the first place I'd look would be the config.

app/main/views/frameworks.py
```diff
@@ -1209,24 +1109,18 @@ def view_contract_variation(framework_slug, variation_slug):
...
-            )
+        notify_client = DMNotifyClient()
+        for address in returned_agreement_email_recipients(supplier_framework):
+            try:
+                notify_client.send_email(
+                    to_email_address=address,
+                    template_name_or_id=notify_client.templates.get(f'{framework_slug}_variation_{variation_slug}'),
+                    personalisation={'framework_name': framework_slug}
+                )
+            except KeyError:
```
benvand: The `.get` method on a dict will not raise a `KeyError` if the key does not exist. It returns none by default. You might be thinking of the square brackets accessor.

benvand: The `try: except` here realistically serves to suppress an exception in the case that the template doesn't exist.
I think I'd rather have an exception. We'd likely notice much faster that this wasn't working.

config.py
```diff
@@ -34,6 +34,8 @@ class Config(object):
         "create_user_account": "84f5d812-df9d-4ab8-804a-06f64f5abd30",
         "invite_contributor": "5eefe42d-1694-4388-8908-991cdfba0a71",
         "framework_agreement_signature_page": "08929c93-f9e7-4b40-b75f-418659308324",
+        'g-cloud': '7362aad7-1ffa-4e36-8ddd-d16cc5305175',
+        'digital-outcomes-and-specialists': '1920c708-66a9-466e-a0ae-3bd13d31e32d'
```
benvand: Hard to tell what these keys relate to at a glance... framework names are a bit vague and could relate to anything `variation_agreement`? `framework_acceptance`? `kicked_off_framework`?

app/main/views/frameworks.py
benvand: Could use detailing here too:
https://alphagov.github.io/digitalmarketplace-manual/contract-variations.html?highlight=variation

app/main/views/frameworks.py
```diff
@@ -1209,24 +1109,18 @@ def view_contract_variation(framework_slug, variation_slug):
...
-            current_app.logger.error(
-                "Variation agreed email failed to send: {error}, supplier_id: {supplier_id}",
-                extra={'error': str(e), 'supplier_id': current_user.supplier_id}
-            )
+        notify_client = DMNotifyClient()
+        for address in returned_agreement_email_recipients(supplier_framework):
+            try:
+                notify_client.send_email(
+                    to_email_address=address,
+                    template_name_or_id=notify_client.templates.get(f'{framework_slug}_variation_{variation_slug}'),
```
benvand: I think leaving the `_agreed` on the end of the key is more descriptive, and keeps in line with the old name of the file.

app/main/views/frameworks.py
```diff
@@ -835,133 +829,35 @@ def framework_agreement(framework_slug):
...
     return render_template(
-        "frameworks/agreement.html",
+        'frameworks/contract_start.html',
+        signature_page_filename=SIGNATURE_PAGE_FILENAME,
         framework=framework,
+        framework_urls=content_loader.get_message(framework_slug, 'urls'),
+        lots=[{
+            'name': lot['name'],
+            'result': lot_results[lot['slug']]
+        } for lot in framework['lots']],
```
benvand: Doesn't requre splitting over multiple lines:
https://gds-way.cloudapps.digital/manuals/programming-languages/python/style-guide.html#line-length-lt-120

app/main/views/frameworks.py
```diff
@@ -91,24 +91,18 @@ def framework_dashboard(framework_slug):
...
-                tags=['{}-application-started'.format(framework_slug)],
-            )
-        except EmailError as e:
-            current_app.logger.error(
-                "Application started email failed to send: {error}, supplier_id: {supplier_id}",
-                extra={'error': str(e), 'supplier_id': current_user.supplier_id}
+        notify_client = DMNotifyClient()
+
+        for address in [user['emailAddress'] for user in supplier_users if user['active']]:
+            # this has no try block as any error is caught and handled by the DMNotifyClient
```
benvand: Good comment for the review, but probably not required for the final merge

app/main/views/frameworks.py
```diff
@@ -1123,10 +1005,10 @@ def contract_review(framework_slug, agreement_id):
...
 
-        email_client = DMNotifyClient()
+        notify_client = DMNotifyClient()
         for email_address in returned_agreement_email_recipients(supplier_framework):
             try:
-                email_client.send_email(
+                notify_client.send_email(
                     to_email_address=email_address,
                     template_name_or_id="framework_agreement_signature_page",
                     personalisation={
```
katstevens: We should add in a `reference` kwarg here for the Notify client, with a hash of the email address (like we do for the clarification question email above)

app/main/views/frameworks.py
```diff
@@ -1175,6 +1057,10 @@ def contract_review(framework_slug, agreement_id):
 @main.route('/frameworks/<framework_slug>/contract-variation/<variation_slug>', methods=['GET', 'POST'])
 @login_required
 def view_contract_variation(framework_slug, variation_slug):
+    """
+    This view asks suppliers to agree to a framework variation and then generates a confirmation email  when they do.
+    """
+    # TODO: create a variation template in Notify web UI before addiing a variation to a framework in the API
```
katstevens: Small typo `addiing` but otherwise 👍

app/main/views/frameworks.py
katstevens: Wait a second - what email is this again? Is this `contract_review` view ever used for G8 and up?

tests/app/main/test_frameworks.py
```diff
@@ -5353,7 +5050,7 @@ def test_api_is_not_called_and_no_email_sent_for_subsequent_posts(self, mandrill
         )
         assert res.status_code == 200
         assert self.data_api_client.agree_framework_variation.called is False
-        assert mandrill_send_email.called is False
+        assert mocked_notify_client.called is False
```
katstevens: Great 👍 Though I've just realised the happy path test above doesn't assert that the Notify client is actually called (I missed that one first time round, sorry).

app/main/views/frameworks.py
jonodrew: In the above, we use the current user's email address. Why is that? And is that what I should replicate?

katstevens: It's so that we can find the unique reference to it in the Notify dashboard if we need to. So just using the email address we're sending to in `to_email_address` is fine.

jonodrew: > Wait a second - what email is this again? Is this `contract_review` view ever used for G8 and up?

yes - we removed the other route, but this is the one that's used for G8 and up

katstevens: Ah yes, sorry. This file is too confusing! So this is the one that's already using Notify, got it. We should still add a reference I think...

tests/app/main/test_frameworks.py
```diff
@@ -4821,9 +4525,11 @@ def test_we_abort_if_agreement_does_not_match_supplier_framework(
             supplier_framework
         )
 
+    @mock.patch('app.main.views.frameworks.DMNotifyClient', autospec=True)
```
katstevens: I think we already patch the `send_email` method of the DMNotifyClient in the setup of this test class so we don't need this patch - we can use `self.notify_send_email` in the test.

tests/app/main/test_frameworks.py
jonodrew: Where sorry?

app/main/views/frameworks.py
```diff
@@ -782,23 +782,13 @@ def framework_updates_email_clarification_question(framework_slug):
...
-                    'email_hash': hash_string(current_user.email_address),
-                    'code': 'clarification-question-confirm-email.fail'
-                })
+        notify_client.send_email(
+            current_user.email_address,
+            template_name_or_id=notify_client.templates['confirmation_of_clarification_question'],
+            personalisation=confirmation_email_personalisation,
+            reference='clarification-question-confirm-{}'.format(hash_string(current_user.email_address)),
+            reply_to_address_id=current_app.config['DM_ENQUIRIES_EMAIL_ADDRESS_UUID']
+        )
```
katstevens: Ah - I think we wanted this email to fail silently, hence the try..except. It's just a confirmation receipt so not crucial. There's a comment on line 775 for this.

jonodrew: I think this will still fail silently if the email doesn't send, with the error being picked up by the client. If there's a `KeyError` then it'll fail more noisily. Is that okay?

katstevens: Hmm I think the client will still raise an `EmailError` (https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/email/dm_notify.py#L178). It'll handle the logging itself but the user will still get a 500.

jonodrew: Ah, you're right. I'll put in the try...except

app/main/views/frameworks.py
```diff
@@ -1209,23 +1096,13 @@ def view_contract_variation(framework_slug, variation_slug):
...
-            )
-        except EmailError as e:
-            current_app.logger.error(
-                "Variation agreed email failed to send: {error}, supplier_id: {supplier_id}",
-                extra={'error': str(e), 'supplier_id': current_user.supplier_id}
+        notify_client = DMNotifyClient()
+        for address in returned_agreement_email_recipients(supplier_framework):
+            notify_client.send_email(
+                to_email_address=address,
+                template_name_or_id=notify_client.templates[f'{framework_slug}_variation_{variation_slug}_agreed'],
```
katstevens: I quite like having this explicitly fetch a template that we haven't created yet - it'll be extra visible that some action is required if/when we come to use it!

app/main/views/frameworks.py
```diff
@@ -835,133 +828,32 @@ def framework_agreement(framework_slug):
...
-            } for lot in framework['lots']],
-            supplier_framework=supplier_framework,
-            supplier_registered_name=get_supplier_registered_name_from_declaration(supplier_framework['declaration']),
-        ), 200
+    drafts, complete_drafts = get_drafts(data_api_client, framework_slug)
+    complete_drafts_by_lot = {
+        lot['slug']: [draft for draft in complete_drafts if draft['lotSlug'] == lot['slug']]
+        for lot in framework['lots']
+    }
+    lot_results = {k: lot_result(v) for k, v in complete_drafts_by_lot.items()}
```
benvand: Readability could benefit from (IMO)
```
lot_results = {lot_slug: lot_result(draft_services) for lot_slug, draft_services in complete_drafts_by_lot.items()}
```
I think this has the advantage of cementing the structure of the above comprehension. But it's a bit of a nitpick tbh :)

benvand:
(👋 bye Mandrill!)

jonodrew:
> (👋 bye Mandrill!)

Not quite...still that pesky Zendesk/Salesforce question to solve!

PR #959, opened 2019-01-22
Replace mandrill with notify

jonodrew:
I'm aiming to [replace mandrill throughout supplier-frontend](https://trello.com/c/jrpQHpGh/301-replace-mandrill-with-notify-in-frameworkupdatesemailclarificationquestion-route-on-supplier-fe), and this is the last thing that needs to be done. After this there'll be a pull request to utils to remove the client, but for now here's the last replacement of mandrill with notify.

I've updated an assertion method in the tests, and just for fun I've moved s3 patches to where they're needed, rather than dropping them on the whole class.

Huge thanks to @katstevens for pairing with me on this

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -19,7 +19,6 @@ from dmutils.documents import (
     degenerate_document_path_and_return_doc_name, get_signed_url, get_extension, file_is_less_than_5mb,
     file_is_image, file_is_pdf, sanitise_supplier_name
 )
-from dmutils.email.dm_mandrill import DMMandrillClient
 from dmutils.email.dm_notify import DMNotifyClient
 from dmutils.email.exceptions import EmailError
 from dmutils.email.helpers import hash_string
@@ -722,42 +721,33 @@ def framework_updates_email_clarification_question(framework_slug):
             default_textbox_value=clarification_question
         )
 
-    # Submit email to Zendesk so the question can be answered
+    # Submit email to CCS so the question can be answered
     # Fail if this email does not send
     if framework['clarificationQuestionsOpen']:
-        subject = "{} clarification question".format(framework['name'])
         to_address = current_app.config['DM_CLARIFICATION_QUESTION_EMAIL']
-        from_address = current_app.config['DM_ENQUIRIES_EMAIL_ADDRESS']
-        email_body = render_template(
-            "emails/clarification_question.html",
-            supplier_id=current_user.supplier_id,
-            message=clarification_question
-        )
-        tags = ["clarification-question"]
+        personalisation = {
+            "framework_name": framework['name'],
+            "supplier_id": current_user.supplier_id,
+            "clarification_question": clarification_question
+        }
+        template = 'framework-clarification-question'
     else:
-        subject = "{} application question".format(framework['name'])
         to_address = current_app.config['DM_FOLLOW_UP_EMAIL_TO']
-        from_address = current_user.email_address
-        email_body = render_template(
-            "emails/follow_up_question.html",
-            supplier_name=current_user.supplier_name,
-            user_name=current_user.name,
-            user_email=current_user.email_address,
-            framework_name=framework['name'],
-            message=clarification_question
-        )
-        tags = ["application-question"]
+        personalisation = {
+            "framework_name": framework['name'],
+            "supplier_name": current_user.supplier_name,
+            "user_name": current_user.name,
+            "user_email": current_user.email_address,
+            "application_question": clarification_question
+        }
+        template = "framework-application-question"
 
-    mandrill_client = DMMandrillClient()
+    notify_client = DMNotifyClient()
     try:
-        mandrill_client.send_email(
-            to_email_addresses=to_address,
-            email_body=email_body,
-            subject=subject,
-            from_email_address=current_app.config["DM_ENQUIRIES_EMAIL_ADDRESS"],
-            from_name="{} Supplier".format(framework['name']),
-            tags=tags,
-            reply_to=from_address,
+        notify_client.send_email(
+            to_email_address=to_address,
+            template_name_or_id=template,
+            personalisation=personalisation,
         )
     except EmailError as e:
         current_app.logger.error(
@@ -784,7 +774,7 @@ def framework_updates_email_clarification_question(framework_slug):
         try:
             notify_client.send_email(
                 current_user.email_address,
-                template_name_or_id=notify_client.templates['confirmation_of_clarification_question'],
+                template_name_or_id='confirmation_of_clarification_question',
                 personalisation=confirmation_email_personalisation,
                 reference='clarification-question-confirm-{}'.format(hash_string(current_user.email_address)),
                 reply_to_address_id=current_app.config['DM_ENQUIRIES_EMAIL_ADDRESS_UUID']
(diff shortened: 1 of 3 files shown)


tests/app/main/test_frameworks.py
```diff
@@ -3395,94 +3392,53 @@ def _send_email(self, clarification_question):
             data={'clarification_question': clarification_question}
         )
 
-    def _assert_clarification_email(self, mandrill_send_email, notify_send_email, is_called=True, succeeds=True,
-                                    clarification_question=''):
+    def _assert_clarification_email(self, notify_send_email, is_called=True, succeeds=True,
+                                    clarification_question='', times_called=0):
```
katstevens: Do we need all of `times_called`, `succeeds` and `is_called`? I'd recommend having something like 

```
assert notify_send_email.call_count == no_of_emails_sent
if no_of_emails_sent:
   assert notify_send_email.call_args_list == [
       mock.call(mock.ANY, to_email_address=...)
   ]
```

katstevens: I'd also keep the `_assert_clarification_email` and `_assert_framework_application_email` separate, then in the test we can call both 

```
# clarification email
self._assert_clarification_email(..., no_of_emails_sent=1)
self._assert_framework_application_email(..., no_of_emails_sent=0)
# framework application email
self._assert_clarification_email(..., no_of_emails_sent=0)
self._assert_framework_application_email(..., no_of_emails_sent=1)
```

tests/app/main/test_frameworks.py
```diff
@@ -3395,94 +3392,53 @@ def _send_email(self, clarification_question):
...
-                from_email_address="enquiries@digitalmarketplace.service.gov.uk",
-                from_name="Test Framework Supplier",
-                subject="Test Framework application question",
-                email_body=FakeMail(*email_body_content),
-                tags=["application-question"],
-                reply_to="email@email.com",
-            )
-
-    def test_should_call_send_email_with_correct_params(self, mandrill_send_email, notify_send_email, s3):
+    @mock.patch('dmutils.s3.S3')
```
katstevens: Where are we using S3 in this view? Do we need to patch it?

tests/app/main/test_frameworks.py
```diff
@@ -3395,94 +3392,53 @@ def _send_email(self, clarification_question):
...
+        self._assert_clarification_email(notify_send_email,
+                                         clarification_question=clarification_question, times_called=2)
 
         assert response.status_code == 200
         assert self.strip_all_whitespace(
             '<p class="banner-message">Your clarification question has been sent. Answers to all ' +
             'clarification questions will be published on this page.</p>'
         ) in self.strip_all_whitespace(response.get_data(as_text=True))
 
-    def test_should_call_send_g7_email_with_correct_params(self, mandrill_send_email, notify_send_email, s3):
```
katstevens: Is this test G7 specific? I think it's the only place we cover the scenario where `clarification_questions_open=False` and we send the framework-application email. Can we replace it with G10 and see what happens?

tests/app/main/test_frameworks.py
```diff
@@ -3535,39 +3492,14 @@ def test_should_create_audit_event(self, mandrill_send_email, notify_send_email,
             data={"question": clarification_question, 'framework': 'g-cloud-7'}
         )
 
-    def test_should_create_g7_question_audit_event(self, mandrill_send_email, notify_send_email, s3):
```
katstevens: I think we can combine this with the other 'G7' test by moving the audit event assert up there.

tests/app/main/test_frameworks.py
jonodrew: Yep. This test calls `_send_email` on l.3387, which posts to https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/841d428158e2508c30d7a2081b87aca9c35b2073/app/main/views/frameworks.py#L685, which instantiates an s3 object. I could move the patch up to `_send_email`, where it's closer to what's using it?

jonodrew: Yeah, that's much neater. I'll do that

katstevens: I think it's the GET view that uses the S3 object, not the POST one?

jonodrew: 🤔 yeah, you're right. So why does it break so hard when I take it out...

tests/app/main/test_frameworks.py
jonodrew: The `succeeds` keyword is useful for testing the behaviour under test in `test_should_be_a_503_if_email_fails`. In that case, we expect there to only be one call, because when it fails the whole view aborts and doesn't even try to send the second email. However, to test that behaviour in the same method I think you need to have
```
if not succeeds:
    notify_send_email.assert_has_calls(# email to DMP)
else:
    notify_send_email.assert_has_calls(email to DMP, receipt to supplier)
```

tests/app/main/test_frameworks.py
katstevens: Ah, the POST view returns a redirect to the success page. So we do need the patch after all (my mistake!)

tests/app/main/test_frameworks.py
katstevens: I think the test scenarios themselves need a bit of a rethink. Let's chat when you have a sec.

app/main/views/frameworks.py
```diff
@@ -725,39 +724,30 @@ def framework_updates_email_clarification_question(framework_slug):
     # Submit email to Zendesk so the question can be answered
```
lfdebrux: Can we change this comment so that we don't have an erroneous reference to Zendesk?

lfdebrux:
```suggestion
    # Submit email to CCS so the question can be answered
```

app/main/views/frameworks.py
```diff
@@ -725,39 +724,30 @@ def framework_updates_email_clarification_question(framework_slug):
...
-            to_email_addresses=to_address,
-            email_body=email_body,
-            subject=subject,
-            from_email_address=current_app.config["DM_ENQUIRIES_EMAIL_ADDRESS"],
-            from_name="{} Supplier".format(framework['name']),
-            tags=tags,
-            reply_to=from_address,
+        notify_client.send_email(
+            to_email_address=to_address,
+            template_name_or_id=notify_client.templates[template],
```
lfdebrux: I think looking up the template name here is strictly unnecessary, it could equally be

```suggestion
            template_name_or_id=template,
```

config.py
```diff
@@ -35,7 +35,9 @@ class Config(object):
         "invite_contributor": "5eefe42d-1694-4388-8908-991cdfba0a71",
         "framework_agreement_signature_page": "08929c93-f9e7-4b40-b75f-418659308324",
         'g-cloud-application-started': '7362aad7-1ffa-4e36-8ddd-d16cc5305175',
-        'digital-outcomes-and-specialists-application-started': '1920c708-66a9-466e-a0ae-3bd13d31e32d'
+        'digital-outcomes-and-specialists-application-started': '1920c708-66a9-466e-a0ae-3bd13d31e32d',
+        'framework-clarification-question': "8574484f-9907-44c0-b9d5-1120beb64ef0",
+        'framework-application-question': '6681d4a1-6e30-407f-9e8b-3f6633d73546'
```
lfdebrux:
```suggestion
        'framework-application-question': '6681d4a1-6e30-407f-9e8b-3f6633d73546',
```

app/main/views/frameworks.py
jonodrew: Ah yes, you're right!

config.py
jonodrew: I don't understand the addition of the comma here?

lfdebrux: It makes a nicer diff when you add another element to the dict (see the first line of this diff for instance); it's strictly cosmetic and a matter of personal preference of course.

PR #991, opened 2019-02-12
Declaration reuse bug

katstevens:
Trello: https://trello.com/c/itTrzlID/363-g11-declaration-re-use-bug

The declaration reuse form had an unnecessary validator on it, causing a bug where the supplier could not proceed if they did not want to copy their declaration. 

The same issue was on a DOS validator for copying services, so I've updated that too.

Also bumps our 3 DM libraries to get rid of some deprecation warnings, and get rid of the unused Mandrill dependency.

### DIFF
--- a/app/main/forms/frameworks.py
+++ b/app/main/forms/frameworks.py
@@ -65,7 +65,6 @@ class OneServiceLimitCopyServiceForm(FlaskForm):
         "Do you want to reuse your previous {lot_name} service?",
         question_advice="You still have to review your service and answer any new questions.",
         false_values=("False", "false", ""),
-        validators=[InputRequired(message="You must answer this question.")],
         widget=DMSelectionButtonBase(type="boolean"),
     )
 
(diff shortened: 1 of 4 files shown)


config.py
```diff
@@ -128,7 +128,7 @@ class Development(Config):
     DM_DOCUMENTS_BUCKET = "digitalmarketplace-dev-uploads"
     DM_ASSETS_URL = "https://{}.s3-eu-west-1.amazonaws.com".format(DM_SUBMISSIONS_BUCKET)
 
-    DM_NOTIFY_API_KEY = "not_a_real_key-00000000-fake-uuid-0000-000000000000"
+    DM_NOTIFY_API_KEY = "katsandbox-95316ff0-e555-462d-a6e7-95d26fbfd091-83356680-310b-4f0f-a25d-613fb10066de"
```
jonodrew: I think this is better as an obviously fake key, so we don't end up using your sandbox key?

katstevens: I've revoked that key anyway.

PR #995, opened 2019-02-14
clarification/follow-up questions: use better default email addresses for these

risicle:
Spurred by https://trello.com/c/49yR7N9O/

These defaults currently end up actually getting used in the functional tests, so it'll probably be good to have more deliberate ones now we're actually going to be exercising them

### DIFF
--- a/config.py
+++ b/config.py
@@ -24,7 +24,7 @@ class Config(object):
     DM_DATA_API_URL = None
     DM_DATA_API_AUTH_TOKEN = None
     DM_NOTIFY_API_KEY = None
-    DM_CLARIFICATION_QUESTION_EMAIL = 'digitalmarketplace@mailinator.com'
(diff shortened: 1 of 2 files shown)


config.py
```diff
@@ -24,7 +24,7 @@ class Config(object):
     DM_DATA_API_URL = None
     DM_DATA_API_AUTH_TOKEN = None
     DM_NOTIFY_API_KEY = None
-    DM_CLARIFICATION_QUESTION_EMAIL = 'digitalmarketplace@mailinator.com'
+    DM_CLARIFICATION_QUESTION_EMAIL = 'clarification-questions@example.gov.uk'
```
katstevens: Do we want `example.gov.uk` or `example.com`?

risicle: I don't think it actually makes a difference here, but I went `gov` simply because it's mimicking what will presumably be a CCS address IRL.

risicle:
Actually I also have to modify the email ref generation process so I might close this and open a broader PR.

PR #996, opened 2019-02-15
clarification/follow-up questions: use custom-generated references for emails

risicle:
For the purpose of https://trello.com/c/49yR7N9O

These should make it easier to write functional tests for this process. Note the critical emails use `allow_resend=True` to prevent any (however unlikely) reference collisions preventing an email getting through.


risicle:
TYVM

PR #1005, opened 2019-03-11
Refactor forms

lfdebrux:
Here's some refactoring I did of some of the forms in the supplier frontend.

The changes are part of what I was doing https://github.com/alphagov/digitalmarketplace-supplier-frontend/tree/ldeb-107-wtforms-dmfields-supplier-forms, but packaged nicer.

I want these changes to be in the supplier frontend now because I intend to reuse (some of) the code for the admin frontend, and it would be helpful to be able to point to the supplier frontend code in `master` as an exemplar.

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -10,7 +10,7 @@ from dmutils.dates import update_framework_with_formatted_dates
 from dmutils.email import send_user_account_email
 from dmutils.email.dm_mailchimp import DMMailChimpClient
 from dmutils.flask import timed_render_template as render_template
-from dmutils.forms.helpers import remove_csrf_token, get_errors_from_wtform
+from dmutils.forms.helpers import get_errors_from_wtform
 from dmutils.errors import render_error_page
 
 from ...main import main, content_loader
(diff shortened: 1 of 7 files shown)


app/main/views/suppliers.py
```diff
@@ -418,7 +416,13 @@ def edit_supplier_organisation_size():
 @main.route('/trading-status/edit', methods=['GET', 'POST'])
 @login_required
 def edit_supplier_trading_status():
-    form = CompanyTradingStatusForm()
+
+    prefill_data = {}
+    if request.method == "GET":
```
risicle: Why does this view only conditionally load the existing supplier data and the others load it for both `GET` and `POST`?

lfdebrux: I was worried about making the GET more expensive; the other views GET was already including the api requests.

I'm happy to take it out though, it is a bit premature optimisation-y

risicle: I was actually looking at it the other way around - why don't we do it in all the others too? One advantage I can think of is a small amount of added clarity - it's more clear the `prefill_data` can't have any effect in the `POST` case if it never gets filled out in the first place.

Unless, I reasoned, the use of supplier-fetching here is being used as a guard to generate 404s if the supplier doesn't exist - but... that's ... not possible here? Guessing most of this apps use a `login_required` decorator that requires `role == "supplier"`?

PR #1008, opened 2019-03-18
Modern slavery validation

katstevens:
Trello: https://trello.com/c/76VkxnU3/425-modern-slavery-validation-messages

The logic for required fields on the Modern Slavery section should go:

`Turnover: yes` + `Reporting requirements: yes` = statement required, explanation text not required
`Turnover: yes` + `Reporting requirements: no` = statement not required, explanation text required
`Turnover: no`: nothing else required

The extra logic triggered the flake8 complexity warning, I ignored it for the validator in the end (but did move out the declaration fields to an attribute of the validator class, which seemed neater).

The tests for the validator are a bit out of date, it might be good to update those later.

Screengrab:
modern-slavery-declaration-validation

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -8,7 +8,7 @@
 # W504: line break after binary operator
 exclude = venv*,__pycache__,node_modules
 ignore = D203,W503,W504
-max-complexity = 7
+max-complexity = 8
 max-line-length = 120
 per-file-ignores =
     app/__init__.py : E402
--- a/app/main/helpers/validation.py
+++ b/app/main/helpers/validation.py
@@ -229,6 +229,12 @@ class DOSValidator(DeclarationValidator):
             if self.answers.get('licenceOrMemberRequired') in ['licensed', 'a member of a relevant organisation']:
                 req_fields.add('licenceOrMemberRequiredDetails')
 
+        # If supplier doesn't meet the Modern Slavery reporting requirements, they don't need to upload a statement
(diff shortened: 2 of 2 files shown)


.flake8
```diff
@@ -15,3 +15,4 @@ per-file-ignores =
     app/main/__init__.py : E402, F401
     app/status/__init__.py : E402, F401
     app/main/views/*.py : C901
+    app/main/helpers/validation.py : C901
```
lfdebrux: Can we add the flask ignore to the particular method rather than the file as a whole?

app/main/helpers/validation.py
```diff
@@ -183,33 +183,35 @@ class DOSValidator(DeclarationValidator):
...
+        ],
+        "modernSlaveryReportingRequirements": [
+            "modernSlaveryTurnover"
+        ],
+    }
+
     email_validation_fields = set(["contactEmailContractNotice", "primaryContactEmail"])
     character_limit = 5000
 
     def get_required_fields(self):
```
lfdebrux: To flake8 ignore just this method as mentioned in previous comment:

```suggestion
    def get_required_fields(self):  #noqa: C901
```

.flake8
katstevens: Tbh I might just bump up the complexity limit from 7 to 8 and get rid of the ignore.

lfdebrux: Haha

PR #1011, opened 2019-03-20
Add display session timeout message during applications

lfdebrux:
We want our users to be aware when filling out long forms that they may be logged out if they do not interact with the website for longer than an hour.

This PR adds a simple text message at the top of framework application forms.

See the [ticket] for more details.

[ticket]: https://trello.com/c/92nGYty4/

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -20,10 +20,13 @@
 //= require ../../../node_modules/digitalmarketplace-frontend-toolkit/toolkit/javascripts/show-hide-content.js
 //= require ../../../node_modules/digitalmarketplace-frontend-toolkit/toolkit/javascripts/user-research-consent-banner.js
 //= require ../../../node_modules/digitalmarketplace-frontend-toolkit/toolkit/javascripts/report-a-problem.js
+//= require ../../../node_modules/govuk-frontend/all.js
 //= require _analytics.js
 //= require _selection-buttons.js
 //= require _stick-at-top-when-scrolling.js
 //= require category-picker.js
+
+GOVUKFrontend.initAll();
 
 (function(GOVUK, GDM) {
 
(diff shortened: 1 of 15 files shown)


tests/app/main/test_session_timeout_warning.py
```diff
@@ -0,0 +1,131 @@
...
+
+        assert doc.xpath(".//div[@id='session-timeout-warning']")
+
+    def test_session_timeout_is_one_hour_in_future(self, form, framework, lot, display_time):
+        with mock.patch("app.main.views.services.render_template") as render_template:
+            render_template.return_value = ""
+
+            self.client.get(f"/suppliers/frameworks/{framework}/submissions/{lot}/1/edit/{form}")
+
+            assert render_template.call_args[1]["session_timeout"] == display_time
```
katstevens: Do we need this extra test module with all the setup/fixtures? I was intending to just add a simple test with a `freeze_time` and an `xpath` assert to the existing test classes for the two views (i.e. `TestSupplierDeclaration` and `TestEditDraftService`).

lfdebrux: I was trying to avoid having the tests for the change be spread across two different files; plus the tests in `TestEditDraftService` scare me a little...

katstevens: Well ok, but it probably makes sense to stick to the convention in this app for having one test module for one code module (so we can easily find all the tests relating to a view).

lfdebrux: Fair point, I will split them up

katstevens: Just a simple `self.client.get` for one of the edit sections, with a `freeze_time` context manager should do the trick. We don't care about the actual form content.

lfdebrux: Okay I've gotten rid of that file and added simpler tests to the classes you suggested

PR #1013, opened 2019-03-20
Change content of session timeout warning

lfdebrux:
Updated the message with suggestions from @pocketstefan.

screenshot including session timeout warning

### DIFF
--- a/app/templates/frameworks/_session_timeout.html
+++ b/app/templates/frameworks/_session_timeout.html
@@ -2,10 +2,10 @@
   <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
   <strong class="govuk-warning-text__text">
     <span class="govuk-warning-text__assistive">Warning</span>
-    <p>You must save your answers by {{session_timeout}}.</p>
+    <p>You must save these answers by {{session_timeout}}.</p>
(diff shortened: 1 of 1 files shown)


app/templates/frameworks/_session_timeout.html
```diff
@@ -2,10 +2,10 @@
   <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
   <strong class="govuk-warning-text__text">
     <span class="govuk-warning-text__assistive">Warning</span>
-    <p>You must save your answers by {{session_timeout}}.</p>
+    <p>You must save these answers by {{session_timeout}}.</p>
     <p>
-      If you don't your session will time out for security reasons and you’ll
-      lose any information you've entered on this page.
+      If you don't save, your session will time out for security reasons.
+      You’ll lose any information you've entered on this page.
```
katstevens: Just noticed we have a curly quote on `You’ll` but not in `don't` or `you've`...

lfdebrux: Argh no!

lfdebrux: Fixed now, good spot!

PR #1016, opened 2019-03-26
Add DMSelectionButtonBase widget to declaration copy form

katstevens:
Trello: https://trello.com/c/I9MdAoMl/666-users-having-problem-copying-declarations-question

Weird bug where the form widget (or lack thereof?) wasn't allowing a True value to be posted.

This changes the presentation of the form to be a Yes/No question, will run it past product/content before merging.

reuse-declaration-yes-no

### DIFF
--- a/app/main/forms/frameworks.py
+++ b/app/main/forms/frameworks.py
@@ -55,6 +55,7 @@ class ReuseDeclarationForm(FlaskForm):
     reuse = DMBooleanField(
         "Do you want to reuse the answers from your earlier declaration?",
         false_values=("False", "false", ""),
+        widget=DMSelectionButtonBase(type="boolean"),
     )
     old_framework_slug = HiddenField()
 


app/main/forms/frameworks.py
```diff
@@ -55,6 +55,7 @@ class ReuseDeclarationForm(FlaskForm):
     reuse = DMBooleanField(
         "Do you want to reuse the answers from your earlier declaration?",
         false_values=("False", "false", ""),
+        widget=DMSelectionButtonBase(type="boolean"),
```
lfdebrux: If you want to retain the checkbox you can use 
```suggestion
        widget=DMCheckboxInput(),
```

katstevens: I think we tried that and it didn't work. Will give it another whirl...

lfdebrux: Actually the DMBooleanField uses DMCheckboxInput by default, there may be an issue with that widget

PR #1022, opened 2019-04-01
Handle new Mailchimp responses

katstevens:
Trello: https://trello.com/c/pzAoSoTC/434-mailchimp-resubscribe-for-users-who-have-unsubscribed

Pulls in https://github.com/alphagov/digitalmarketplace-utils/pull/509 and https://github.com/alphagov/digitalmarketplace-utils/pull/511 (breaking change to subscribing users to Mailchimp lists).

The response has changed for both success and error statuses. We can now differentiate between errors where the user has already subscribed (i.e. no action required), where the user is providing a problematic email address (i.e. action should be to use a different email address) or where there is an unexpected error (do nothing for now).

Also added new flash messages type to cover this extra action (still need content approval for one of the two).


katstevens:
Have now updated the flash message with approved content from Stefan 👍

PR #1028, opened 2019-04-23
Bump digitalmarketplace-utils dependency to v48.2.0

risicle:
Also bump digitalmarketplace-content-loader to 5.2.1 and other version bumps that come with the re-freeze.

This principally brings us the last-ditch exception logging.


risicle:
This is a local bump for local developers.

PR #1033, opened 2019-05-02
801-metrics-on-buyer-and-supplier-we-need-to-find-evidence-of-what-is-causing-the-slow-requests

benvand:
* Add metrics
* Add tests

https://trello.com/c/7NhpkDWC/801-metrics-on-buyer-and-supplier-we-need-to-find-evidence-of-what-is-causing-the-slow-requests

### DIFF
--- a/app/metrics.py
+++ b/app/metrics.py
@@ -1,0 +1,9 @@
+from flask import Blueprint
+from dmutils.metrics import DMGDSMetrics
+
+
+metrics = Blueprint('metrics', __name__)
+
+gds_metrics = DMGDSMetrics()
+
+metrics.add_url_rule(gds_metrics.metrics_path, 'metrics', gds_metrics.metrics_endpoint)
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,8 +14,8 @@ git+https://github.com/alphagov/digitalmarketplace-apiclient.git@19.15.1#egg=dig
 ## The following requirements were added by pip freeze:
 asn1crypto==0.24.0
 blinker==1.4
-boto3==1.9.134
-botocore==1.12.134
+boto3==1.9.139
+botocore==1.12.139
 certifi==2019.3.9
 cffi==1.12.3
 chardet==3.0.4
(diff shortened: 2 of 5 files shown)


app/metrics.py
```diff
@@ -0,0 +1,9 @@
+from flask import Blueprint
```
katstevens: On Brief Responses FE we put this in `app/metrics/__init__.py` while the APIs use `app/metrics.py` - it probably doesn't matter which one we use, but should we pick one for consistency?

benvand: Yeah, I think `app/metrics.py` is more pythonic.

benvand: https://github.com/alphagov/digitalmarketplace-brief-responses-frontend/pull/117

requirements.txt
```diff
@@ -14,8 +14,8 @@ git+https://github.com/alphagov/digitalmarketplace-apiclient.git@19.15.1#egg=dig
 ## The following requirements were added by pip freeze:
 asn1crypto==0.24.0
 blinker==1.4
-boto3==1.9.134
-botocore==1.12.134
+boto3==1.9.139
+botocore==1.12.139
```
lfdebrux: It's very minor, but can we remove this bump from the commit? Just to keep things cleaner.

tests/app/helpers.py
```diff
@@ -243,6 +243,9 @@ def reset_application_company_details_confirmed_decorator(self):
 
 class BaseApplicationTest:
     def setup_method(self, method):
+        self.app_env_var_mock = patch.dict('gds_metrics.os.environ', {'PROMETHEUS_METRICS_PATH': '_metrics'})
+        self.app_env_var_mock.start()
```
lfdebrux: Does every test need the metics path envvar? I realise you've already done a load of apps this way 😅

lfdebrux:
I think this ought to do some good 👍

tests/app/helpers.py
benvand: Remembered why I did it like this now.
When we do it like so:
https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/889

It means that for most of the tests the metrics will be on the wrong endpoint. It makes our tests less representative.

Whereas the way its done here preserves the _metrics location for the whole test suite

benvand: (The first commit on that pr is an example of the way we shouldn't do it)

PR #1067, opened 2019-06-03
content loader: implement this as a thread/context-local

risicle:
Well this was unexpectedly straightforward.

https://trello.com/c/i4r9f1Hd
https://trello.com/c/RW2tzjbM

I was at least expecting the mocks to break.

The content loader isn't threadsafe, so the easiest avenue towards thread-enabling an app is probably to supply each thread/context with its own copy of the content loader. the thread that manages to perform the init_app also gets to pre-load its content, in the process performing the init-time content checking that we're used to. other threads will construct their content loader on first access (it's probably quite tricky to get them to pre-load too)

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -17,6 +17,7 @@ csrf = CSRFProtect()
 
 from app.main.helpers.services import parse_document_upload_time
 from app.main.helpers.frameworks import question_references
+from app.main import get_content_loader
 
 
 def create_app(config_name):
@@ -62,6 +63,10 @@ def create_app(config_name):
 
     application.add_template_filter(question_references)
     application.add_template_filter(parse_document_upload_time)
+
+    # the lucky thread that gets to do create_app also gets to pre-load its content_loader, in the process
+    # asserting that all required content files are present and in good order
+    get_content_loader()
 
     return application
 
--- a/app/main/__init__.py
+++ b/app/main/__init__.py
@@ -3,6 +3,8 @@ from werkzeug.local import Local, LocalProxy
 
 from dmcontent.content_loader import ContentLoader
 
+from dmutils.timing import logged_duration
+
 main = Blueprint('main', __name__)
 
 # we use our own Local for objects we explicitly want to be able to retain between requests but shouldn't
@@ -10,6 +12,7 @@ main = Blueprint('main', __name__)
 _local = Local()
 
 
+@logged_duration(message="Spent {duration_real}s in get_content_loader")
 def get_content_loader():
     if hasattr(_local, "content_loader"):
         return _local.content_loader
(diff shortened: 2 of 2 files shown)


risicle:
I don't know if I should combine it with some apiclient changes before merging...

katstevens:
I've just been reading the werkzeug `Local` docs, interesting! https://werkzeug.palletsprojects.com/en/0.15.x/local/ This looks pretty nifty to me. I do see however that the Buyer FE handles the content loader init significantly differently (though whether it needs to do that or not is another question 🤔). And I'd like to have a consistent approach for the FE apps wherever possible. Do you see any potential gotchas if we reuse this there?

risicle:
Yeah I haven't scouted other apps yet, the biggest potential for breakage is probably mocking.

risicle:
Looking at the buyer fe it looks like it should actually be more straightforward.

app/__init__.py
```diff
@@ -63,6 +64,10 @@ def refresh_session():
     application.add_template_filter(question_references)
     application.add_template_filter(parse_document_upload_time)
 
+    # the lucky thread that gets to do create_app also gets to pre-load its content_loader, in the process
+    # asserting that all required content files are present and in good order
+    get_content_loader()
+
     return application
```
benvand: > the thread that manages to perform the init_app also gets to pre-load its content

I can't find any documentation on this. 

I was under the understanding that requests are initialised in a new thread in the same process with a new app context.

I may be missing when the threads are spawned though as the documentation on that is super sparse

risicle: Yep the documentation properly sucks.

The threads actually get spawned by `uwsgi` (https://github.com/unbit/uwsgi/blob/b6308cae818dab78da5f51eae8c903b6e2122b7a/plugins/python/python_plugin.c#L1388) and are each just handed requests as they come. The "app init" code though, if you trace it back, gets called at import-time. Imports are locked and done serially in python so they work more or less "by magic" and you don't typically have to worry about the thread safety of things you do at import-time. When the first thread gets handed a request it will need to perform all the imports required to set up & access the "application" object.

When a second thread gets handed a request it will these imports will have already been performed and that thread will be able to access that module's attributes like a big dict. Normal module import behaviour really, just ensuring two threads can't be trying to execute import-time code at the same time.

risicle: But really, the amount we're supposed to worry about this kind of stuff is "not much". The basic `uwsgi` flask examples really just seem to enable threads and fuhgettaboutit. We're only having to worry a bit because we've got mutable globals, which is generally a bad idea in a webapp altogether.

risicle:
Ahh so this _works_ fine. The annoying thing though is constructing the `content_loader` for a thread which hasn't yet got an initialized one can take like 30s. Which would be a kinda annoying penalty to slap on to random requests.

Might give this another go with a deepcopy-based approach instead.

risicle:
This needs to be put back in the oven.

PR #1080, opened 2019-06-25
DOS4 declaration word limit

katstevens:
https://trello.com/c/Bw46fBz5/559-declaration-questions-for-mitigatingfactors-have-no-maxlengthinwords

We don't yet use JSON schema validation for declaration questions 😿 (there's already a tech debt ticket for that...)

I started a new test module for the `SharedValidator` so we don't have to keep tweaking G7/G8/DOS 1 submissions.

Tests will fail until https://github.com/alphagov/digitalmarketplace-frameworks/pull/581 is merged.

### DIFF
--- a/app/main/helpers/validation.py
+++ b/app/main/helpers/validation.py
@@ -101,9 +101,12 @@ class DeclarationValidator(object):
     def word_limit_errors(self):
         errors_map = {}
         for question_id in self.all_fields():
-            if self.content.get_question(question_id).get('type') in ['text', 'textbox_large']:
+            question = self.content.get_question(question_id)
+            if question.get('type') in ['text', 'textbox_large']:
+                # Get word limit from question content, fall back to class attribute
+                word_limit = question.get('max_length_in_words', self.word_limit)
                 answer = self.answers.get(question_id) or ''
-                if self.word_limit is not None and len(answer.split(" ")) > self.word_limit:
+                if word_limit is not None and len(answer.split()) > word_limit:
                     errors_map[question_id] = "under_word_limit"
 
         return errors_map
--- a/app/main/helpers/validation.py
+++ b/app/main/helpers/validation.py
@@ -22,6 +22,7 @@ class DeclarationValidator(object):
     email_validation_fields = []
     number_string_fields = []
     character_limit = None
+    word_limit = None
     optional_fields = set([])
 
     def __init__(self, content, answers):
@@ -72,6 +73,7 @@ class DeclarationValidator(object):
     def errors(self):
         errors_map = {}
         errors_map.update(self.character_limit_errors())
+        errors_map.update(self.word_limit_errors())
         errors_map.update(self.formatting_errors(self.answers))
         errors_map.update(self.answer_required_errors())
         return errors_map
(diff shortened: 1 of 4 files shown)


app/main/helpers/validation.py
```diff
@@ -96,6 +98,16 @@ def character_limit_errors(self):
 
         return errors_map
 
+    def word_limit_errors(self):
+        errors_map = {}
+        for question_id in self.all_fields():
+            if self.content.get_question(question_id).get('type') in ['text', 'textbox_large']:
+                answer = self.answers.get(question_id) or ''
```
lfdebrux:
```suggestion
                answer = self.answers.get(question_id, '')
```

app/main/helpers/validation.py
```diff
@@ -96,6 +98,16 @@ def character_limit_errors(self):
...
+                if self.word_limit is not None and len(answer.split(" ")) > self.word_limit:
```
lfdebrux: Is there particular reason to use `split(" ")` instead of just `split()`?

app/main/helpers/validation.py
```diff
@@ -96,6 +98,16 @@ def character_limit_errors(self):
...
+                    errors_map[question_id] = "under_word_limit"
```
lfdebrux: I'm confused, why is this error called `under_word_limit`?

app/main/helpers/validation.py
```diff
@@ -243,6 +255,7 @@ def get_required_fields(self):
 class SharedValidator(DOSValidator):
     # From DOS2 and G8 onwards, validate DUNS number length
     number_string_fields = [('dunsNumber', 9)]
+    word_limit = 500
```
lfdebrux: Is there no way we can pull this value from the question?

app/main/helpers/validation.py
katstevens: It matches the existing `under_character_limit`.

app/main/helpers/validation.py
katstevens: Good ol' copy and paste there! Will amend

app/main/helpers/validation.py
katstevens: Yes, probably. However I think a catch-all limit for all questions would be useful, in case people forget to add a specific limit to future free-text questions.

app/main/helpers/validation.py
katstevens: Haha none whatsoever 🤦‍♀ Will fix

app/main/helpers/validation.py
lfdebrux: Any chance we can change both to `over_*_limit`? Or just `*_limit`?

app/main/helpers/validation.py
katstevens: The validator now checks whether `max_length_in_words` is available.

app/main/helpers/validation.py
katstevens: Ah - we do need this, as `question_id` may be present in `self.answers` but still return `None`, and we want to validate that.

lfdebrux: Right, got it, my bad

PR #1113, opened 2019-08-27
consistently use current_user.email_address for updated_by in API requests

risicle:
https://trello.com/c/IvrgIUND

This was quick & simple enough. Was going to do a total "audit" of places we were screwing this up by searching the prod database's audit table for non-`@`-containing `updated_by`s but realized we don't use email addresses in most of our script/manual updates.


lfdebrux:
So you think this is the only place?

risicle:
Can't be sure. Can't think of a quick & easy way of finding out reliably.

PR #1114, opened 2019-09-03
Update NodeJS from version 8 to version 10

lfdebrux:
https://trello.com/c/YknBKmMU/261-update-node-from-8-to-10


.travis.yml
```diff
@@ -2,7 +2,7 @@ language: python
 python:
   - "3.6"
 before_install:
-  - nvm install 6.12.2
+  - nvm install
```
aliuk2012: Lets see if we can get Travis to reuse `.nvmrc` instead of hard coding a version

PR #1127, opened 2019-10-24
Upgrade Jquery to 3.x

katstevens:
https://trello.com/c/aQbqH2uC/1196-vulnerability-fix-weekly-tracking

Addresses https://app.snyk.io/vuln/npm:jquery:20150627 and https://app.snyk.io/vuln/SNYK-JS-JQUERY-174006 as highlighted by Snyk.

We no longer need to support IE7/8. 

There's one breaking change affecting the Supplier FE: `.load(fn)` has been deprecated (see [Changelog](https://jquery.com/upgrade-guide/3.0/#breaking-change-load-unload-and-error-removed)) so and needs to be replaced with `.on('load', fn)`. 

The occurrence was in one of the old govuk_frontend_toolkit modules, so I've copied that file to `/assets` and made the change to the old `$(global).load(function (), ...)` call.

## Context
The checkbox-tree component itself is only used in the supplier framework application journey (when selecting service categories), so the next time this will be used is March 2020. It's unclear how well this component currently meets WCAG guidelines but anecdotally it's not a great experience to click or scroll through anyway, even for developers. Our long term goal should be to upgrade or replace this component, and work with CCS to potentially reduce the number of categories on the page at any one time.

### DIFF
--- a/app/assets/javascripts/_stop-scrolling-at-footer.js
+++ b/app/assets/javascripts/_stop-scrolling-at-footer.js
@@ -1,0 +1,141 @@
+// Stop scrolling at footer.
+//
+// This can be added to elements with `position: fixed` to stop them from
+// overflowing on the footer.
+//
+// Usage:
+//
+//    GOVUK.stopScrollingAtFooter.addEl($(node), $(node).height());
+//
+// Height is passed in separatly incase the scrolling element has no height
+// itself.
+
+// Copied from original GOVUK frontend toolkit to add Jquery 3.x support for .load(...) -> .on('load', ...)
+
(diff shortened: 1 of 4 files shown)


lfdebrux:
Is there any reason not to pull the `_stop_scrolling_at_footer.js` file into frontend toolkit?

katstevens:
@lfdebrux I'd rather not encourage the use of this particular pattern elsewhere if we're thinking about retiring it! Even though we technically use the checkbox tree component on the Admin FE, we don't have the accompanying javascript there anyway.

PR #1133, opened 2019-11-11
switch most buttons to use digitalmarketplace-govuk-frontend

risicle:
https://trello.com/c/el9OyFgd

Don't know what else to say here.

### DIFF
--- a/app/assets/scss/application.scss
+++ b/app/assets/scss/application.scss
@@ -79,6 +79,9 @@ $govuk-compatibility-govukelements: true;
 @import "_updates.scss";
 @import "_view-service-link.scss";
 
+// Overrides
+@import "overrides/_notifications_banner";
+
 // Misc styles
 // TODO: Move misc styling into their own partial files or the Digital Marketplace FE Toolkit
 #wrapper, .wrapper {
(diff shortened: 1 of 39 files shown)


lfdebrux:
> switch most buttons to use digitalmarketplace-govuk-frontend

Do you have a list of the buttons not switched?

risicle:
> Do you have a list of the buttons not switched with reasons why?

It's just buttons that are generated through other means, like `toolkit/temporary-message.html`

lfdebrux:
@risicle what about these three buttons?

```
gds5506:digitalmarketplace-supplier-frontend laurencedebruxelles$ rg 'toolkit/button'
app/templates/suppliers/details.html
190:            {% include "toolkit/button.html" %}

app/templates/suppliers/registered_address.html
87:      {% include "toolkit/button.html" %}

app/templates/suppliers/create_company_summary.html
98:  {% include "toolkit/button.html" %}
```

risicle:
Actually there are a couple of other things I haven't checked too...

risicle:
Ok I think I've got them all now - there don't seem to be any uses of `<input type="submit"...`.

PR #1139, opened 2019-12-03
This replaces toolkit/page-headings with govuk frontend page titles

agalamatis:
trello: https://trello.com/c/9LvUzYaP/158-2-update-page-headings-to-govuk-frontend-style-in-supplier-frontend

### DIFF
--- a/app/templates/frameworks/contract_review.html
+++ b/app/templates/frameworks/contract_review.html
@@ -30,13 +30,6 @@
 
 <div class="grid-row">
   <div class="column-two-thirds">
-    {%
-      with
-      heading = "Check the details you’ve given before returning the signature page for %s" | format(supplier_registered_name),
-      smaller = True
-    %}
-      {% include "toolkit/page-heading.html" %}
-    {% endwith %}
     <h1 class="govuk-heading-l">{{ 'Check the details you’ve given before returning the signature page for {}'.format(supplier_registered_name) }}</h1>
   </div>
 </div>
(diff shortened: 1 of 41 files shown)


app/templates/frameworks/contract_review.html
```diff
@@ -37,6 +37,7 @@
     %}
       {% include "toolkit/page-heading.html" %}
     {% endwith %}
```
lfdebrux: Lines above need to be removed

lfdebrux:
Appears to be a problem with spacing and banners on the framework dashboard page, even though the override has been added correctly 🤔 

image

Any thoughts @aliuk2012?

agalamatis:
@lfdebrux / @aliuk2012 I think I found the issue, fix in the last commit.

lfdebrux:
> @lfdebrux / @aliuk2012 I think I found the issue, fix in the last commit.

Ah, good spot, unfortunately the issue is still there (because the banner is inside the #content div).

lfdebrux:
Missed one here also 

https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1139/files#diff-d52d46ec4d8bd41c04744e59862dce32R93-R99

agalamatis:
> Missed one here also
> 
> https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1139/files#diff-d52d46ec4d8bd41c04744e59862dce32R93-R99

fixed, pushed

lfdebrux:
> > @lfdebrux / @aliuk2012 I think I found the issue, fix in the last commit.
> 
> Ah, good spot, unfortunately the issue is still there (because the banner is inside the #content div).

Looks like we can apply patch https://github.com/alphagov/digitalmarketplace-admin-frontend/commit/52e9b1550369ddfaaaa8603d8b60940c7a176f66 from admin frontend to fix the notification banners

PR #1148, opened 2020-01-09
count_unanswered_questions: use implementation from dmcontent

risicle:
This function got moved to `dmcontent` in the course of https://trello.com/c/g5P0RxQn


risicle:
Er no didn't see it.

PR #1149, opened 2020-01-16
248-update-content-on-duns-number-page

benvand:
https://trello.com/c/6vSysI7W/248-update-content-on-duns-number-page

* Update create_duns_number.html with content suggestions
  * https://docs.google.com/presentation/d/1sukhMNNbGTl3KDYYVedtBcNT5gQliaXjbrGW5Mrf2ak/edit#slide=id.g759a4b2e43_0_2
  * Discussion with Nora
* Update error messages on supplier registration forms as per:
  * Content guidlines here:
    * https://design-system.service.gov.uk/components/error-message/
  * Discussion with Nora:
    * https://docs.google.com/presentation/d/1sukhMNNbGTl3KDYYVedtBcNT5gQliaXjbrGW5Mrf2ak/edit#slide=id.g759a4b2e43_0_2

PR #1150, opened 2020-01-16
212-supplier-details-lookup-page

benvand:
https://trello.com/c/b9iVLjxZ/212-supplier-details-lookup-page

Add new view, template and form for supplier details lookup page.
Created inline with prototype here:
https://www.draw.io/#G1WTubaHiJXizYMuGu0XroQH0iza0ZwlMv

Awaiting: https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1149

### Required to pass tests:
https://github.com/alphagov/digitalmarketplace-utils/pull/552

### Tests pass locally

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -13,7 +13,7 @@ from dmutils.flask import timed_render_template as render_template
 from dmutils.forms.helpers import get_errors_from_wtform
 from dmutils.errors import render_error_page
 
-from ...main import main, content_loader
+from ...main import main, content_loader, direct_plus_client
 from ... import data_api_client
 from ..forms.suppliers import (
     AddCompanyRegisteredNameForm,
@@ -25,6 +25,7 @@ from ..forms.suppliers import (
     EditRegisteredAddressForm,
     EditSupplierInformationForm,
     EmailAddressForm,
+    LookupConfirmForm,
 )
 from ..helpers.frameworks import (
     get_frameworks_by_status,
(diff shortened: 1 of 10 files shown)


app/main/views/suppliers.py
```diff
@@ -512,25 +513,49 @@ def create_new_supplier():
...
+        # Fail on existing account with this duns number first
+        if len(data_api_client.find_suppliers(duns_number=form.duns_number.data)['suppliers']) > 0:
             form.duns_number.errors = ["DUNS number already used"]
+        else:
+            # Check for number in Direct Plus API
+            # If we cannot contact the API then skip this step
+            # If the client doesn't return an organization then there is not a record for that DUNs
+            # Otherwise the number is good, add the 'primaryName' to the session and continue
+            try:
+                organization = direct_plus_client.get_organization_by_duns_number(form.duns_number.data)
```
risicle: _I_ approve of `organization` with a Z (I would personally use american spelling in all variable names) but note that we seem to use `organisation` throughout the rest of the code.

risicle: Of course the cheat to avoid the subject would be to just call it `org` everywhere

benvand: Yeah, I've kept it in link with what it's named in the response from DnB. American throughout would probably not be a bad idea.

app/main/views/suppliers.py
```diff
@@ -512,25 +513,49 @@ def create_new_supplier():
...
-        if form.validate_on_submit():
-            suppliers = data_api_client.find_suppliers(duns_number=form.duns_number.data)
-            if len(suppliers["suppliers"]) <= 0:
-                session[form.duns_number.name] = form.duns_number.data
-                return redirect(url_for(".company_details"))
+    if request.method == 'GET':
+        if request.args.get('retry', None) == 'true':
+            # When redirected back from the lookup page
+            form.duns_number.errors = ["Enter a different DUNS number"]
+        elif form.duns_number.name in session:
```
risicle: Thinking out loud here, but is it worth making this just an if, so that the user gets shown the existing duns number they entered even if they get the Enter a different DUNS number? So they can see any obvious errors they made or if they have multiple possible DUNS numbers to try they can see which one to not try again or... iunno.

app/main/views/suppliers.py
```diff
@@ -512,25 +513,49 @@ def create_new_supplier():
...
+            # Otherwise the number is good, add the 'primaryName' to the session and continue
+            try:
+                organization = direct_plus_client.get_organization_by_duns_number(form.duns_number.data)
+                if organization is not None:
+                    company_name = organization['primaryName']
+            except KeyError:
+                # An unexpected error. Something other than supplier data in the response. Skip this part of sign up.
+                return redirect(url_for(".company_details"))
+
+            organization = direct_plus_client.get_organization_by_duns_number(form.duns_number.data)
```
risicle: Why are we doing this call again when we've potentially just fallen out of the above `try` block where we already do this and why is this call not in a `try` block? What happens if this one fails?

benvand: Good catch, merge conflict thing I think

app/main/views/suppliers.py
```diff
@@ -541,6 +564,33 @@ def duns_number():
     ), 200 if not errors else 400
 
 
+@main.route('/create/lookup', methods=['GET', 'POST'])
+def lookup():
```
risicle: From a dev's angle at 1000ft, I _would_ have expected the `lookup` view to be the thing that does the _lookup_, rather than being the view that just does the confirmation.

benvand: Could change the view function name to be `lookup_result`?

tests/app/main/test_suppliers.py
```diff
@@ -1714,55 +1716,71 @@ def test_should_be_an_error_if_missing_or_invalid_duns_number(self, duns_number)
...
-            ]
-        }
-        res = self.client.post(
-            "/suppliers/create/duns-number",
-            data={
-                'duns_number': "123456789"
-            }
-        )
+        with self.app.app_context():
+            with mock.patch(self.direct_plus_api_method, return_value={'primaryName': 'COMPANY LTD'}):
```
risicle: Not going to assert the calls made to this mock?

risicle: Same for `data_api_client`

benvand: Yeah, will write another test for that methinks.

benvand: @risicle 
https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1150/commits/3310e7ecf1077928ade6354f76a793b324d8eb35

app/main/views/suppliers.py
benvand: `lookup_confirm` 👍

risicle:
How do we (interactively) test this locally?

benvand:
> Seems right to me, haven't tested in-person (can I?)

Yeah, totally possible, update the `DM_DNB_API_USERNAME` and `DM_DNB_API_PASSWORD` variables in `config.py` with the credentials from digitalmarketplace-credentials/vars/production.yml::supplier_frontend

Then it'll work locally (don't forget to update venv with the new utils this brings in)

benvand:
Thanks for the review btw!

benvand:
Oh, it might also be worth setting a fw to coming/open directly in your db

benvand:
OK, going to merge to test on preview. Will not be available to suppliers until G12 is coming

PR #1152, opened 2020-01-16
Replace main_content with mainContent 

katstevens:
https://trello.com/c/oYTyw0pb/97-1-replace-pagetitle-maincontent-block-with-pagetitle-maincontent-in-supplier-frontend

Following the example commits in https://github.com/alphagov/digitalmarketplace-brief-responses-frontend/pull/168.

- Updates Node
- Bumps digitalmarketplace-govuk-frontend
- Find + replace for `{% block main_content %}` to `{% block mainContent %}`, putting the latter inside the former on `_base_page.html`
- Small spacing fix for the invite_contributor link 
- Small sizing fix for some paragraphs that looked weirdly huge compared to their parent `h2`
- Styling fix for the draft service action buttons (see pics below - the buttons are only visible when frameworks are open). This style is still part of the custom CSS that we eventually want to delete, but I don't want to start revamping that in this PR - merely getting the button style back to where it was. 

Before:
draft-service-action-buttons-old

After:
draft-service-action-buttons-new

(I changed the text on the button because 'Delete' felt weirdly ambiguous - delete what? Happy to put it back if people feel strongly)

### DIFF
--- a/app/templates/services/service_submission.html
+++ b/app/templates/services/service_submission.html
@@ -125,10 +125,10 @@
         <input type="hidden" name="delete_confirmed" value="true" />
         <div class="banner-destructive-with-action">
           <p class="banner-message">
-            Are you sure you want to delete {{ service_data.lotName.lower() if lot.oneServiceLimit else 'this {}'.format(lot.unitSingular) }}?
+            Are you sure you want to remove {{ service_data.lotName.lower() if lot.oneServiceLimit else 'this {}'.format(lot.unitSingular) }}?
           </p>
           {{ govukButton({
-            "text": "Yes, delete",
+            "text": "Yes, remove",
             "classes": "govuk-button--warning app-banner-action",
           }) }}
         </div>
--- a/app/templates/services/service_submission.html
+++ b/app/templates/services/service_submission.html
@@ -187,7 +187,7 @@
 
 {% block after_sections %}
   {% if not delete_requested %}
-    <div class="grid-row">
+
       <div class="column-two-thirds">
         &nbsp;
       </div>
@@ -201,8 +201,8 @@
         <form action="{{ url_for('.delete_draft_service', framework_slug=framework.slug, lot_slug=service_data.lot, service_id=service_id ) }}" method="POST">
           <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
           {{ govukButton({
-            "text": "Delete",
-            "classes": "govuk-button--warning",
+            "text": "Delete draft service",
+            "classes": "govuk-button--warning button-destructive",
           }) }}
         </form>
         {% endif %}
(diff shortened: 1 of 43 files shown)


app/assets/scss/_service_submission.scss
```diff
@@ -32,8 +32,8 @@
 
   margin-top: $gutter;
 
-  .button-destructive,
-  .button-save {
+  .govuk-button--warning,
+  .govuk-button {
```
aliuk2012: Instead of changing these and modifying govuk-frontend styling, how about either keeping these class names or rename them to something else and add those class names to the govukButton params.

For example, if we were to keep the class names the same:
```
{{ govukButton({
	"text": "Delete",
    "text": "Delete draft service",
    "classes": "govuk-button--warning button-destructive",
}) }}

```

The benefit is that we are not patching govuk-frontend and its less likely that if 
govuk-front changes the buttons it will not break our button styling

app/templates/services/service_submission.html
```diff
@@ -187,7 +187,7 @@ <h2 class="temporary-message-heading" id="temporary-message-heading">
 
 {% block after_sections %}
   {% if not delete_requested %}
-    <div class="grid-row">
```
aliuk2012: why is this being deleted? if we using a column class then it should be inside a grid row. 🤔 not sure if it might be worth renaming the classes here to use govuk-frontend classes but I'm happy if you decide to leave it for another ticket

app/templates/services/service_submission.html
```diff
@@ -201,7 +201,7 @@ <h2 class="temporary-message-heading" id="temporary-message-heading">
         <form action="{{ url_for('.delete_draft_service', framework_slug=framework.slug, lot_slug=service_data.lot, service_id=service_id ) }}" method="POST">
           <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
           {{ govukButton({
-            "text": "Delete",
+            "text": "Delete draft service",
```
aliuk2012: I think this change. This is much clearer 👏

app/templates/services/service_submission.html
katstevens: It already is inside a grid row! I spotted some weird alignment when QAing locally and found it was extending `grid-row` from `_base_service_page.html`. (I should have mentioned this in the PR description, whoops.)

app/templates/services/service_submission.html
katstevens: I'll run it past Nora to make sure.

app/assets/scss/_service_submission.scss
katstevens: Cool - have used your suggestion and amended the last commit, thanks Al 👍

app/templates/services/service_submission.html
katstevens: Nora's suggested a change - I've added a couple of commits.

PR #1159, opened 2020-01-23
🚀 Replacing govuk_template ☠️

katstevens:
https://trello.com/c/q3iJkiZI/249-2-remove-govuktemplate-from-supplier-frontend

- Pulls in latest `digitalmarketplace-govuk-frontend`
- Switch base template to use GOV.UK Frontend base template (some Python tests needed updating)
- Removes unused user research banner / report a problem scss
- Includes fallback styles for basic elements not yet upgraded (links etc)
- Removes temporary hacks needed to ensure a smooth transition 
- Removes dependency on old GOV.UK Template (remove gulp copying tasks)

Some styling for the framework application journey has changed (font colours inside notification banners are now black instead of green/red). We'll review this as part of the G12 work. 

## Before
supplier-fe-govuk-template-before

## After
supplier-fe-govuk-template-after


lfdebrux:
I think some assets might be missing :( 
image

app/templates/_base_page.html
```diff
@@ -1,26 +1,59 @@
...
-  </div>
+  {% include "toolkit/flash_messages.html" %}
+  {% block errorSummary %}
+    {% if errors %}
+      {{ govukErrorSummary({
+      "titleText": "There is a problem",
+      "errorList": errors.values(),
+    }) }}
+    {% endif %}
+  {% endblock %}
```
lfdebrux: This block shouldn't be here yet, otherwise we end up with duplicated error summaries.

katstevens: Copied it from https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/master/app/templates/_base_page.html#L41 - does that need updating too?

lfdebrux: No, admin frontend is slightly ahead, there we've replaced the old validation banners with error summaries

https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/600

katstevens:
> I think some assets might be missing :(
Ah the footer image? Will take a look

app/templates/_base_page.html
```diff
@@ -1,26 +1,59 @@
...
+{% from "govuk-frontend/components/input/macro.njk" import govukInput %}
+{% from "govuk-frontend/components/phase-banner/macro.njk" import govukPhaseBanner %}
 
-{# TODO: Remove once we start removing govuk_template, GOV.UK Frontend template uses pageTitle #}
-{% block page_title %}{% block pageTitle %}{% endblock %}{% endblock %}
+{# Import DM components #}
+{% from "digitalmarketplace/components/header/macro.njk" import dmHeader %}
+{% from "digitalmarketplace/components/footer/macro.njk" import dmFooter %}
+
+{% set assetPath = '/user/static' %}
```
lfdebrux: I think this line might be the cause of the assets issue

katstevens: It certainly covered it up (as my local env was happily using the User FE assets). Now fixed.

lfdebrux:
The country picker seems to have stopped working 🤔 

image

lfdebrux:
Okay, let's put this onto preview and give it a whirl!

PR #1165, opened 2020-01-29
Update link text for 'G-Cloud supplier A to Z' for footer link

lfdebrux:
Ticket: https://trello.com/c/i4eJBbNC/274-replace-content-supplier-a-z-with-supplier-a-to-z

Updates digitalmarketplace-govuk-frontend from 0.3.1 to 0.4.1, pulling in the new footer.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2096,9 +2096,9 @@
       }
     },
     "digitalmarketplace-govuk-frontend": {
-      "version": "0.3.1",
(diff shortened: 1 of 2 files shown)


lfdebrux:
> I think we only test the footer link text from the homepage (Buyer FE), so the functional tests should still pass for this app?

Yes, I think so, I couldn't find any other references to the link text in the functional tests repo.

PR #1166, opened 2020-01-31
Fix styles for temporary message text

lfdebrux:
Ticket: https://trello.com/c/lbW5IQFx/

Copied from alphagov/digitalmarketplace-buyer-frontend#982

The framework pending message (if you've started a framework application but not completed it) also has the black text on dark blue background problem; using @katsteven's fix worked a treat :)

Also I cleaned up a temporary message component instance from 2015, and made a small tweak to our gulpfile to make it more dev friendly.

#### After

Supplier account dashboard, for a supplier that had started an application to G-Cloud 12 but not completed it, with G-Cloud 12 in the pending state. After the fix is applied the text explaining that the supplier did not submit an application is easy to read against the dark background.

#### Before

Supplier account dashboard, for a supplier that had started an application to G-Cloud 12 but not completed it, with G-Cloud 12 in the pending state. Before the fix is applied the text explaining that the supplier did not submit an application is difficult to read against the dark background.

### DIFF
--- a/app/templates/suppliers/_frameworks_coming.html
+++ b/app/templates/suppliers/_frameworks_coming.html
@@ -1,19 +1,3 @@
 {% for framework in frameworks.coming %}
-  {% if framework.slug == 'digital-outcomes-and-specialists' %}
-
-    {% set blog_link %}
-    <a class="govuk-link" href='https://digitalmarketplace.blog.gov.uk/2015/11/12/digital-outcomes-and-specialists-suitable-suppliers-and-services/'>Find out if your services are suitable</a>
-    {% endset %}
-
-    {%
-      with
-      messages = [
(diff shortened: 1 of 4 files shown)


app/templates/suppliers/_frameworks_coming.html
```diff
@@ -1,19 +1,3 @@
 {% for framework in frameworks.coming %}
-  {% if framework.slug == 'digital-outcomes-and-specialists' %}
-
-    {% set blog_link %}
```
katstevens: `blog_link`! I thought we removed all these 😹 Good spot

lfdebrux: I couldn't believe my eyes 😂

PR #1168, opened 2020-02-04
Bump expected node version to 10.18.1

risicle:
https://trello.com/c/25rCohcs

For docker images and development setups.


package.json
```diff
@@ -4,7 +4,7 @@
   "version": "1.0.0",
   "private": true,
   "engines": {
-    "node": "^10.16.3"
+    "node": "^ 10.18.1"
```
lfdebrux: I'm not sure whether the space is allowed, the npm documentation suggests not https://docs.npmjs.com/misc/semver#range-grammar

risicle: This is simply because the repo I created the initial commit on had a space. It wasn't a conscious decision on my part. I'll strip this one but I think I've already committed some with a space.

lfdebrux: Yeah it happens

PR #1169, opened 2020-02-07
Add shared validator for G12

katstevens:
https://trello.com/c/ypNRnhEP/259-update-declaration-validation-code-in-supplier-fe-2

- Explicitly set `g-cloud-12` to use the `SharedValidator`. This is in case we want to switch to a custom validator at a later point (e.g. to deal with any dependency change in the first two questions)
- Use the `SharedValidator` as the default, so future frameworks won't immediately break

### DIFF
--- a/app/main/helpers/validation.py
+++ b/app/main/helpers/validation.py
@@ -271,4 +271,5 @@ VALIDATORS = {
     "digital-outcomes-and-specialists-3": SharedValidator,
     "g-cloud-11": SharedValidator,
     "digital-outcomes-and-specialists-4": SharedValidator,
+    "g-cloud-12": SharedValidator,
 }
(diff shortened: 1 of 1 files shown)


app/main/helpers/validation.py
```diff
@@ -14,7 +14,7 @@ def get_validator(framework, content, answers):
     if framework is None:
         raise ValueError("a framework dictionary must be provided")
     if framework is not None:
-        validator_cls = VALIDATORS.get(framework['slug'])
+        validator_cls = VALIDATORS.get(framework['slug'], SharedValidator)
```
lfdebrux: Nice idea!

katstevens: I will update the manual to say that we now only need to do this if there's a custom validator.

PR #1173, opened 2020-02-11
258 update questions in frameworks repo 2

benvand:
https://trello.com/c/EoMme9Md/258-update-questions-in-frameworks-repo-2

See commits

#### Depends on:
https://github.com/alphagov/digitalmarketplace-frameworks/pull/605/files

screencapture-localhost-suppliers-frameworks-g-cloud-12-declaration-edit-providing-suitable-services-2020-02-11-08_26_45

### DIFF
--- a/app/main/helpers/validation.py
+++ b/app/main/helpers/validation.py
@@ -261,6 +261,27 @@ class SharedValidator(DOSValidator):
     word_limit = 500
 
 
+class G12Validator(SharedValidator):
+
+    def errors(self):
+
+        errors_map = super().errors()
+
+        q1_answer = self.answers.get('servicesHaveOrSupportCloudHostingCloudSoftware')
+        q2_answer = self.answers.get('servicesHaveOrSupportCloudSupport')
+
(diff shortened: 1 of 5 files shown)


lfdebrux:
Hmm, I think I agree with @risicle on this one, if we're doing validation at the time of the declaration then these questions are at best redundant and at worst confusing and contradictory.

If we are able to I think we should instead ask these questions about each and every service submitted.

risicle:
Of course, something being the right way to do something and us actually agreeing it with our wonderful partners and making the technical changes required for it in time are two different things...

benvand:
@risicle @lfdebrux 
This is out of scope of this PR.

The 'declaration' journey step needs a rethink form the ground up.

benvand:
Sorry, that sounds like a snap, it's not, it's just me and @katstevens both raised this when the ask came in and got nowhere.

lfdebrux:
> Sorry, that sounds like a snap, it's not, it's just me and @katstevens both raised this when the ask came in and got nowhere.

Understood

PR #1179, opened 2020-02-17
318-content-check-for-new-error-messages-for-new-declaration-questions

benvand:
Bring in new frameworks version:
https://github.com/alphagov/digitalmarketplace-frameworks/pull/607/files


benvand:
Master is now at fws version 17.6.0 We no longer need this!

PR #1184, opened 2020-02-19
enclosing button and anchor in paragraphs

agalamatis:
trello: https://trello.com/c/ZabEvfWC/337-1-clarification-question-submit-form-link-spacing

### DIFF
--- a/app/templates/frameworks/updates.html
+++ b/app/templates/frameworks/updates.html
@@ -146,7 +146,7 @@
                 "text": "Ask a question"
               }) }}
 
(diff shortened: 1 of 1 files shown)


agalamatis:
§
p-closure

katstevens:
I think we used a class last time (https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1143) - is there a reason that won't work here?

agalamatis:
> I think we used a class last time (#1143) - is there a reason that won't work here?

Switch to the class that turns the anchor into a block.

PR #1194, opened 2020-02-25
1480-benchmark-suppliers-frameworks-g-cloud-12-submissions-cloud-hosting

benvand:
Perform loading of content for lot services outside of loop
Lot is always the same in this view

Benchmarking results show improvement:
https://docs.google.com/spreadsheets/d/1yqdHPAxG1WmXMnAP6BF44jt-vlF_q8PxVsKmBuGhqMs/edit#gid=0

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -345,7 +345,7 @@ def framework_submission_services(framework_slug, lot_slug):
     ).filter(context={'lot': lot_slug}, inplace_allowed=True)
 
     with logged_duration(message="Annotated draft details in {duration_real}s"):
-        for draft in chain(drafts, complete_drafts):
+        for draft in drafts:
             sections = lot_service_sections.summary(draft, inplace_allowed=True)
 
             unanswered_required, unanswered_optional = count_unanswered_questions(sections)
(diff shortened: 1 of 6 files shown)


app/main/views/frameworks.py
```diff
@@ -340,13 +339,14 @@ def framework_submission_services(framework_slug, lot_slug):
...
+
     with logged_duration(message="Annotated draft details in {duration_real}s"):
-        for draft in chain(drafts, complete_drafts):
-            draft['priceString'] = format_service_price(draft)
-            sections = content_loader.get_manifest(
-                framework_slug,
-                'edit_submission',
-            ).filter(draft, inplace_allowed=True).summary(draft, inplace_allowed=True)
+        for draft in drafts:
+            sections = lot_service_sections.summary(draft, inplace_allowed=True)
```
risicle: I'm a _little bit_ nervous about this change, especially used in conjunction with `inplace_allowed`. I didn't make this optimization because we're now mutating the same `ContentManifest` over and over again, overwriting it with each new service's details. And with the obfuscated design of the content loader I couldn't be certain that no data leaks from one iteration to the next.

PR #1198, opened 2020-03-02
config: re-enable WTF_CSRF_ENABLED for staging

risicle:
https://trello.com/c/085wlnUP

This was presumably disabled to allow gatling scripts to run, but we now have a mechanism in place that allows gatling to work properly with csrf.


risicle:
I'll hold off on merging this until people have stopped playing with gatling for this round.

PR #1200, opened 2020-03-03
Handle pre-existing file uploads in declaration data

katstevens:
Bug found during G12 QA: if a supplier goes back and edits the Modern Slavery statement (multiquestion) section, then re-submits, it gives an error that the existing file hasn’t been uploaded. This is because the form macro doesn't set the value on the file upload input (for good reason - it's a security hole).

Therefore we need to address this on the validation side. I can recreate it with the new unit test. It's a complete hack, given the short timescale - better suggestions very welcome!  

g12-declaration-modern-slavery-file-upload-error


risicle:
Obviously not ideal, but probably the most sensible thing we'll get with such short notice. Tests still running...

PR #1201, opened 2020-03-03
edit_submission_section template: fix switched breadcrumb links

risicle:
Oops. these links were the wrong way round.


katstevens:
I'm creating a tech debt ticket to rename these views to something less confusing...

PR #1202, opened 2020-03-03
Use dynamic guidance link instead of hardcoded DOS link

katstevens:
Bugfix: the guidance link above the Clarification Question form points to DOS guidance (hardcoded from last year).

I've changed this to use the dynamic guidance link we use elsewhere on the Supplier FE.


katstevens:
Argh - this is missing the view context 😿

PR #1231, opened 2020-06-15
Copy services using questions_to_exclude list

katstevens:
https://trello.com/c/WBjRZd14/80-2-frameworks-repo-copyservices-list-exclude-rather-than-include-fields

Implements the changes in https://github.com/alphagov/digitalmarketplace-api/pull/1038.

If a `questionsToExclude` list is provided in the frameworks metadata, use that instead of the `questionsToCopy` list (which is to be deprecated).  For G12 we are still using the `questionsToCopy` list.

Notes to reviewers: 
- I tried splitting this into smaller commits, but without refactoring a bunch of stuff I couldn't get each commit in a state where the tests were passing.
- The bulk-copy API endpoint expects a list of questions, not a set or tuple, so I've changed the test data returned from the mock metadata call to reflect this.

~Functional tests PR coming shortly!~ Functional tests now merged 😸

### DIFF
--- a/app/main/helpers/services.py
+++ b/app/main/helpers/services.py
@@ -60,6 +60,7 @@ def copy_service_from_previous_framework(data_api_client, content_loader, framew
     # Suppliers must have registered interest in a framework before they can edit draft services
     if not get_supplier_framework_info(data_api_client, framework_slug):
         abort(404)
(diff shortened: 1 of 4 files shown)


lfdebrux:
### My review checklist

I've:

- [ ] looked at the code
- [ ] tested locally

PR #1232, opened 2020-06-17
Disable FlaskLogin flash messages

lfdebrux:
Ticket: https://trello.com/c/sxhsW20p/88-replace-flash-messages-with-alert-component-in-briefs-frontend

We don't want to use the default flash message for when a user needs an account to access a page, so this PR disables the message flashing in FlaskLogin.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -49,7 +49,7 @@ def create_app(config_name):
     # the external NotImplemented routes in the dm-utils external blueprint).
     application.register_blueprint(external_blueprint)
 
-    login_manager.login_message_category = "must_login"
+    login_manager.login_message = None  # don't flash message to user
     main_blueprint.config = application.config.copy()
 
     gds_metrics.init_app(application)


lfdebrux:
There is an interesting failure in the unit tests:

```
    def test_cannot_deactivate_user_unless_logged_in(self):
        res = self.client.post(
            '/suppliers/users/123/deactivate'
        )
        assert res.status_code == 302
>       assert res.location == 'http://localhost/user/login'
E       AssertionError: assert 'http://local...%2Fdeactivate' == 'http://localhost/user/login'
E         - http://localhost/user/login?next=%2Fsuppliers%2Fusers%2F123%2Fdeactivate
E         + http://localhost/user/login
tests/app/main/test_users.py:115: AssertionError
```

Looks like a small change in behaviour when using `login_manager.login_view`, POSTs now have the next param, which is not something we want?

Do we want to change the test or drop the change to `login_view`? Thoughts @katstevens?

katstevens:
@lfdebrux ughhh 😿 Let's drop it for now.

PR #1239, opened 2020-08-10
e-signature flow

gidsg:
Implements e-signature flow as per: 
https://trello.com/c/p8qUfLUA/230-3-ui-to-sign-the-fw-agreement-on-the-supplier-side-functional-tests
https://dmp-prototype-sign-agreement.herokuapp.com/

----------------------------

- [x] add legal authority question
- [x] add form validation
- [x] add no response page
- [x] add yes response page
- [x] fix breadcrumbs
- [x] add legal text
- [x] handle form submission
- [x] confirmation page
- [x] confirmation email
~~check details page `framework-signed`~~
- [x] unit tests
- [x] fix supplier dashboard


app/main/views/frameworks.py
```diff
@@ -1213,3 +1217,20 @@ def view_contract_variation(framework_slug, variation_slug):
 @login_required
 def opportunities_dashboard_deprecated(framework_slug):
     return redirect(url_for('external.opportunities_dashboard', framework_slug=framework_slug), code=301)
+
+
+@main.route('/frameworks/<framework_slug>/legal-authority', methods=['GET', 'POST'])
+@login_required
+def legal_authority(framework_slug):
```
katstevens, line 1224: We should have a check at the beginning of the view to see if the supplier has made an application or not. They can't start the signing process if their `supplierFramework` object does not have `onFramework: true`.

gidsg:
@katstevens do you know if the 'yes' response page i.e. https://dmp-prototype-sign-agreement.herokuapp.com/agreement should replace an existing route, be a new route, or be tied to a specific framework?

katstevens:
> @katstevens do you know if the 'yes' response page i.e. https://dmp-prototype-sign-agreement.herokuapp.com/agreement should replace an existing route, be a new route, or be tied to a specific framework?

I think we agreed to do new routes, so we can easily switch back to the old process if necessary.

The new process should be for iterations of G-Cloud >= 12, and DOS >= 5.

gidsg:
@katstevens Do we also want to restrict this flow to a specific framework status?

app/main/views/frameworks.py
```diff
@@ -1268,6 +1268,15 @@ def sign_framework_agreement(framework_slug):
     framework_pdf_metadata = {
         'g-cloud-12': {'file_size': '962KB', 'page_count': 65}
     }
+    _drafts, complete_drafts = get_drafts(data_api_client, framework_slug)
+    lots = [
+        dict(lot,
+             complete_count=count_drafts_by_lot(complete_drafts, lot['slug']))
+        for lot in framework['lots']]
```
katstevens, line 1275: Looking at these (fairly ancient) helper functions, I think we can now do this in a much more efficient way, without having to iterate through 30,000+ draft services (or indeed, any draft services at all!).

Something like:

```
complete_drafts = data_api_client.find_draft_services_by_framework(framework_slug, supplier_id=supplier_id, status="submitted", lot=lot)
on_lot = complete_drafts['meta']['total'] > 0
```

The snag is that we currently we can't filter this API endpoint by `lot`. But this is really easy to add! And would be really useful elsewhere. I'm happy to rustle up that PR.

katstevens: https://github.com/alphagov/digitalmarketplace-api/pull/1049

tests/app/main/test_frameworks.py
```diff
@@ -5751,3 +5750,136 @@ def test_false_values(self, falsey_value):
...
+        self.data_api_client.get_supplier_framework_info.return_value = self.supplier_framework(
+            on_framework=True)
+
+        res = self.client.get(f"/suppliers/frameworks/g-cloud-12/sign-framework-agreement")
+        assert res.status_code == 200
+        res = self.client.post(f"/suppliers/frameworks/g-cloud-12/sign-framework-agreement",
+                               data={"signerName": "Jane Doe",
+                                     "signerRole": "Director",
+                                     "signer_terms_and_conditions": "True"})
+        self.data_api_client.create_framework_agreement.return_value = {"agreement": {"id": 789}}
```
katstevens, line 5881: Should this be set before the client calls?

gidsg: Yes, looks like that was the issue 😅

app/main/forms/frameworks.py
```diff
@@ -76,3 +76,48 @@ def __init__(self, lot_name, *args, **kwargs):
...
+
+
+class SignFrameworkAgreementForm(FlaskForm):
+
+    # Intended use of camel case here to match expected API fields
+    signerName = DMStripWhitespaceStringField(
+        "Your full name",
+        validators=[
+            DataRequired(message="You must provide your full name."),
+            Length(max=255, message="You must provide a name under 256 characters."),
```
katstevens, line 107: I think the [guidance](https://design-system.service.gov.uk/components/error-message/) now is to avoid the word 'must' - instead use an instruction eg 'Enter a name under 256 characters'. (cc @NoraGDS)

NoraGDS: Yes - on the error message not using 'must' and not sure max length is applicable for this question

•Enter your full name
•Enter your first and last name

gidsg: Thanks, I've updated these error messages now. @NoraGDS max length is a technical constraint of our database.

app/main/views/frameworks.py
gidsg: Thanks for implementing this @katstevens. I updated the view to use it in 7f54158.

app/templates/frameworks/sign_framework_agreement.html
```diff
@@ -0,0 +1,57 @@
...
+
+{% block mainContent %}
+    {% include 'toolkit/forms/validation.html' %}
+    <span class="govuk-caption-xl">{{ framework.name }} {{ contract_title }}</span>
+    <h1 class="govuk-heading-xl">Supplier appointment terms</h1>
+    {% with company_details=company_details %}
+        {% include "frameworks/_framework_agreement_legal_text_{0}.html".format(framework_slug) %}
+        <p class="govuk-clearfix">
+            <a href="{{ framework_pdf_url }}">
+            <img style="float:left; border: 4px solid #ccc; margin-right: 10px;" alt="" src="{{ url_for('static', filename='images/pdf-thumbnails/{0}.png'.format(framework_slug)) }}">
```
gidsg, line 38: 👀  @domoscargin I took this directly from the prototype, so there may be a better way to do it?

app/templates/frameworks/_agreement_returned_legal.html
```diff
@@ -3,22 +3,24 @@
...
-        Your framework agreement signature page has been sent to
-        the Crown Commercial Service (<abbr title="Crown Commercial Service">CCS</abbr>).
-        They’ll check and countersign the agreement, or contact you if there’s a
-        problem. You don’t need to wait for <abbr title="Crown Commercial Service">CCS</abbr> to
-        countersign your agreement before you start selling services.
-      </p>
-    {% endif %}
-    </div>
+      {% if signed_agreement_document_name %}
+            <div class="dmspeak">
```
domoscargin, line 7: Don't think we need this div here. We're getting rid of `dmspeak` classes elsewhere in favour of standard govuk frontend spacing, so may as well remove it here if we're not relying on the visual difference.

app/templates/frameworks/_framework_agreement_legal_text_g-cloud-12.html
```diff
@@ -0,0 +1,61 @@
+<h2 class="govuk-heading-m">1.      The Appointment</h2>
+
+
+<p>
+    This section includes terms about the Supplier appointment.</p>
+
+<h3 class="govuk-heading-s">1.1       Appointment</h3>
```
domoscargin, line 7: 1.1 Appointment is an `<h3>` but 1.2, etc are `<p>`, which would indicate that 1.1 describes them all.

app/templates/frameworks/legal_authority_no.html
```diff
@@ -0,0 +1,33 @@
...
+      {
+        "text": "Sign agreement"
+      }
+    ]
+  }) }}
+{% endblock %}
+
+{% block mainContent %}
+    <h1 class="govuk-heading-xl">You cannot sign the Framework Agreement</h1>
+    <p class="govuk-body">Only someone with legal authority in your company, like a director or company secretary, can sign the Framework Agreement.</p>
```
domoscargin, line 31: I wonder if this is an appropriate use case for the [warning text component](https://design-system.service.gov.uk/components/warning-text/).

app/templates/frameworks/legal_authority_no.html
```diff
@@ -0,0 +1,33 @@
...
+    <a href="/suppliers">Return to Digital Marketplace account</a>
```
domoscargin, line 32: Add `class="govuk-link"`

app/templates/frameworks/sign_framework_agreement.html
```diff
@@ -0,0 +1,57 @@
...
+                Read {{ framework.name }} {{  contract_title }}
+            </a>
+            <br>
+            <span class="govuk-body-s">PDF, {{ framework_pdf_metadata.file_size }}, {{ framework_pdf_metadata.page_count }} pages</span>
+        </p>
+    {% endwith %}
+
+
+    <form method="POST" action="{{ url_for(".sign_framework_agreement", framework_slug=framework_slug) }}">
```
domoscargin, line 47: We should add novalidate to form tags.

app/templates/frameworks/_framework_agreement_legal_text_g-cloud-12.html
```diff
@@ -0,0 +1,61 @@
...
+
+<p>company number: {{  company_details['registration_number'] }}</p>
+
+
+
+<p>(the ‘Supplier’), as a Framework Agreement Supplier of G-Cloud Services who can be considered for Call-Off Contracts as outlined in the Contract Notice in the Official Journal of the European Union reference RM1557.12 (OJEU Contract Notice).</p>
+
+
+
+<p>1.2       Appointment is to: {{ completed_lots | join(', ')}}</p>
```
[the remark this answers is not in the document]
gidsg: @domoscargin I am not sure to be honest, this is what I have been given, but I'm not sure how intentional/significant the layout is. CC @NoraGDS 
Screenshot 2020-08-20 at 16 57 07

app/templates/frameworks/sign_framework_agreement.html
domoscargin: I think a float in this case is fine. That's how it's handled in [publishing components](https://components.publishing.service.gov.uk/component-guide/attachment). 

Give the image a class - something like `app-file-image` and transfer the styles to SCSS (not sure if one of the existing scss files is suitable or if a new one has to be created). Use `$govuk-border-colour` instead of `#ccc`.

Looking at the publishing component, it might be good to add an `<abbr>` around PDF.

app/templates/frameworks/legal_authority_no.html
domoscargin: Talked myself out of it - we're not really warning the user about an action they can take.

app/templates/frameworks/_framework_agreement_legal_text_g-cloud-12.html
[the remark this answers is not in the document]
NoraGDS: layout and numbering is set by CCS and agreed with its legal team - it's taken from the legal contract and we agreed to keep it consistent. 

although the points shouldn't be indented - just aligned left for all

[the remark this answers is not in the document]
NoraGDS: 1.2 should be a variable - whatever lot the supplier was successful in applying for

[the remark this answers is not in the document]
NoraGDS: layout and numbering are from CCS's legal contract - we agreed to keep it consistent with that

all lines should be left aligned, no indent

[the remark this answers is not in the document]
domoscargin: It's possible to achieve "1.1" type numberings with an `<ol>` and some CSS: https://stackoverflow.com/questions/4098195/can-ordered-list-produce-result-that-looks-like-1-1-1-2-1-3-instead-of-just-1

But I think in this case `<p>` tags under 1 - The Appointment are fine.

app/templates/frameworks/sign_framework_agreement_confirmation.html
```diff
@@ -0,0 +1,39 @@
...
+    {{ framework.name }} Signed agreement – Digital Marketplace
+{% endblock %}
+
+
+
+{% block mainContent %}
+    <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+      {{ govukPanel({
+        "titleText": "You’ve signed the {} {}".format(framework.name, contract_title),
```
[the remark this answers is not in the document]
gidsg: ...OK, I got some help from design system team on Slack and have resolved this. You can use `titleHtml`.

gidsg:
> Also one thing we'll need to update is the link on the _framework_actions.html template, to get to the new flow from the supplier dashboard

Thanks for the guidance on this @katstevens. I have updated this in 7452ee9.

app/templates/frameworks/sign_framework_agreement_confirmation.html
```diff
@@ -0,0 +1,39 @@
...
+    {{ framework.name }} Signed agreement – Digital Marketplace
+{% endblock %}
+
+
+
+{% block mainContent %}
+    <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+      {{ govukPanel({
+        "titleHtml": "You’ve signed the <span style=\"white-space: nowrap\">{}</span> {}".format(framework.name, contract_title),
```
katstevens, line 17: Looking fab 🏆

app/templates/frameworks/_framework_actions.html
```diff
@@ -14,8 +14,9 @@
   {% endif %}
 
   {% if supplier_is_on_framework and not supplier_framework.agreementReturned %}
+  {% set sign_agreement_route = url_for('.legal_authority', framework_slug=framework.slug) if is_e_signature_supported_framework else url_for('.framework_agreement', framework_slug=framework.slug) %}
```
katstevens, line 17: I've spotted two more includes that the old URL is used - https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/templates/suppliers/_frameworks_standstill.html#L32 and https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/templates/suppliers/_frameworks_live.html#L6

gidsg: good spot @katstevens 🕵️‍♀️  .
I've added the same logic to these 2 templates now in 4dbb001.
As the framework wasn't passed directly in from the view I couldn't figure out a way to implement this except directly in the templates. Let me know if you have strong feelings about this and if so any way you can think to abstract it out.

katstevens: Using the config will work OK. However the `frameworks` list being passed to these templates is in fact a list of 'enhanced' frameworks annotated with extra info. We can add in a `is_e_signature_supported_framework` key in the `suppliers.dashboard` view here: https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/main/views/suppliers.py#L83

gidsg: thanks for the tip, I have done this refactor in 9582ab7.

app/templates/frameworks/legal_authority.html
```diff
@@ -0,0 +1,38 @@
...
+      }
+    ]
+  }) }}
+{% endblock %}
+
+{% block mainContent %}
+    {% include 'toolkit/forms/validation.html' %}
+    <form method="POST" novalidate action="{{ url_for(".legal_authority", framework_slug=framework_slug) }}">
+        <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
+        {{ form.legal_authority }}
```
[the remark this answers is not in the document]
domoscargin: Think this needs an `h1`.

Some options off the top of my head:

* Use `govuk_options` from dmutils: https://github.com/alphagov/digitalmarketplace-utils/blob/79cbc5eac674cb438fc90c598b76395f50491a25/dmutils/forms/helpers.py#L70

This would possibly involve bumping dmutils and refactoring the view code, but we might then be able to separate the information.

* Figure out a way to exclude the label from `LegalAuthorityForm`

Can we create a `isPageHeading` flag which for the moment just ignores the question label?

* Repeat the question
Not ideal, but include the question as an `h1` tag as well as at the top of the form.

app/templates/frameworks/_framework_actions.html
katstevens: Great - looks good to me.

app/templates/frameworks/legal_authority.html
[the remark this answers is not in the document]
gidsg: @katstevens @domoscargin OK, I've refactored this in df2278b to use `govuk_options` and `govukRadios`, 
this also adds an `h1` by virtue of the `isPageHeading` option.

app/main/views/frameworks.py
```diff
@@ -1236,6 +1236,20 @@ def legal_authority(framework_slug):
     if not get_supplier_on_framework_from_info(supplier_framework):
         return render_error_page(status_code=400, error_message="You must be on the framework to sign agreement.")
     form = LegalAuthorityForm()
+    legal_authority_gov_uk_radios = {
```
domoscargin, line 1239: 👍

phantas (commented):
👍

app/main/helpers/suppliers.py
```diff
@@ -46,3 +46,23 @@ def supplier_company_details_are_complete(supplier_data):
              for f in
              contact_required_fields])
     )
+
+
+def get_company_details_from_supplier(supplier):
+    address = {"country": supplier.get("registrationCountry")}
```
katstevens, line 52: Are we actually using `country` in the `legal_text` template? I ask because in the DB this field is stored in the format `"country:GB"`, which needs transforming via a template filter before use (see [this example in the Admin](https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/master/app/templates/suppliers/_company_details_table.html#L12)).

If we're not using it, great - let's delete 😸

app/main/helpers/suppliers.py
```diff
@@ -46,3 +46,23 @@ def supplier_company_details_are_complete(supplier_data):
...
+            "locality": supplier["contactInformation"][0].get("city"),
+            "postcode": supplier["contactInformation"][0].get("postcode"),
+        })
+    return {
+        "registration_number": (
+            supplier.get("companiesHouseNumber")
+            or
+            supplier.get("otherCompanyRegistrationNumber")
+        ),
+        "registered_with": "companies_house" if "companiesHouseNumber" in supplier else "unknown",
```
katstevens, line 65: Are we using `registered_with` anywhere? Probably better to have an empty/falsy string rather than `unknown` on a legal doc if we do...

app/main/helpers/suppliers.py
gidsg: We are not using it currently. It's unclear from the prototype what format/resolution of address is needed. I can delete it or transform it as you say.

@NoraGDS do you know if country is required in the address displayed in the contract?

tests/app/main/test_frameworks.py
```diff
@@ -5758,3 +5757,209 @@ def test_false_values(self, falsey_value):
...
+            on_framework=True)
+        self.data_api_client.find_draft_services_by_framework.return_value = {
+            'meta': {'total': 1}
+        }
+        self.data_api_client.get_framework.return_value = self.framework(status='standstill',
+                                                                         slug='g-cloud-12',
+                                                                         framework_agreement_version="1")
+
+        self.login()
+        res = self.client.get("/suppliers/frameworks/g-cloud-12/sign-framework-agreement")
```
katstevens, line 5915: Do we need the GET request here?

tests/app/main/test_frameworks.py
```diff
@@ -5758,3 +5757,209 @@ def test_false_values(self, falsey_value):
...
+        assert res.status_code == 200
+        res = self.client.post("/suppliers/frameworks/g-cloud-12/sign-framework-agreement",
+                               data={"signerName": "Jane Doe",
+                                     "signerRole": "Director",
+                                     "signer_terms_and_conditions": "True"})
+
+        self.data_api_client.create_framework_agreement.assert_called_once_with(1234, 'g-cloud-12', 'email@email.com')
+
+        self.data_api_client.update_framework_agreement.assert_called_with(789, {
```
katstevens, line 5924: Should this be `assert_called_once_with`, like the others?

tests/app/main/test_frameworks.py
```diff
@@ -5758,3 +5757,209 @@ def test_false_values(self, falsey_value):
...
+
+    @pytest.mark.parametrize(
+        ('framework_slug', 'on_framework', 'status_code'),
+        (
+            ('g-cloud-11', True, 404),
+            ('g-cloud-12', True, 200),
+            ('g-cloud-12', False, 400),
+        )
+    )
+    def test_only_works_for_supported_frameworks(self, framework_slug, on_framework, status_code):
```
katstevens, line 5871: Could we also test the page works for both `live` and `standstill` framework statuses? Probably better off in a separate test, something like:

```
@pytest.mark.parametrize(
        ('framework_status', 'status_code'),
        (
            ('coming', 404),
            ('open', 404),
            ('pending', 404),
            ('standstill', 200),
            ('live', 200),
            ('expired', 404),
        )
    )
    def test_only_works_for_live_and_standstill_frameworks(self, framework_slug, status_code):
        ...
``` 

Could do with having a test like this for the `/start-framework-agreement-signing` view too?

katstevens (commented):
Have left a few comments, most of them are tiny things. 

I've made a couple of suggestions about extra tests, see what you think.

app/main/helpers/suppliers.py
NoraGDS: Yes, it probably is. EU and US-based companies are often suppliers.

app/main/helpers/suppliers.py
gidsg: good spot, removed in ada1077

app/main/views/frameworks.py
```diff
@@ -50,16 +52,22 @@
...
 )
-from ..helpers.suppliers import supplier_company_details_are_complete
+from ..helpers.suppliers import supplier_company_details_are_complete, get_company_details_from_supplier
 from ..helpers.validation import get_validator
-from ..forms.frameworks import SignerDetailsForm, ContractReviewForm, AcceptAgreementVariationForm, ReuseDeclarationForm
+from ..forms.frameworks import (SignerDetailsForm,
+                                ContractReviewForm,
+                                AcceptAgreementVariationForm,
+                                ReuseDeclarationForm,
+                                LegalAuthorityForm, SignFrameworkAgreementForm)
```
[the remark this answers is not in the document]
gidsg: fixed in e0639ed1a4620b1be799f04017f7e3249dbf1ba5

app/templates/frameworks/legal_authority_no.html
```diff
@@ -0,0 +1,33 @@
...
+      {
+        "text": "Your account",
+        "href": url_for('.dashboard')
+      },
+      {
+        "text": framework.name,
+        "href": url_for(".framework_dashboard", framework_slug=framework.slug)
+      },
+      {
+        "text": "Sign agreement"
```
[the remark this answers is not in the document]
gidsg: good spot. Updated in 85144b3643a3b39358f5f61981c34f0314d000ac

app/templates/frameworks/sign_framework_agreement.html
```diff
@@ -0,0 +1,60 @@
...
+    <h1 class="govuk-heading-xl">Supplier appointment terms</h1>
+    {% with company_details=company_details %}
+        <p class="govuk-!-font-weight-bold">
+            If any of the information about your company is incorrect contact <a href="mailto:cloud_digital@crowncommercial.gov.uk">cloud_digital@crowncommercial.gov.uk</a> before signing.
+        </p>
+        {% include "frameworks/_framework_agreement_legal_text_{0}.html".format(framework_slug) %}
+        <p class="govuk-clearfix">
+            <a href="{{ framework_pdf_url }}">
+            <img class="framework-agreement-thumbnail" alt="" src="{{ url_for('static', filename='images/pdf-thumbnails/{0}.png'.format(framework_slug)) }}">
+                Read {{ framework.name }} {{  contract_title }}
```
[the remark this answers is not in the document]
gidsg: This text is taken directly from  the prototype: 
Screenshot 2020-08-25 at 17 01 36
CC @NoraGDS in case it is a typo?

tests/app/main/test_frameworks.py
gidsg: I am not sure we enforce this currently but in adding a test we can confirm and enforce if necessary.

tests/app/main/test_frameworks.py
```diff
@@ -5758,3 +5757,209 @@ def test_false_values(self, falsey_value):
...
+        self.login()
+        self.data_api_client.get_framework.return_value = self.framework(status='standstill', slug=framework_slug)
+        self.data_api_client.find_draft_services_iter.return_value = [
+            {'serviceName': 'A service', 'status': 'submitted', 'lotSlug': 'iaas'}
```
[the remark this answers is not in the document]
gidsg: good spot. Removed in a9702df.

tests/app/main/test_frameworks.py
```diff
@@ -5758,3 +5757,209 @@ def test_false_values(self, falsey_value):
...
+        self.data_api_client.get_framework.return_value = self.framework(status='standstill',
+                                                                         slug=framework_slug,
+                                                                         framework_agreement_version="1")
+        self.data_api_client.find_draft_services_iter.return_value = [
+            {'serviceName': 'A service', 'status': 'submitted', 'lotSlug': 'iaas'}
+        ]
+        self.data_api_client.get_supplier_framework_info.return_value = self.supplier_framework(
+            on_framework=True)
+
+        res = self.client.get(f"/suppliers/frameworks/{framework_slug}/start-framework-agreement-signing")
```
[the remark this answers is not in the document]
gidsg: good spot. Removed in 294392eda1d86b6609af6dc0b25a7e0b149ba355.

tests/app/main/test_frameworks.py
gidsg: see https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1239#discussion_r476572666

tests/app/main/test_frameworks.py
gidsg: fixed in 83553c9032951cb599a437ef4a4bd5f58e9df04c

tests/app/main/test_frameworks.py
gidsg: fixed in 7c1f9f2

katstevens (approved):
Absolutely epic stuff. 🏆 

I think there's just the one question about the potential missing 'the' - if necessary we can fix that in a separate PR. 

If there's a sane way of squashing commits before merging then go for it 😹

app/templates/frameworks/sign_framework_agreement.html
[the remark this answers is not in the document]
NoraGDS: Good spot. It should have 'the'

app/main/helpers/suppliers.py
katstevens: I've pushed a commit to fix.

app/templates/frameworks/sign_framework_agreement.html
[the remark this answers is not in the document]
katstevens: Pushed a commit to fix 👍

app/main/helpers/suppliers.py
katstevens: We can leave this for now I think.

katstevens:
I've had a go at squashing some of the commits but have ran out of steam 😹 Will merge this now so we can try out on Preview.

PR #1240, opened 2020-08-19
Improve framework dashboard load time for local envs

katstevens:
On the supplier's framework dashboard, in order to show the 'last updated' dates on the framework legal documents, we traverse the entire framework directory in the S3 comms bucket. 

For local developers, this means *everything* in the dev-uploads bucket is parsed, which could be hundreds of documents (from functional tests). The longer the framework runs, the more documents will be in the bucket, and the slower it will get.

This PR narrows down the scope to just the communications folder (legal docs and clarification updates), which should be a fixed number of files throughout the lifetime of the framework.

This change won't affect Preview/Staging/Production as submission documents go to a separate bucket. There are potential improvements to be made here too, but they're more involved.

Local testing improved page load times from ~30s to ~2s.

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -163,7 +163,8 @@ def framework_dashboard(framework_slug):
             supplier_framework_info['agreementPath']
         )
 
-    key_list = s3.S3(current_app.config['DM_COMMUNICATIONS_BUCKET']).list(framework_slug, load_timestamps=True)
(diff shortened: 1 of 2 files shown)


gidsg (approved):
nice spot ✨  👍

lfdebrux (approved):
🙌 

I've always just commented out these lines locally 😅

PR #1241, opened 2020-08-20
add secrets baseline

gidsg:
Adds secrets baseline as per
https://github.com/alphagov/gds-pre-commit/
https://trello.com/c/jbk1Il5K/1851-install-gds-pre-commit-and-commit-secretsbaseline-to-repos

## Notes on setup.

 I was initially getting some weird behaviour where `detect-secrets` would not work if virtualenv was enabled. Or the global `detect-secrets` did not produce the file in the same format expected by the pre-commit hook. To workaround this I ran:
`pip3 install detect-secrets==0.13.1` as suggested by @heathd. 
this installs the same version used by the pre-commit hook into pyenv environment.
When generating the baseline I used
`detect-secrets scan --exclude-files package-lock.json > .secrets.baseline`
this avoids many false positives associated with package integrity hashes.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1,0 +1,120 @@
+{
+  "exclude": {
+    "files": "package-lock.json",
+    "lines": null
+  },
+  "generated_at": "2020-08-20T10:45:38Z",
+  "plugins_used": [
+    {
+      "name": "AWSKeyDetector"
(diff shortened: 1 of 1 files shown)


katstevens (approved):
Let's give it a try.

PR #1242, opened 2020-09-02
Prevent Chrome autofilling company name

katstevens:
https://trello.com/c/XGSVuDnG/

We need to explicitly disable autocomplete for this field - Chrome appears to be inferring that this is a Title field because the following field is Contact Name.

[Chrome does not support `autocomplete="off"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) so I've gone for a dummy value instead.

### DIFF
--- a/app/templates/suppliers/create_company_details.html
+++ b/app/templates/suppliers/create_company_details.html
@@ -47,6 +47,7 @@
             question = "Company name",
             name = "company_name",
             value = form.company_name.data,
+            autocomplete = "new-supplier-organisation",
             error = errors.get("company_name", {}).get("message", None),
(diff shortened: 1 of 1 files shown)


domoscargin:
In Firefox, I'm still getting a suggestion for the input (though I suppose I have to manually select it)

Screenshot 2020-09-02 at 12 04 38

domoscargin:
Perhaps if we're doing a random string, we can make it semantic - something like "new-supplier-organisation" or similar.

domoscargin (approved):

katstevens:
> Perhaps if we're doing a random string, we can make it semantic - something like "new-supplier-organisation" or similar.

Yep that's a good shout - will amend

PR #1243, opened 2020-09-03
Add tests for legal text

gidsg:
Follow-up after #1239

adds test for legal text.

Thought it better to add these as unit tests than adding to https://github.com/alphagov/digitalmarketplace-functional-tests/pull/788


NoraGDS:
agree.

gidsg:
@NoraGDS OK as per https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1243#issuecomment-686588329 I have reverted adding the country.

katstevens (approved):

PR #1245, opened 2020-09-07
add inline error message

gidsg:
Add missing inline error to legal authority form 
Screenshot 2020-09-07 at 14 31 32

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -1245,6 +1245,7 @@ def legal_authority(framework_slug):
         if response == 'yes':
             return redirect(url_for('.sign_framework_agreement', framework_slug=framework_slug))
     errors = get_errors_from_wtform(form)
+    field_name = "legal_authority"
     legal_authority_gov_uk_radios = {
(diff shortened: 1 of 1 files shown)


app/main/views/frameworks.py
```diff
@@ -1249,16 +1257,9 @@ def legal_authority(framework_slug):
         "name": "legal_authority",
         "hint": {"text": LegalAuthorityForm.HINT},
         "classes": "govuk-radios--inline",
-        "items": govuk_options(LegalAuthorityForm.OPTIONS)
+        "items": govuk_options(LegalAuthorityForm.OPTIONS),
+        "errorMessage": list(errors.values())[0]['errorMessage'] if errors else None
```
katstevens, line 1261: Is this the only we way can extract the error message from the `errors` dict? I know we only have 1 field in this form but It'd be good if we could retrieve by field name...

gidsg: good point, fixed in 4a29640.

katstevens (approved):

PR #1247, opened 2020-09-11
bump frameworks version

gidsg:
bump to pick up new URLs﻿


katstevens:
I think this will need the `package-lock.json` updating as well (via `npm install`)

katstevens (approved):

gidsg:
Thanks @katstevens something weird happened where `npm install` was not updating the lock file, but it did work by running
`npm install github:alphagov/digitalmarketplace-frameworks#v17.12.0`

PR #1257, opened 2020-09-25
Show services and documents links on dashboard

domoscargin:
https://trello.com/c/ZUgsBovZ/361-2-prep-and-test-buying-journey-for-g12-on-preview

When a framework is live, we want users to be able to view their services and documents, even if they have an agreement signature outstanding.

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -31,7 +31,7 @@ from ..helpers.frameworks import (
     get_frameworks_by_status,
     get_frameworks_closed_and_open_for_applications,
     get_unconfirmed_open_supplier_frameworks,
-    is_e_signature_supported_framework
+    is_e_signature_supported_framework,
 )
 from ..helpers.suppliers import (
     COUNTRY_TUPLE,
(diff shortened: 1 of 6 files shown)


lfdebrux (approved):
Looks good to me.

lfdebrux (changes requested):
Could we have it that they have the sign framework agreement link even if they're suspended?

domoscargin:
So this now fails [`test_opportunities_dashboard_link_fails_with_incomplete_data`](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/8e82a0f300d5b80c881aeb9bdd3389f66a06e544/tests/app/main/test_suppliers.py#L1120-L1134), since this logic doesn't use the `onFramework` property. The "view services/documents" links all appear on this test.

I'm not entirely sure I've copied the logic correctly, and I'm not entirely sure what should be showing in this test case now - the copied logic seems to bypass `onFramework` and `services_count`, but are there specific cases that should happen?

lfdebrux:
I've pushed some changes to try and fix the tests.

lfdebrux (approved):

PR #1263, opened 2020-09-30
Validate declaration questions for DOS 5

bjgill:
Trello: https://trello.com/c/Oqg9AUup/328-3-update-conditional-logic-for-declaration-question-q53-for-dos5

For DOS 5, we've renamed `primaryContactEmail` to `contactEmail`. Add a new validator for DOS 5 (+?) to handle this renamed field.

Also add validation for the new percent question - it's a number between 0 and 100. We need a new validator because the existing one used for the DUNS number validates numbers with a fixed length, which is not right for a percentage.

### DIFF
--- a/app/main/helpers/validation.py
+++ b/app/main/helpers/validation.py
@@ -300,6 +300,25 @@ class DOS5Validator(SharedValidator):
     email_validation_fields = {"contactEmailContractNotice", "contactEmail"}
     percentage_fields = ["subcontractingInvoicesPaid"]
 
+    optional_fields = SharedValidator.optional_fields.union({
+        "subcontracting30DayPayments",
+        "subcontractingInvoicesPaid"}
+    )
+
+    def get_required_fields(self):
+        req_fields = super(DOS5Validator, self).get_required_fields()
+
+        # as per subcontracting configuration on digitalmarketplace-frameworks
+        if self.answers.get("subcontracting") in [
+            "as a prime contractor, using third parties (subcontractors) to provide some services",
+            "as part of a consortium or special purpose vehicle, using third parties (subcontractors) to provide some "
+            "services"
+        ]:
+            req_fields.add("subcontracting30DayPayments")
+            req_fields.add("subcontractingInvoicesPaid")
+
+        return req_fields
+
     def formatting_errors(self, answers):
         error_map = super(DOS5Validator, self).formatting_errors(answers)
 
--- a/app/main/helpers/validation.py
+++ b/app/main/helpers/validation.py
@@ -282,33 +282,31 @@ class G12Validator(SharedValidator):
         return errors_map
 
 
+def is_valid_percentage(value):
+    # Design System guidance is that we should allow users to provide answers with or without units.
+    if isinstance(value, str):
+        value = value.rstrip('%')
+
+    try:
+        number = float(value)
+    except ValueError:
+        return False
+    else:
+        return 0 <= number <= 100
+
+
 class DOS5Validator(SharedValidator):
     """Following an accessibility review, a number of questions and answers were changed for DOS 5"""
     email_validation_fields = {"contactEmailContractNotice", "contactEmail"}
     percentage_fields = ["subcontractingInvoicesPaid"]
 
-    def is_valid_percentage(self, field):
-        value = self.answers.get(field)
-
-        # Design System guidance is that we should allow users to provide answers with or without units.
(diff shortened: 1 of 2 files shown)


lfdebrux:
### My review checklist

I've:

- [ ] looked at the code
- [ ] tested locally

app/main/helpers/validation.py
```diff
@@ -282,6 +282,27 @@ def errors(self):
...
+    email_validation_fields = {"contactEmailContractNotice", "contactEmail"}
+
+    number_fields = [("subcontracting30DayPayments", 0, 100)]
+
+    def formatting_errors(self, answers):
+        error_map = super(DOS5Validator, self).formatting_errors(answers)
+
+        for field, minimum, maximum in self.number_fields:
+            try:
+                number = float(self.answers.get(field))
```
lfdebrux, line 296: We want users to be able to enter the percentage sign [[1]](https://design-system.service.gov.uk/components/text-input/#prefixes-and-suffixes).

app/main/helpers/validation.py
```diff
@@ -282,6 +282,27 @@ def errors(self):
         return errors_map
 
 
+class DOS5Validator(SharedValidator):
+    """Following an accessibility review, a number of questions and answers were changed for DOS 5"""
+    email_validation_fields = {"contactEmailContractNotice", "contactEmail"}
+
+    number_fields = [("subcontracting30DayPayments", 0, 100)]
+
```
lfdebrux, line 290: Are you planning to add the validation for the subcontracting logic in alphagov/digitalmarketplace-frameworks#633 in a separate PR?

app/main/helpers/validation.py
```diff
@@ -282,6 +282,27 @@ def errors(self):
         return errors_map
 
 
+class DOS5Validator(SharedValidator):
+    """Following an accessibility review, a number of questions and answers were changed for DOS 5"""
+    email_validation_fields = {"contactEmailContractNotice", "contactEmail"}
+
+    number_fields = [("subcontracting30DayPayments", 0, 100)]
```
lfdebrux, line 289:
```suggestion
    number_fields = [("subcontractingInvoicesPaid", 0, 100)]
```

app/main/helpers/validation.py
lfdebrux, line 296: I got a TypeError if field is not in answers

```
TypeError: float() argument must be a string or a number, not 'NoneType'
```

app/main/helpers/validation.py
```diff
@@ -282,6 +282,37 @@ def errors(self):
...
+        except ValueError:
+            return False
+        else:
+            return 0 <= number <= 100
+
+    def formatting_errors(self, answers):
+        error_map = super(DOS5Validator, self).formatting_errors(answers)
+
+        error_map.update({
+            field: 'invalid_format'
```
lfdebrux, line 308: I think this needs to be 'not_a_number'? [[1]](https://github.com/alphagov/digitalmarketplace-frameworks/blob/master/frameworks/digital-outcomes-and-specialists-5/questions/declaration/subcontractingInvoicesPaid.yml).

Currently I'm getting the error message 'You must enter a valid email address.'

You must enter a valid email address.

app/main/helpers/validation.py
```diff
@@ -282,6 +282,37 @@ def errors(self):
...
+    """Following an accessibility review, a number of questions and answers were changed for DOS 5"""
+    email_validation_fields = {"contactEmailContractNotice", "contactEmail"}
+    percentage_fields = ["subcontractingInvoicesPaid"]
+
+    def is_valid_percentage(self, field):
+        value = self.answers.get(field)
+
+        # Design System guidance is that we should allow users to provide answers with or without units.
+        if isinstance(value, str):
+            value = value.rstrip('%')
```
lfdebrux, line 295: Neat trick! We might need to save this back to `answers`, as otherwise the declaration gets updated with the percentage sign, which is probably not what we want.

bjgill: Hmm. This makes me quite uneasy. At the moment, the validators do not change any of the parameters they're passed. Can we easily do this in a thread-safe way?

Is it right to modify the user input, or should we be making this change where we display it?

We might want to discuss with you and Brett. As such, we're not going to fix this in this PR.

phantas (approved):
with the caveat that I co-authored and we addressed other feedback

PR #1265, opened 2020-10-06
use generic template for started framework application

gidsg:
https://trello.com/c/q5k7renU/435-combine-two-you-have-started-an-application-emails-05
Combine these two emails into one generic one


bjgill (approved):
I've looked at both code and template, and both look good.

PR #1273, opened 2020-10-14
add intellij flask run configuration

gidsg:
https://trello.com/c/fwLv6sMf/61-work-out-and-document-how-to-attach-dmp-apps-to-an-ide-debugger

This adds a run configuration using the [Share configurations](https://www.jetbrains.com/help/idea/run-debug-configuration.html#share-configurations) feature of IntelliJ.

## How to use
* Before using this ensure the `dmrunner` process has been killed: from `dmrunner` do `k supplier-frontend`
* Open the project in IntelliJ you should see a Run configuration called _Flask (application.py)_ which you should be able to Run or Debug.

## Known issues
* Credentials are not injected so any route relying on credentials currently fails

## Questions
* Does this work in Pycharm?
* Does this work in the free versions of IntelliJ or Pycharm? (I think it may not as it uses a premium plugin, but not sure?)

### DIFF
--- a/.idea/runConfigurations/Flask__application_py_.xml
+++ b/.idea/runConfigurations/Flask__application_py_.xml
@@ -1,0 +1,18 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="Flask (application.py)" type="Python.FlaskServer" nameIsGenerated="true">
+    <option name="flaskDebug" value="true" />
+    <module name="digitalmarketplace-supplier-frontend" />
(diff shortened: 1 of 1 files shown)


bjgill:
Alas, it doesn't seem to work for me on community PyCharm:

> Run Configuration Error: Broken configuration due to unavailable plugin or invalid configuration data.

gidsg:
@bjgill Argh, I think it may only work on Pycharm Professional as per https://www.jetbrains.com/help/pycharm/creating-web-application-with-flask.html can you confirm if you have the professional or free edition?

PR #1275, opened 2020-10-16
Tidy up validators

bjgill:
Use set literals and add type hints where I can.

### DIFF
--- a/app/main/helpers/validation.py
+++ b/app/main/helpers/validation.py
@@ -143,12 +143,12 @@ class G7Validator(DeclarationValidator):
     """
     Validator for G-Cloud 7.
     """
-    optional_fields = set([
+    optional_fields = {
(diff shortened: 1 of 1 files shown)


app/main/helpers/validation.py
```diff
@@ -282,7 +283,7 @@ def errors(self):
         return errors_map
 
 
-def is_valid_percentage(value):
+def is_valid_percentage(value: Any) -> bool:
```
gidsg, line 286: Would `(value: Union[str, int]) -> bool` work?

bjgill: That would work for the function as currently used, but I'm not sure `Any` is wrong. `is_valid_percentage` will happily accept and correctly process a value of any type.

gidsg (approved):
👍 🦖

PR #1276, opened 2020-10-28
Add DM_S3_ENDPOINT_URL to Development config

lfdebrux:
For development instances we want developers to be able to use a tool like [localstack] instead of real AWS endpoints. The Flask configuration variable DM_S3_ENDPOINT_URL can be used (if set) by dmutils.s3.S3 to connect to a localstack instance (see alphagov/digitalmarketplace-utils#580).

This commit adds DM_S3_ENDPOINT_URL to the Development configuration only, and makes it dependant on an environment variable, DM_S3_ENDPOINT_PORT (so we can set the port only, or the port and hostname, using envvars).

[localstack]: https://github.com/localstack/localstack

### DIFF
--- a/config.py
+++ b/config.py
@@ -137,11 +137,20 @@ class Development(Config):
     DM_DATA_API_AUTH_TOKEN = "myToken"
     DM_API_AUTH_TOKEN = "myToken"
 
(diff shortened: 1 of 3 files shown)


bjgill (approved):
:tada:

PR #1283, opened 2020-11-27
Un-hard-code previous DOS framework

bjgill:
Fulfil [Kat's vow](https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1123). We need to show the previous DOS framework so that suppliers can see their applications on the previous DOS framework. Previously, we did this by hard coding the last DOS framework in the code.

Now do it by choosing the expired DOS framework with the latest expiry date. This will save us from needing to remember to update this manually when we make DOS 5 live.


alex9smith (approved):
🔢  👍

PR #1284, opened 2020-11-27
Use mypy to test typing

bjgill:
This PR adds [`mypy`](http://mypy-lang.org/) static type checking to the checks performed on `make test` and on Travis CI (but not on commit).

I've also fixed/ignored the issues mypy picked up as appropriate.

I copied the approach from [dmrunner](https://github.com/alphagov/digitalmarketplace-runner/pull/51), which we did first.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,7 @@ htmlcov/
 .cache
 nosetests.xml
 coverage.xml
+.mypy_cache
 
 # Translations
 *.mo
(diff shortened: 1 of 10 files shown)


lfdebrux (approved):
Nice!

bjgill:
Oh dear. We've hit https://github.com/python/mypy/issues/9656.

lfdebrux:
> Oh dear. We've hit [python/mypy#9656](https://github.com/python/mypy/issues/9656).

Maybe we should just pin to the older version of mypy and hope they find a way around this issue?

bjgill:
Agreed. There is a workaround (assign the lambda to a variable), but it's a bit silly.

PR #1288, opened 2020-12-01
Distinguish G12 recovery suppliers

bjgill:
Trello: https://trello.com/c/KikQx3R2/593-as-a-g12-recovery-supplier-i-can-see-something-other-suppliers-cannot

We had an outage during G12 closing. Some suppliers will be involved in a recovery process to account for the outage causing them to fail to submit their services in time.

We have previously provided a list of the affected supplier IDs via the environment - see https://github.com/alphagov/digitalmarketplace-credentials/pull/359 and https://github.com/alphagov/digitalmarketplace-aws/pull/797. This PR proves that this all works by displaying a placeholder to the affected suppliers.

Currently, the only supplier so configured is our test supplier, so this will not be shown to any real users and is therefore safe for production. The two tests check this.

image

### DIFF
--- a/app/main/helpers/suppliers.py
+++ b/app/main/helpers/suppliers.py
@@ -1,5 +1,8 @@
 import os
 import json
+from typing import Union
+
+from flask import current_app
 
 
 def load_countries():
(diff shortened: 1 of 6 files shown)


agalamatis:
Test cases for the attribute DM_G12_RECOVERY_SUPPLIER_IDS not existing, being a string, an empty iterable or containing non-int objects? I might be getting paranoid though, but it can be anything and getting a server error in a supplier's landing page by mistyping an ID, won't be fun to deal with.

bjgill:
Good idea. Whilst it should always be present, and of the right type (`Optional[str]`), it would be easy for us to mess up the contents of the string. I've added logging and a test.

bjgill:
Once this is merged, I'll copy the function over to the API app and use it there. Given we're only using it temporarily and in two apps, I think this is OK.

agalamatis (approved):
Looks good, on to the API :)

PR #1291, opened 2020-12-04
fix cache key by using correct variable

gidsg:
Spotted this issue when reviewing https://github.com/alphagov/digitalmarketplace-scripts/pull/588 the second variable wasn't being resolved which made the cache key non-unique across matrix runs.


alex9smith (approved):
👍

PR #1293, opened 2020-12-08
Replace flash messages with alerts

lfdebrux:
Ticket: https://trello.com/c/Vp3FZhKc/289-replace-flash-messages-with-digital-marketplace-govuk-frontend-alert-component-in-supplier-frontend

We want to use the dmAlerts macro instead of the old flash-messages 
template, as dmAlert is based on GOV.UK Design System styles and guidance.


domoscargin (approved):
LGTM

Just running functional tests on my machine with this and the brief-responses changes.

PR #1294, opened 2020-12-11
e-signature signing page for DOS 5

gidsg:
https://trello.com/c/75hbcRW8/607-2-implement-dos-5-e-signature

Adds content for e-signature signing page for DOS 5. This mostly involved adding some fields and extracting out a load of variables where there is variance between what is displayed on g-cloud-12 and DOS5.

Content has been signed off by @NoraGDS & CCS

localhost_suppliers_frameworks_digital-outcomes-and-specialists-5_sign-framework-agreement(iPad Pro) (5)


app/main/forms/frameworks.py
```diff
@@ -100,8 +100,14 @@ class LegalAuthorityForm(FlaskForm):
...
 
 class SignFrameworkAgreementForm(FlaskForm):
+    def __init__(self, contract_title, **kw):
+        super(SignFrameworkAgreementForm, self).__init__(**kw)
+        self.signer_terms_and_conditions.label.text = f"I accept the terms and conditions of the {contract_title}"
+        self.signer_terms_and_conditions.validators[0].message = f"Accept the terms and conditions of the" \
+                                                                 f" {contract_title}."
 
     # Intended use of camel case here to match expected API fields
+
```
bjgill, line 110: Nit: why this white space?

app/main/views/frameworks.py
```diff
@@ -1353,27 +1353,38 @@ def sign_framework_agreement(framework_slug):
...
+    include_govuk_link = framework_specific_labels.get(framework_slug).get('include_govuk_link')
+    framework_govuk_url = framework.get('frameworkAgreementDetails').get('frameworkURL')
+
+    form = SignFrameworkAgreementForm(contract_title)
 
     # TODO: can we derive this metadata programmatically?
     framework_pdf_metadata = {
-        'g-cloud-12': {'file_size': '487KB', 'page_count': 62}
+        'g-cloud-12': {'file_size': '487KB', 'page_count': 62},
+        'digital-outcomes-and-specialists-5': {'file_size': '97KB', 'page_count': 8}
```
bjgill, line 1376: Do you know if there's a story tracking this?

app/main/views/frameworks.py
```diff
@@ -1353,27 +1353,38 @@ def sign_framework_agreement(framework_slug):
...
     framework_pdf_url = content_loader.get_message(framework_slug, 'urls').get('framework_agreement_pdf_url')
-    contract_titles = {
-        'g-cloud-12': 'Framework Agreement'
+    framework_specific_labels = {
+        'g-cloud-12': {'title': 'Sign agreement',
+                       'contract_title': 'Framework Agreement',
+                       'include_govuk_link': False},
+        'digital-outcomes-and-specialists-5': {'title': 'Sign contract',
+                                               'contract_title': 'Framework Award Form',
+                                               'include_govuk_link': True}
```
bjgill, line 1364: Have we documented that this will need to be updated for G13/DOS6?

app/templates/frameworks/_framework_agreement_legal_text_digital-outcomes-and-specialists-5.html
```diff
@@ -0,0 +1,18 @@
...
+     to supply the Deliverables in {{ completed_lots | join(', ') }}.
+     You cannot deliver in any other Lot under this Contract. Any references made to other Lots in this Contract do not apply. This opportunity is advertised in the Contract Notice in the Official Journal of the European Union reference 2020/S 197-478606 (OJEU Contract Notice).
+</li>
+<li><strong>Deliverables</strong>: See Framework Schedule 1 (Specification) for further details.</li>
+<li><strong>Framework Start Date</strong>: 20 January 2021</li>
+<li><strong>Framework Expiry Date</strong>: 23:59 on 19 January 2022</li>
+<li><strong>Framework Initial Period</strong>: From Framework Start Date to Framework Expiry Date, inclusive.</li>
+<li><strong>Framework Optional Extension Period</strong>: Up to twelve (12) months, by CCS giving written notice to Suppliers.</li>
+</ol>
+<p>18. <strong>Supplier Authorised Representative</strong>: {{ declaration.get('contactNameContractNotice') }}, {{ declaration.get('contactEmailContractNotice') }}</p>
```
bjgill, line 17: Why is this number 18 and not in the `ol`?

app/templates/frameworks/_framework_agreement_legal_text_digital-outcomes-and-specialists-5.html
```diff
@@ -0,0 +1,18 @@
...
+<p>The finalised Framework Award Form includes an electronically signed Framework Award Form. It is electronically signed by the Supplier when they made the legal declaration confirming their agreement to the Digital Outcomes and Specialists 5 Framework Award Form, and countersigned by the Crown Commercial Service.</p>
```
bjgill, line 18: It looks as if we expect this content to change from framework to framework. Is supplier-frontend the right repo for this content?

And if so, do we need to update https://alphagov.github.io/digitalmarketplace-manual/content-and-frameworks/adding-frameworks.html?

app/main/views/frameworks.py
```diff
@@ -1353,27 +1353,38 @@ def sign_framework_agreement(framework_slug):
 
     supplier = data_api_client.get_supplier(current_user.supplier_id)["suppliers"]
     company_details = get_company_details_from_supplier(supplier)
-    form = SignFrameworkAgreementForm()
+    declaration = data_api_client.get_supplier_declaration(current_user.supplier_id, framework_slug).get('declaration')
```
bjgill, line 1356: A general point about `sign_framework_agreement` - it's now ~100 lines long. Would it be worth breaking bits out into functions to make it easier to understand what's going on?

bjgill (commented):
Looks good overall. A few nits about style and so forth we might want to think about.

My main thought is whether supplier-frontend should be where we keep framework-specific content long-term. I understand that time is tight, so I think it would be OK to leave content where it is in this PR and then move it to the right place afterwards.

app/templates/frameworks/_framework_agreement_legal_text_digital-outcomes-and-specialists-5.html
gidsg: This is a legal thing, to match the PDF.

app/main/views/frameworks.py
gidsg: @bjgill Do you mean a story for the TODO or something else?

bjgill: Yes - the TODO

gidsg:
> My main thought is whether supplier-frontend should be where we keep framework-specific content long-term.

@bjgill this is a good point. There were a bunch of reactive changes here, It is hard to know after 2 frameworks to what extent this content will vary across frameworks and framework families. I think by G-Cloud 13 we will be a better position to understand what is custom and what is standardised and thus what is a good candidate to be moved somework else. e.g. content loader may be a candidate for some variables.

bjgill:
> > My main thought is whether supplier-frontend should be where we keep framework-specific content long-term.
> 
> @bjgill this is a good point. There were a bunch of reactive changes here, It is hard to know after 2 frameworks to what extent this content will vary across frameworks and framework families. I think by G-Cloud 13 we will be a better position to understand what is custom and what is standardised and thus what is a good candidate to be moved somework else. e.g. content loader may be a candidate for some variables.

Good idea. Agree that this would make sense as a G13 story.

app/main/views/frameworks.py
gidsg: I am not sure in principal if it is a good idea to get this at runtime as this would involve an external call out to gov.uk but it may make more sense to move it out to the content loader or something. I can make a card to think about this.

app/main/views/frameworks.py
gidsg: As discussed elsewhere it is likely this process will change again in future, but it may make sense to add some pointers to the manual about updating e-signature content.

app/main/views/frameworks.py
```diff
@@ -1353,27 +1353,31 @@ def sign_framework_agreement(framework_slug):
...
+    form = SignFrameworkAgreementForm(contract_title)
 
     # TODO: can we derive this metadata programmatically?
     framework_pdf_metadata = {
-        'g-cloud-12': {'file_size': '487KB', 'page_count': 62}
+        'g-cloud-12': {'file_size': '487KB', 'page_count': 62},
+        'digital-outcomes-and-specialists-5': {'file_size': '97KB', 'page_count': 8}
     }
+
     lots = framework['lots']
```
bjgill, line 1379: It looks as if we use this only once. Does it need to be a variable?

bjgill (approved):
One minor suggestion.

You're going to create some stories and update some docs. The code looks good to go :tada:

app/main/views/frameworks.py
gidsg: added in 0fbf3f3

app/main/views/frameworks.py
gidsg: updated in f61397b

PR #1295, opened 2020-12-11
G12 recovery ticket 3

lfdebrux:
Ticket: https://trello.com/c/ex2Bkn90

### DIFF
--- a/tests/app/main/test_services.py
+++ b/tests/app/main/test_services.py
@@ -11,7 +11,13 @@ import mock
 import pytest
 
 from dmapiclient import HTTPError
-from dmtestutils.api_model_stubs import FrameworkStub, LotStub, SupplierStub, ServiceStub
+from dmtestutils.api_model_stubs import (
+    DraftServiceStub,
+    FrameworkStub,
+    LotStub,
+    SupplierStub,
+    ServiceStub,
+)
 from dmtestutils.fixtures import valid_pdf_bytes, valid_odt_bytes
 
 from app.main.helpers.services import parse_document_upload_time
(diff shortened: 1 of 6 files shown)


tests/app/main/test_services.py
```diff
@@ -2962,3 +3013,58 @@ def test_the_correct_flash_message_should_be_set(self, flash):
...
+    def teardown_method(self, method):
+        self.data_api_client_patch.stop()
+        super().teardown_method(method)
+
+    @pytest.mark.parametrize("framework_slug, framework_status, supplier_id, allows_edits", (
+        ("g-cloud-12", "live", 577184, True),      # allows edits for g12 recovery supplier
+        ("g-cloud-12", "live", 1, False),          # does not allow edits if not g12 recovery supplier
+        ("g-cloud-12", "pending", 577184, False),  # does not allow edits if not live
+        ("g-cloud-12", "open", 577184, True),      # still allows edits if g12 open
+        # ("g-cloud-11", "live", 577184, False),  # does not allow edits if live but not g12
```
bjgill, line 3036: Can this be un-commented?

bjgill (commented):
As discussed, this all looks fine to me. As a recovery supplier user, I can navigate to the draft service page and then edit and submit my drafts. So I think we've done everything needed by the story.

The journey to get to the draft service page is a bit wonky and the breadcrumbs are rather broken. That'll be fixed by a story next sprint, though, so I think that's OK and doesn't need to be fixed in this PR.

I particularly liked the way you've been able to avoid the large amounts of copy-paste I had assumed would be needed - very efficient :slightly_smiling_face:.

Just one minor question. Happy to approve once you've fixed that (and the tests which I think you're currently looking at).

tests/app/main/test_services.py
lfdebrux: Yes it can, thanks for catching that

bjgill (approved):
:tada:

PR #1297, opened 2020-12-15
Add G12 recovery banner to supplier dashboard

bjgill:

We show this banner to suppliers involved in the G12 recovery - and none others. It will help them understand when the deadline is for the recovery process and what they need to do.

Content is not final - just putting the banner into place for the moment.

Hard-code the closing date. This is fine because we're going to remove all this code after the recovery process is complete, so it doesn't need to be re-usable.

Add tests to check that this shows up iff the supplier is a recovery supplier.

image

### DIFF
--- a/app/templates/partials/g12_recovery_information.html
+++ b/app/templates/partials/g12_recovery_information.html
@@ -1,0 +1,12 @@
+{% if supplier.g12_recovery %}
+  <div class="wrapper">
+    {# TODO: replace with govukNotificationBanner #}
+    {%
+      with
+      message = 'You have until 23:59 on 1 January 2525 to complete <a href="'|safe + url_for('.list_services', framework_slug='g-cloud-12') + '">your additional G12 services</a>.'|safe,
+      type = 'information'
+    %}
+      {% include 'toolkit/notification-banner.html' %}
+    {% endwith %}
+  </div>
+{% endif %}
--- a/tests/app/main/test_suppliers.py
+++ b/tests/app/main/test_suppliers.py
@@ -662,6 +662,26 @@ class TestSuppliersDashboard(BaseApplicationTest):
                 u3="/suppliers/frameworks/g-cloud-12",
             )
 
+    def test_recovery_supplier_sees_banner(self):
+        self.data_api_client.get_supplier.return_value = get_supplier(id=577184)  # Test.DM_G12_RECOVERY_SUPPLIER_IDS
+        self.login()
+
+        with self.app.app_context():
+            response = self.client.get("/suppliers")
+        doc = html.fromstring(response.get_data(as_text=True))
+
+        assert doc.cssselect("p.banner-message:contains('You have until')")
+
+    def test_supplier_cannot_see_banner(self):
+        self.data_api_client.get_supplier.return_value = get_supplier(id=1)
+        self.login()
+
+        with self.app.app_context():
+            response = self.client.get("/suppliers")
+        doc = html.fromstring(response.get_data(as_text=True))
+
+        assert not doc.cssselect("p.banner-message:contains('You have until')")
+
 
 class TestSupplierDetails(BaseApplicationTest):
 
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -67,10 +67,14 @@ def list_services(framework_slug):
         framework=framework_slug,
     )["services"]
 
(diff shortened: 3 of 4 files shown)


tests/app/main/test_suppliers.py
```diff
@@ -662,6 +662,26 @@ def test_recovery_supplier_sees_g12_links(self, framework_interest, on_framework
                 u3="/suppliers/frameworks/g-cloud-12",
             )
 
+    def test_recovery_supplier_sees_banner(self):
+        self.data_api_client.get_supplier.return_value = get_supplier(id='577184')  # Test.DM_G12_RECOVERY_SUPPLIER_IDS
```
agalamatis, line 666: I think thee id is an integer, instead of string. Either way, better to test with both types.

app/templates/partials/g12_recovery_information.html
```diff
@@ -0,0 +1,11 @@
+{% if supplier.g12_recovery %}
+  <div class="wrapper">
+    {%
+      with
+      message = 'You have until 23:59 on 1 January 2525 to complete <a href="'|safe + url_for('.list_services', framework_slug='g-cloud-12') + '">your additional G12 services</a>.'|safe,
+      type = 'information'
+    %}
+      {% include 'toolkit/notification-banner.html' %}
```
gidsg, line 8: Could we use the new notification banner from the design system: https://design-system.service.gov.uk/components/notification-banner/ rather than the legacy toolkit?  However, if this is a lot of work may make more sense as a TODO?

tests/app/main/test_suppliers.py
bjgill: See https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1288/ - we handle it being either a string or integer.

app/templates/partials/g12_recovery_information.html
bjgill: Hmm. I'm not very familiar with the design system. Do we have any example of how to use the design system banner?

bjgill: On further thought, I didn't think we'd ported supplier-frontend over to the Design System, so is this even possible?

Given that this should only exist in the codebase for a couple of months, I'm not sure we need to do any more than the bare minimum.

tests/app/main/test_suppliers.py
bjgill: I've switched to integers.

app/templates/partials/g12_recovery_information.html
gidsg: @bjgill 
>  I didn't think we'd ported supplier-frontend over to the Design System, so is this even possible?

Yes, we do use `govuk-frontend` components alongside the old toolkit in supplier-frontend. Here is an example usage:  https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/templates/frameworks/legal_authority.html that said, this is a new component so I guess we may have to update dependencies to pick it up?

bjgill: I see what you mean. There exists a `banner` component that sounds as if it might be what we want: https://github.com/alphagov/digitalmarketplace-govuk-frontend/tree/master/src/digitalmarketplace/components/banner. However, I can see no examples of us previously using the `banner` component.

Would it be OK to merge this as-is, and make a note in the story to look into switching to the `govuk-frontend` component later?

bjgill: Actually - ^ is a DM component, not a govuk one.

There is a govuk component that does what we want, the `notification-banner`. However, that would need us to upgrade govuk-frontend to [3.10.0](https://github.com/alphagov/govuk-frontend/blob/master/CHANGELOG.md#3100-feature-release) - we're currently using 2.8.0.

Given the time pressure and temporary nature of this code, I don't think that this story should cover upgrading govuk-frontend. So, unless you object, I propose to leave as-is.

gidsg: @bjgill 👍  fine, perhaps we can add a TODO, in case we get around to switching this?

gidsg (approved):

app/templates/partials/g12_recovery_information.html
bjgill: Added.

PR #1298, opened 2020-12-16
Update content in G12 recovery banner

bjgill:

The G12 recovery process is for suppliers who were affected by the outage during G12 closing. We will give these suppliers an opportunity to submit the services that the outage disrupted.

In https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1297, we created a banner that we will display to suppliers involved in the G12 recovery - and none others. It will help them understand when the deadline is for the recovery process and what they need to do.

This PR updates the banner with the content from @NoraGDS 

(Note that the deadline is currently hardcoded to a placeholder in 2525. We'll need to update when we know the actual date.)

image

### DIFF
--- a/app/main/helpers/suppliers.py
+++ b/app/main/helpers/suppliers.py
@@ -1,6 +1,7 @@
 import os
 import json
 from typing import Union
+from datetime import datetime, timedelta
 
 from flask import current_app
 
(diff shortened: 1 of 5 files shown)


alex9smith (approved):
LGTM!

PR #1303, opened 2020-12-18
Set the G12 recovery deadline in the past 

bjgill:

We don't yet know when the deadline is, so set it to a dummy date in the far future. However, this means that the number of days remaining decrements every day and breaks the visual regression tests.

Switch to a dummy date in the far past. The banner will now always read '0 seconds remaining'.

Also fix a bug I found with dates more than 1 day in the past.

### DIFF
--- a/app/main/helpers/suppliers.py
+++ b/app/main/helpers/suppliers.py
@@ -94,13 +94,13 @@ def g12_recovery_time_remaining() -> str:
 
 
 def format_g12_recovery_time_remaining(time_to_deadline: timedelta) -> str:
-    if time_to_deadline.days > 0:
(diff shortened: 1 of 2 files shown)


alex9smith (approved):
Good spot on the negative days message.

I like that this is a nice simple fix for the functional tests - although it'll only work until we need to set the actual date. Maybe at that point we can remove our test supplier as a G12 recovery supplier? It's probably not something we need to worry about at the moment

bjgill:
Yes - I suspect we'll want to do something of the sort. I've made a note in the 'preparing for user research' story, so we'll think about it then.

PR #1306, opened 2020-12-21
Stop DOS5 text clipping

gidsg:
https://trello.com/c/75hbcRW8/607-3-implement-dos-5-e-signature
Previously on the e-signature confirmation page the framework name was clipping as it the `nowrap` style prevented it from wrapping. For now this just removes this so it doesn't clip. Although it will wrap for G-Cloud-12.
**Before**
 Screenshot 2020-12-21 at 14 26 14
**After**
Screenshot 2020-12-21 at 16 05 23


domoscargin (approved):
👍

PR #1314, opened 2021-01-05
Hide 'Remove draft service' button for G12 recovery suppliers

alex9smith:
We're going to stop G12 recovery suppliers from creating new draft services other than the ones they've already submitted in previous recovery phases, so we need to prevent a supplier from accidentally deleting a draft and needing to contact support to get it re-added.

This PR hides the 'Remove draft service' button on the service submission page for G12 recovery suppliers. I've checked other pages and can't see any other method for a supplier to delete a draft service, so I think hiding this button is enough to finish the below story.

https://trello.com/c/1Jgs4nZ4/684-1-stop-users-being-able-to-delete-draft-services

### DIFF
--- a/app/templates/services/service_submission.html
+++ b/app/templates/services/service_submission.html
@@ -199,7 +199,7 @@
             {% include "partials/complete_service.html" %}
           </div>
         {% endif %}
-        {% if framework.status == 'open' %}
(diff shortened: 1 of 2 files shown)


bjgill (approved):
Nice and simple - looks good.

PR #1317, opened 2021-01-08
Change submissions and service submissions pages to redirect to all services page for G12 recovery suppliers

lfdebrux:
Ticket: https://trello.com/c/QBCaiojj/698-fix-mark-as-complete-redirect-in-g12-recovery-flow

For G12 recovery suppliers we need to fix breadcrumbs that link to the "Services" and lot page, and also change where the user ends up after marking a service as complete.

This commit does this in a fairly brute-force way, by making those urls always redirect to the `all_services` route; however I think this should be fine because they will not need to see the pages we are redirecting away from anyway.

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -259,6 +259,10 @@ def framework_dashboard(framework_slug):
 def framework_submission_lots(framework_slug):
     framework = get_framework_or_404(data_api_client, framework_slug)
 
+    if is_g12_recovery_supplier(current_user.supplier_id) and framework_slug == "g-cloud-12":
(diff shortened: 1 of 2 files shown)


bjgill (commented):
A couple of test failures - but nothing that looks significant. Happy to approve once you've fixed up the tests.

lfdebrux:
> A couple of test failures - but nothing that looks significant. Happy to approve once you've fixed up the tests.

Fixed

bjgill (approved):
Looks good. I think these are the pages we tinkered with before we'd settled on the final agreed user flow. So it's safe to excise them from G12 recovery support.

PR #1322, opened 2021-01-19
Always fail open if D&B API errors

bjgill:
Trello: https://trello.com/c/bZ4XlASA/617-1-log-directplusclient-errors-instead-of-swallowing-them

At the moment, we allow users to create an account if the D&B API produces an error that is valid JSON, but not if it's not valid JSON. This is odd and inconsistent. And I think it's unintentional - looking at utils, it looks as if the original author didn't know that about the different error for invalid JSON.

Change the behaviour to be consistent by always failing open. The behaviour now fully matches the intents of the comments.

Also pull in alphagov/digitalmarketplace-utils#596, which improves logging of D&B API errors.


alex9smith (approved):
LGTM!

PR #1323, opened 2021-01-20
Remove manual signing endpoints

bjgill:
Trello: https://trello.com/c/mtsSX6a5/719-2-delete-unused-non-e-signature-framework-code

All frameworks from DOS 5 and G12 onwards use e-signature.

The manual signature flow was used by all frameworks up to DOS 4 and G11. All those frameworks are now closed. So this flow will never be used again. So we can delete all endpoints in that flow to reduce the amount of code we need to support.

Suppliers who want to see their countersigned agreements for a closed framework will need to raise a support ticket. But this is already what they need to do - this PR doesn't change that. So that's fine.


alex9smith (approved):
This looks good to me. Do we have any functional or visual regression tests that touch these routes?

bjgill:
Good point - I didn't think so, but hadn't checked.

I've just taken a look and can't see anything related to manual signing in either.

PR #1325, opened 2021-01-27
Show only selected drafts for the G12 recovery

bjgill:
Trello: https://trello.com/c/13Ryt5o5/679-5-as-a-supplier-i-cannot-see-submit-any-draft-services-that-ccs-will-not-consider

As part of the G12 recovery process, users are only able to submit certain drafts. To avoid confusion and unnecessary work, we're going to hide all other drafts from the G12 recovery process for suppliers.

Only show drafts that are in the allow-list - both submitted and un-submitted.

Tests prove that this is working.

This PR is very similar to https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1288.

### DIFF
--- a/app/main/helpers/suppliers.py
+++ b/app/main/helpers/suppliers.py
@@ -1,6 +1,6 @@
 import os
 import json
-from typing import Union
+from typing import Union, Set
 from datetime import datetime, timedelta
 
 from flask import current_app
@@ -84,6 +84,18 @@ def is_g12_recovery_supplier(supplier_id: Union[str, int]) -> bool:
         return False
 
     return int(supplier_id) in supplier_ids
+
+
+def get_g12_recovery_draft_ids() -> Set[int]:
+    draft_ids_string = current_app.config.get('DM_G12_RECOVERY_DRAFT_IDS') or ''
+
+    try:
+        return {int(s) for s in draft_ids_string.split(sep=',')}
+    except AttributeError as e:
+        current_app.logger.error("DM_G12_RECOVERY_DRAFT_IDS not a string", extra={'error': str(e)})
+    except ValueError as e:
+        current_app.logger.error("DM_G12_RECOVERY_DRAFT_IDS not a list of draft IDs", extra={'error': str(e)})
+    return set()
 
 
 G12_RECOVERY_DEADLINE = datetime(year=1970, month=1, day=1, hour=17)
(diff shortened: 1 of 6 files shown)


app/main/helpers/suppliers.py
```diff
@@ -86,6 +86,18 @@ def is_g12_recovery_supplier(supplier_id: Union[str, int]) -> bool:
...
+def get_g12_recovery_draft_ids() -> Set[int]:
+    draft_ids_string = current_app.config.get('DM_G12_RECOVERY_DRAFT_IDS') or ''
+
+    try:
+        return {int(s) for s in draft_ids_string.split(sep=',')}
+    except AttributeError as e:
+        current_app.logger.error("DM_G12_RECOVERY_DRAFT_IDS not a string", extra={'error': str(e)})
+    except ValueError as e:
+        current_app.logger.error("DM_G12_RECOVERY_DRAFT_IDS not a list of draft IDs", extra={'error': str(e)})
+    return set()
```
lfdebrux, line 98: It might be better to do this once when initialising the config object rather than every time we need the list?

tests/app/main/helpers/test_suppliers.py
```diff
@@ -75,6 +75,25 @@ def test_returns_expected_value_for_input(self, g12_recovery_supplier_ids, expec
...
+    @pytest.mark.parametrize(
+        'draft_ids_config, expected_result',
+        [
+            (None, set()),
+            ('', set()),
+            (42, set()),
+            ('12:32', set()),
+            ([123456, 789012], set()),
+            ('123456', {123456}),
+            ('123456,789012', {123456, 789012}),
```
lfdebrux, line 88: Is it worth testing what happens if a stray space or newline gets in there?

bjgill: Good idea - added.

app/main/helpers/suppliers.py
bjgill: I agree that that would be preferable. Do you know of a way to do that without conflicting with the magic that loads config values from environment variables? Is there any precedence for this?

If there's a known way of doing this, I think it would be worth switching. Otherwise, I don't think it would be worth the trouble for something that only gets called once per page, and where the code will be deleted in a month's time.

bjgill: I think I've worked out a way of doing it - I can load the list inside `init_app`.

Since this applies to both the list of draft IDs and supplier IDs, I'd like to suggest merging this PR as-is. I'll sort out loading this only once in a follow-up PR.

lfdebrux (approved):

PR #1326, opened 2021-01-28
Parse list of IDs from config only once

bjgill:
Trello: https://trello.com/c/13Ryt5o5/679-5-as-a-supplier-i-cannot-see-submit-any-draft-services-that-ccs-will-not-consider

We have some lists of IDs that we use as allow-lists. These are needed every time certain pages are loaded.

Previously, we did the string parsing every time. Now, perform the string parsing during app initialisation and store the result. This should remove the performance impact of these new allowlists.

Follows on from https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1325#discussion_r566014562

### DIFF
--- a/app/main/helpers/suppliers.py
+++ b/app/main/helpers/suppliers.py
@@ -1,6 +1,6 @@
 import os
 import json
-from typing import Union, Set
+from typing import Union
 from datetime import datetime, timedelta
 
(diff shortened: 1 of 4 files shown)


lfdebrux (approved):
👍

PR #1327, opened 2021-01-28
Extract ID lists to new keys

bjgill:
We've recently added functionality to get allow-lists of IDs from environment variables. We pass them in as environment variables, and then convert them from a comma-separated string into a `Set[int]`. However, it seems as if something is [going wrong](https://ci.marketplace.team/job/apps-are-up-preview/257100/smoke_20test_20report/). Either the extraction isn't happening, or the Set somehow gets overwritten with the original string.

My hypothesis is that it's the latter - our app init is somewhat complicated, so it seems plausible. If I'm right, we can fix it by writing to a new key so it won't get overwritten.

If I'm wrong, this won't fix Preview, and I'll just revert everything.

Fixes bug introduced by https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1326

### DIFF
--- a/app/main/helpers/suppliers.py
+++ b/app/main/helpers/suppliers.py
@@ -72,11 +72,11 @@ def get_company_details_from_supplier(supplier):
 
 
 def is_g12_recovery_supplier(supplier_id: Union[str, int]) -> bool:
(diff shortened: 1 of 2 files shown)


gidsg (approved):
🤞

PR #1329, opened 2021-01-29
Change G12 banner when all drafts complete

alex9smith:
In our user research we discovered that users expected the banner to change when they had completed all their drafts.

Change the banner to a success message if the user has no unsubmitted G12 recovery drafts (see #1325).

Content on this banner is not final.

There is G12 recovery logic to show a banner on the `/frameworks/<string:framework_slug>/services` route, however I've not updated this as we've now added a new route (`/all-services`, see #1301) for the G12 recovery suppliers to see their services and therefore the banner will never be shown on the `/services` route.

image

https://trello.com/c/p3BVSpj6/740-2-change-banner-when-all-drafts-complete

### DIFF
--- a/app/main/helpers/suppliers.py
+++ b/app/main/helpers/suppliers.py
@@ -125,11 +125,10 @@ def format_g12_recovery_time_remaining(time_to_deadline: timedelta) -> str:
 
 
 def count_unsubmitted_g12_recovery_drafts(data_api_client: DataAPIClient, supplier_id: int) -> int:
-    unsubmitted_recovery_drafts = [
-        draft for draft in
+    unsubmitted_draft_ids = {
+        draft["id"] for draft in
         data_api_client.find_draft_services_by_framework_iter(
             framework_slug="g-cloud-12", supplier_id=supplier_id, status="not-submitted"
         )
-        if draft["id"] in get_g12_recovery_draft_ids()
-    ]
-    return len(unsubmitted_recovery_drafts)
+    }
+    return len(unsubmitted_draft_ids.intersection(get_g12_recovery_draft_ids()))
--- a/app/main/helpers/suppliers.py
+++ b/app/main/helpers/suppliers.py
@@ -4,6 +4,7 @@ from typing import Union, Set
 from datetime import datetime, timedelta
 
 from flask import current_app
+from dmapiclient import DataAPIClient
 
 
 def load_countries():
@@ -121,3 +122,14 @@ def format_g12_recovery_time_remaining(time_to_deadline: timedelta) -> str:
         unit += 's'
 
     return f'{number} {unit}'
+
+
+def count_unsubmitted_g12_recovery_drafts(data_api_client: DataAPIClient, supplier_id: int) -> int:
+    unsubmitted_recovery_drafts = [
+        draft for draft in
+        data_api_client.find_draft_services_by_framework_iter(
+            framework_slug="g-cloud-12", supplier_id=supplier_id, status="not-submitted"
+        )
+        if draft["id"] in get_g12_recovery_draft_ids()
+    ]
+    return len(unsubmitted_recovery_drafts)
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -14,7 +14,12 @@ from dmutils.flask import timed_render_template as render_template
 from dmutils.forms.helpers import get_errors_from_wtform
 from dmutils.errors import render_error_page
 
-from ..helpers.suppliers import is_g12_recovery_supplier, g12_recovery_time_remaining, get_g12_recovery_draft_ids
+from ..helpers.suppliers import (
+    is_g12_recovery_supplier,
+    g12_recovery_time_remaining,
+    get_g12_recovery_draft_ids,
+    count_unsubmitted_g12_recovery_drafts,
+)
 from ... import data_api_client
 from ...main import main, content_loader
 from ..helpers import login_required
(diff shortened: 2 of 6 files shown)


app/main/helpers/suppliers.py
```diff
@@ -121,3 +122,15 @@ def format_g12_recovery_time_remaining(time_to_deadline: timedelta) -> str:
         unit += 's'
 
     return f'{number} {unit}'
+
+
+def count_unsubmitted_g12_recovery_drafts(data_api_client: DataAPIClient, supplier_id: int) -> int:
+    unsubmitted_recovery_drafts = [
+        draft for draft in
+        data_api_client.find_draft_services_iter(
```
bjgill, line 130: Can we reduce the number of network requests needed by using `find_draft_services_by_framework_iter`, which can get the API to filter by status for us?

app/main/helpers/suppliers.py
```diff
@@ -121,3 +122,15 @@ def format_g12_recovery_time_remaining(time_to_deadline: timedelta) -> str:
...
+            supplier_id=supplier_id, framework="g-cloud-12"
+        )
+        if draft["id"] in get_g12_recovery_draft_ids()
```
bjgill, line 133: From my work on https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1325, I know that `get_g12_recovery_draft_ids` takes ~>0.1ms to run. Can we reduce the performance impact of this for suppliers with very large numbers of draft G12 services?

bjgill (commented):
Your approach looks sensible. Just a couple of questions about the performance impact.

If it helps, this is the microbenchmark I used. It runs on my machine in ~1.5s for a list of ~200 IDs.

```
    def test_benchmark(self):
        self.data_api_client.find_draft_services_iter.return_value = [
            DraftServiceStub(
                id=id,
                service_name="test",
                framework_slug="g-cloud-12",
                status="not-submitted",
            ).response()
            for id in []  # replace with large list of IDs
        ]
        for _ in range(100):
            with self.app.app_context():
                assert count_unsubmitted_g12_recovery_drafts(self.data_api_client, self.g12_recovery_draft_id) == 0
```

app/main/helpers/suppliers.py
alex9smith: That sounds very sensible

app/main/helpers/suppliers.py
alex9smith: Hmm ok. That's longer than I thought. The list of G12 drafts will only change if the app is restarted, so we could save the result of `get_g12_recovery_draft_ids()` as a variable in `helpers/suppliers.py` so it only gets called once and not on every page load.

That doesn't really help the case for suppliers with many services though. I don't think we can cache the supplier's drafts in the same way because (hopefully) they're going to go through and submit them, so we're still going to need to check membership of each element in a list against a set. 

I've done some testing and it looks like it's always faster to convert the list to a set and find the intersection than it is to iterate through, which surprised me:

```
local rw
In [24]: LONG_LIST = [i for i in range(1000)]

local rw
In [25]: LONG_SET = {i for i in range(1000)}

local rw
In [26]: def iterate_through_list(long_list, long_set):
    ...:     return len([i for i in long_list if i in long_set])

local rw
In [27]: def convert_to_set_then_intersection(long_list, long_set):
    ...:     return len(set(long_list).intersection(long_set))

local rw
In [28]: %timeit iterate_through_list(LONG_LIST, LONG_SET)
48.2 µs ± 230 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

local rw
In [29]: %timeit convert_to_set_then_intersection(LONG_LIST, LONG_SET)
32.7 µs ± 268 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

local rw
In [30]: SHORT_LIST = [i for i in range(10)]

local rw
In [31]: %timeit iterate_through_list(SHORT_LIST, LONG_SET)
1.11 µs ± 462 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

local rw
In [32]: %timeit convert_to_set_then_intersection(SHORT_LIST, LONG_SET)
648 ns ± 4.17 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

I presume a set comprehension builds a list as it goes and then converts to a set at the end. So I think we can do something like this (ignoring the other change I mentioned above):
```
def count_unsubmitted_g12_recovery_drafts(data_api_client: DataAPIClient, supplier_id: int) -> int:
    unsubmitted_draft_ids = {
        draft["id"] for draft in
        data_api_client.find_draft_services_by_framework_iter(
            framework_slug="g-cloud-12", supplier_id=supplier_id, status="not-submitted"
        )
    }
    return len(unsubmitted_draft_ids.intersection(get_g12_recovery_draft_ids()))
```

Does that sound sensible?

bjgill: :+1: That all looks very sensible.

bjgill: (as an aside, by your previous comment, you have introduced me to the wonderful world of IPython magic commands. Thanks!)

alex9smith: The set parts work fine, but I can't store the output of `get_g12_recovery_draft_ids()` in `helpers/suppliers.py` because it needs the app context. We don't have that at import time for the tests, so the test suite refuses to even run 😞 .

My next thought for caching the result was to try something in `init_app` or the config, but that's what you looked at in #1326 and #1327 which broke Preview.  Unless you have any more ideas I think we might have to leave this as is.

bjgill: Ah - sorry. I misread your original comment. I agree that trying to cache the result of `get_g12_recovery_draft_ids` is a hard problem and not one worth fixing here.

Instead, what I meant was that in the original code, you appeared to be calling `get_g12_recovery_draft_ids` once per draft service. So 0.1ms/call would quickly add up for suppliers with 100-200 drafts.

But it's only ~0.1ms, so calling it once every time the page is loaded should have a negligible performance impact. So I think your proposed code where you call `get_g12_recovery_draft_ids` once inside the `intersection` should be easily sufficient.

bjgill (approved):
:shipit:

NoraGDS:
You have marked all your services as complete. Your services will be automatically submitted on 25 February.

[some suppliers may not complete all draft services. should we also consider?]
You have marked X services as complete. Your services will be automatically submitted on 25 February.

PR #1331, opened 2021-02-02
Update G12 recovery banner with final content

alex9smith:
In #1329 we added logic to the banner so it showed a different message if the supplier had submitted all their draft services. We hadn't decided what the content should be at that time.

In content review, we decided that this behaviour wasn't quite what we wanted.

Update the banner with the final content, and add more logic so the banner tells the supplier how many services they've marked as complete as they go. If the supplier completes all their services, show a third message.

Note the nested `{% with %}` in the template. Putting `"service" if g12_recovery.submitted_drafts == 1 else " services"`
in the string concatenation for `message` caused Jinja to only render the string up to and including service(s). I don't like this solution very much, but it's the only one I was able to get working. If there is a better way, please let me know.

https://trello.com/c/p3BVSpj6/740-2-change-banner-when-all-drafts-complete

image

### DIFF
--- a/app/templates/partials/g12_recovery_information.html
+++ b/app/templates/partials/g12_recovery_information.html
@@ -1,13 +1,23 @@
 {% if g12_recovery %}
   <div class="wrapper">
     {# TODO: replace with govukNotificationBanner #}
-    {% if g12_recovery.drafts_remaining == 0 %}
+    {% if g12_recovery.not_submitted_drafts == 0 %}
       {%
         with
-        message = 'Thank you for completing your application.',
+        message = 'You have marked all your services as complete. Your services will be automatically submitted on 25 February.',
         type = 'success'
       %}
         {% include 'toolkit/notification-banner.html' %}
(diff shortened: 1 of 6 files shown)


app/templates/partials/g12_recovery_information.html
```diff
@@ -1,14 +1,24 @@
 {% if g12_recovery %}
   <div class="wrapper">
     {# TODO: replace with govukNotificationBanner #}
-    {% if g12_recovery.drafts_remaining == 0 %}
+    {% if g12_recovery.not_submitted_drafts == 0 %}
       {%
         with
-        message = 'Thank you for completing your application.',
+        message = 'You have marked all your services as complete. Your services will be automatically submitted on 25 February.',
```
bjgill, line 7: Do we want to give a time?

bjgill (approved):
Looks good - like the comprehensive tests.

Just one query about the content - feel free to merge once that's resolved.

PR #1335, opened 2021-02-05
Fix breadcrumbs for G12 recovery

bjgill:
Trello: https://trello.com/c/k4g4FDoX/787-2-odd-breadcrumbs-on-draft-service-page

The standard breadcrumbs don't work for the G12 recovery, because the user journey is different to normal. Mainly, we have the new 'Your G-Cloud 12 services' page that covers both the services and the lot page from a normal framework. So update the breadcrumbs accordingly.

This will break the breadcrumbs for non-G12 recovery suppliers. That's fine because no framework will be open during the G12 recovery. We will revert this code once the G12 recovery process has closed.

With changes:

image
image

### DIFF
--- a/app/templates/services/edit_submission_section.html
+++ b/app/templates/services/edit_submission_section.html
@@ -11,16 +11,8 @@
         "href": url_for('.dashboard')
       },
       {
-        "text": "Apply to " + framework.name,
-        "href": url_for(".framework_dashboard", framework_slug=framework.slug)
-      },
-      {
-        "text": "Services",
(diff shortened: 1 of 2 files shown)


bjgill:
Huh. Apparently we don't have any tests for the breadcrumbs on these pages. Given the screenshots, I'm not sure these need further testing.

alex9smith (approved):
🐛  🚫

bjgill:
Yes - I don't think there's anything meaningful to test here. Merging...

PR #1338, opened 2021-02-09
optimise caching

gidsg:
Optimise python caching, based on template https://github.com/alphagov/.github/blob/master/workflow-templates/digitalmarketplace-python-ci.yml

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,18 +9,22 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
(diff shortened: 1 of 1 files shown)


bjgill (approved):
Looks sensible. Would the same thing help with the node dependencies?

gidsg:
@bjgill I took a look at this but it looks like the npm step is wrapped by `make test` so we don't have a good way to conditionally run it.

bjgill (approved):

PR #1340, opened 2021-02-12
Replace validation banner with GOV.UK Frontend error summary component

alex9smith:
As part of migrating our apps to use GOV.UK Design System styles and components, we want to use the [error summary component](https://design-system.service.gov.uk/components/error-summary/) to replace our existing validation banners.

To do this, we add an instance of the error summary component to the app base page template. This ensures that the error summary (if present) is in a consistent location across all pages in the app.

https://trello.com/c/WVjA4w66/771-2-replace-validation-banner-with-govuk-frontend-error-summary-component-in-the-supplier-frontend

### DIFF
--- a/app/main/forms/suppliers.py
+++ b/app/main/forms/suppliers.py
@@ -91,11 +91,13 @@ class AddCompanyRegisteredNameForm(FlaskForm):
 class AddCompanyRegistrationNumberForm(FlaskForm):
     has_companies_house_number = RadioField(
         "Are you registered with Companies House?",
+        id="input-has_companies_house_number-1",
         validators=[InputRequired(message="You need to answer this question.")],
         choices=[('Yes', 'Yes'), ('No', 'No')]
     )
     companies_house_number = DMStripWhitespaceStringField(
         'Companies House number',
+        id="input-has_companies_house_number-1-companies_house_number",
         default='',
         validators=[
             Optional(),
@@ -106,6 +108,7 @@ class AddCompanyRegistrationNumberForm(FlaskForm):
     )
     other_company_registration_number = DMStripWhitespaceStringField(
         'Other company registration number',
+        id="input-has_companies_house_number-2-other_company_registration_number",
         default='',
         validators=[
             Optional(),
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -840,6 +840,12 @@ def framework_updates(framework_slug, error_message=None, default_textbox_value=
         file['path'] = '/'.join(path_parts[2:])
         files[path_parts[3]].append(file)
 
+    errors = {CLARIFICATION_QUESTION_NAME: {
+        "text": error_message,
+        "href": "#" + CLARIFICATION_QUESTION_NAME,
+        "errorMessage": error_message,
+    }} if error_message else {}
+
     return render_template(
         "frameworks/updates.html",
         framework=framework,
@@ -847,6 +853,8 @@ def framework_updates(framework_slug, error_message=None, default_textbox_value=
         clarification_question_name=CLARIFICATION_QUESTION_NAME,
         clarification_question_value=default_textbox_value,
         error_message=error_message,
+        errors=errors,
+        error_title="There was a problem with your submitted question",
         files=files,
         agreement_countersigned=bool(supplier_framework_info and supplier_framework_info['countersignedPath']),
     ), 200 if not error_message else 400
@@ -867,7 +875,7 @@ def framework_updates_email_clarification_question(framework_slug):
     clarification_question = request.form.get(CLARIFICATION_QUESTION_NAME, '').strip()
 
     if not clarification_question:
-        return framework_updates(framework_slug, "Add text if you want to ask a question.")
+        return framework_updates(framework_slug, error_message="Add text if you want to ask a question.")
     elif len(clarification_question) > 5000:
         return framework_updates(
             framework_slug,
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -5,6 +5,7 @@ from datetime import datetime, timedelta
 from dmutils.errors import render_error_page
 from itertools import chain
 
+from dmutils.forms.errors import govuk_errors
 from flask import request, abort, flash, redirect, url_for, current_app, session
 from flask_login import current_user
 
@@ -710,7 +711,7 @@ def framework_supplier_declaration_edit(framework_slug, section_id):
             )
 
             if document_errors:
-                errors = section.get_error_messages(document_errors, question_descriptor_from="question")
+                errors = govuk_errors(section.get_error_messages(document_errors, question_descriptor_from="question"))
             else:
                 submitted_answers.update(uploaded_documents)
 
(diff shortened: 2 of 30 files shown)


alex9smith:
We're going to check the functional tests for anything that looks for the old components or wording

alex9smith:
Marking this as draft until we've sorted out the tests

alex9smith:
Unit tests are passing so this PR is ready for review while I work on the functional tests.

lfdebrux:
### My review checklist

I've:

- [ ] looked at the code
- [ ] tested locally

app/main/views/frameworks.py
```diff
@@ -839,13 +840,21 @@ def framework_updates(framework_slug, error_message=None, default_textbox_value=
         file['path'] = '/'.join(path_parts[2:])
         files[path_parts[3]].append(file)
 
+    errors = {CLARIFICATION_QUESTION_NAME: {
+        "text": error_message,
+        "href": f"#{CLARIFICATION_QUESTION_NAME}",
```
lfdebrux, line 845: Normally I prefer f-strings for any sort of string manipulation, but I wonder if in this case it would be more readable to do simple concatenation, as for a second here I thought I was reading Ruby... What are your thoughts?

app/main/views/suppliers.py
```diff
@@ -580,12 +580,22 @@ def duns_number():
...
     errors = get_errors_from_wtform(form)
+    error_title = None
+    error_description = None
+    if errors.get("duns_number", {}).get("message", None) == 'DUNS number already used':
+        error_title = "A supplier account already exists with that DUNS number"
+        support_email_address = current_app.config['SUPPORT_EMAIL_ADDRESS']
+        error_description = Markup(
+            'If you no longer have your account details, or if you think this may be an error, '
+            f'email <a class="govuk-link" href="mailto:{support_email_address}?subject=DUNS%20number%20question" '
+            f'title="Please contact {support_email_address}">{support_email_address}</a>')
```
lfdebrux, line 591: What was the reasoning behind moving this out of the template?

app/templates/_base_page.html
```diff
@@ -63,6 +63,15 @@
       {% endfor %}
     {% endwith %}
   {% endblock flashMessages %}
+  {% block errorSummary %}
+    {% if errors %}
+      {{ govukErrorSummary({
+      "titleText": error_title or "There is a problem",
+      "descriptionText": error_description or None,
+      "errorList": errors.values(),
+    }) }}
```
lfdebrux, line 72: Indentation
```suggestion
        "titleText": error_title or "There is a problem",
        "descriptionText": error_description or None,
        "errorList": errors.values(),
      }) }}
```

app/templates/_base_page.html
```diff
@@ -63,6 +63,15 @@
       {% endfor %}
     {% endwith %}
   {% endblock flashMessages %}
+  {% block errorSummary %}
+    {% if errors %}
+      {{ govukErrorSummary({
+      "titleText": error_title or "There is a problem",
+      "descriptionText": error_description or None,
```
lfdebrux, line 70: I think we need to use `descriptionHtml`, for when there is an embedded link

tests/app/main/test_suppliers.py
```diff
@@ -2988,31 +2983,31 @@ def test_add_registration_number_page_loads(self):
                  'companies_house_number': '',
                  'other_company_registration_number': ''
                  },
-                'Are you registered with Companies House?',
+                'You need to answer this question.',
```
lfdebrux, line 2986: I'm confused, why has this error message text changed?

app/templates/_base_page.html
alex9smith: That's what I thought too at first, but wrapping the text with a link in `Markup()` seems to work fine: 

image

tests/app/main/test_suppliers.py
alex9smith: This is no different to any of the other radio button questions - the error message in the top summary is now the instructions for how to solve the problem rather than the question name. It's this question for reference: 

image

app/main/views/frameworks.py
alex9smith: Hmm yes I see your point. I don't think I have a strong opinion either way so I'm happy to change it

app/main/views/suppliers.py
alex9smith: We removed all of the other custom error logic / messages from the templates (eg. [app/templates/frameworks/updates.html](https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1340/files#diff-2dc1079cb1ccc1cf010d712b6ac4d4e01d9c0872852307d69d4a3dd54fabfb57L30)) when switching out the component. Everywhere else has the logic in the python side, so we moved this out of the template so it's in the same place for all pages.

tests/app/main/test_suppliers.py
lfdebrux: Ah, of course. I think we need to revise the error message content, I've added a ticket to do this later https://trello.com/c/AnpNWBC7/285-update-error-messages-for-create-a-supplier-account-journey

alex9smith: Yeah I agree. If we update them for the declaration questions as well we can revert #1343 and get rid of even more custom error logic. I'll put a comment on that card.

lfdebrux: The declaration question error message content is a whole other bag of worms unfortunately 😅 I tried to update them in the past, but it became cumbersome.

alex9smith: 😢

lfdebrux: For this page (`/suppliers/registration-number/edit`) the error summary does not link to the first radio button as it should do

alex9smith: Good spot, thanks

lfdebrux: It should be straightforward to fix

alex9smith: Should be fixed in c7bbcd7

lfdebrux: Oh, the page with the "Do. you have a companies house number?" question still isn't working, I think we also need to add `id` to `has_companies_house_number`

https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/main/forms/suppliers.py#L92

app/main/forms/suppliers.py
```diff
@@ -96,6 +96,7 @@ class AddCompanyRegistrationNumberForm(FlaskForm):
     )
     companies_house_number = DMStripWhitespaceStringField(
         'Companies House number',
+        id='companies_house_number',
```
lfdebrux, line 99: Because these are nested the ids have the id of the component they are nested under prepended by the frontend toolkit template, so this should be

```suggestion
        id='input-has_companies_house_number-1-companies_house_number',
```

Or at least that what the id for the input is in Chrome dev tools on my machine

app/main/forms/suppliers.py
```diff
@@ -106,6 +107,7 @@ class AddCompanyRegistrationNumberForm(FlaskForm):
     )
     other_company_registration_number = DMStripWhitespaceStringField(
         'Other company registration number',
+        id='other_company_registration_number',
```
lfdebrux, line 110:
```suggestion
        id='input-has_companies_house_number-2-other_company_registration_number',
```

tests/app/main/test_suppliers.py
alex9smith: Yup, adding IDs to all three fields seems to have sorted it

lfdebrux (approved):
This looks great to me, thanks for this change! 🏆

PR #1343, opened 2021-02-16
Change declaration errors to use question name in summary

alex9smith:
In #1340 we replaced the error summary on this page with the GOV.UK error summary. This has the side effect of changing the text in the summary from the question name to the error that needs fixing.

On most pages this is more useful, however the declaration pages have many radio button questions which all need an answer. Each of these questions has the same error message - 'You must answer this question'. It's not helpful for the user to see the same error repeated many times, so the previous behaviour of having the question name in the summary box was more useful.

https://trello.com/c/WVjA4w66/771-2-replace-validation-banner-with-govuk-frontend-error-summary-component-in-the-supplier-frontend


lfdebrux (approved):
Yeah ~this change~ keeping this behaviour the same for this page makes sense to me, at least while the journey is the way it is

PR #1344, opened 2021-02-16
Remove unnecessary <br> tags

domoscargin:
The combination of `govuk-body` styles mixed with `<br>` tags makes this page look odd:

Screenshot 2021-02-16 at 15 04 08

So let's jettison the `<br>` tags. And as a happy side effect, we can get rid of the custom `framework-dashboard` style as well.

Screenshot 2021-02-16 at 15 04 15


bjgill (approved):
:tada:

PR #1346, opened 2021-02-16
Close G12 recovery process

bjgill:
Trello: https://trello.com/c/mhab9cpq/826-1-as-a-developer-i-can-stop-suppliers-submitting-draft-services-after-the-deadline

Once the deadline arrives, we want to stop suppliers involved in the recovery from being able to edit and submit their draft services.

Make the simplest possible change to implement this - suppliers will now get 404s if they try to do either.

This should be something small and low-risk we can deploy at the deadline. After the deadline, we can tidy up all the other G12 recovery code still lying around.

## How to deploy

Merge this to master on the day of the deadline and get it through to staging. Deploy to production when the deadline hits.

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -499,12 +499,7 @@ def copy_draft_service(framework_slug, lot_slug, service_id):
 @EnsureApplicationCompanyDetailsHaveBeenConfirmed(data_api_client)
 @return_404_if_applications_closed(lambda: data_api_client)
 def complete_draft_service(framework_slug, lot_slug, service_id):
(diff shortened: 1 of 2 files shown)


gidsg:
@bjgill looks sensible to me, only thought is could we implement this behind config/env var? So it could be deployed now and the value changed at the intended time and we can then re-deploy the app to pick up the new config?

bjgill:
Hmm. That's a good point. I wonder if we can go one better, though, and do it automatically at the right time without a redeploy. I'll see if I can get that to work.

bjgill:
I think I've found a better way of doing this: https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1347

PR #1347, opened 2021-02-18
Automatically close G12 recovery at deadline

bjgill:
Trello: https://trello.com/c/mhab9cpq/826-1-as-a-developer-i-can-stop-suppliers-submitting-draft-services-after-the-deadline

Once the deadline arrives, we want to stop suppliers involved in the recovery from being able to edit and submit their draft services.

Make a simple change to implement this - suppliers will now get 404s if they try to do either after the deadline.

After the deadline, we can tidy up all the other G12 recovery code still lying around.

Supersedes: https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1346

### DIFF
--- a/app/main/helpers/suppliers.py
+++ b/app/main/helpers/suppliers.py
@@ -102,6 +102,10 @@ def get_g12_recovery_draft_ids() -> Set[int]:
 G12_RECOVERY_DEADLINE = datetime(year=2021, month=2, day=25, hour=14)
 
 
+def is_g12_recovery_open() -> bool:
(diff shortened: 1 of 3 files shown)


bjgill:
@gidsg - we had a quick discussion about this. As you said, this could cause problems if the deadline changes. However, the date is already hard-coded into the app for the banner, so I don't think this makes changing the deadline any more difficult.

gidsg (approved):
LGTM 🕐 👍

PR #1348, opened 2021-02-19
EXPERIMENTAL: lazily load manifests for old frameworks

bjgill:
Currently, it takes ~20 seconds to import all the content. This is a problem because it means that running the very smallest unit test includes ~20 seconds of content importing. This makes running the unit tests slower, which adds up across all developers.

Experiment with lazily importing manifests for old frameworks (prior to DOS5/G12). They will only be imported when needed.

From my testing, this seems to reduce the time taken to import all the content to ~5 seconds. This will save developers a lot of time, and also make deploys a little bit faster.

The trade-off is that the first user who needs a manifest for an old framework will have their request take ~1s longer. I think this is OK because users don't access information about old frameworks very often.

This is experimental - I'm creating this for feedback from other developers. If people think it's a good idea, I'll port these changes to content-loader.

For reference, in the past 7 days, we've seen the following request counts to supplier frontend endpoints that use manifest data:

* G7: 0
* G8: 0
* G9: 4
* G10: 10
* G11: 11
* G12: many
* DOS: unknown
* DOS2: 0
* DOS3: 0
* DOS4: 0
* DOS5: many

### DIFF
--- a/app/main/__init__.py
+++ b/app/main/__init__.py
@@ -1,11 +1,16 @@
+import collections
+import os
 from copy import deepcopy
+from functools import partial
+from typing import Dict
 
+from dmcontent.errors import ContentNotFoundError
 from flask import Blueprint, current_app
 from werkzeug.local import Local, LocalProxy
 
 from dmutils.direct_plus_client import DirectPlusClient
 
-from dmcontent.content_loader import ContentLoader
+from dmcontent.content_loader import ContentLoader, read_yaml
 
 from dmutils.timing import logged_duration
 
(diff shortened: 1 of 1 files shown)


lfdebrux:
> For reference, in the past 7 days, we've seen the following request counts to supplier frontend endpoints that use manifest data:

> G7: 0
G8: 0
G9: 4
G10: 10
G11: 11
G12: many
DOS: unknown
DOS2: 0
DOS3: 0
DOS4: 0
DOS5: many

This is really interesting data, thank you for collecting it.

lfdebrux:
I think this is a good idea in principle, I would definitely like to see it in content-loader

bjgill:
This experiment was successful, so I've created a PR to implement it properly: https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1368

PR #1353, opened 2021-02-24
bump country and territory autocomplete version

gidsg:
https://trello.com/c/Hey9OS9t/473-mitigate-registers-switch-off
As part of mitigating registers switch off we want to make sure associated dependencies are up to date.


alex9smith (approved):
🗺️  👍

PR #1355, opened 2021-02-24
Bump govuk-country-and-territory-autocomplete to 0.5.0

gidsg:
https://trello.com/c/Hey9OS9t/805-2-mitigate-registers-switch-off

This appears to be the last version which is compatible with the functional tests. 

It appears later versions use different javascript which requires a further interaction after entering a value e.g. click on dropdown value or press return key for the input to be accepted. It is unclear why this was implemented or if it is intentional. I can make an enter keypress to get this to work using Firefox/geckodriver however, I was unable to get it to work using PhantomJS. 

So this PR changes to the latest version as supported by the current tests

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1755,9 +1755,9 @@
       "integrity": "sha1-tf1UIgqivFq1eqtxQMlAdUUDwac="
     },
     "corejs-typeahead": {
-      "version": "1.2.1",
-      "resolved": "https://registry.npmjs.org/corejs-typeahead/-/corejs-typeahead-1.2.1.tgz",
-      "integrity": "sha1-NFqK/mZMxJQHW1m2R3eAfws/Eys=",
+      "version": "1.3.1",
+      "resolved": "https://registry.npmjs.org/corejs-typeahead/-/corejs-typeahead-1.3.1.tgz",
+      "integrity": "sha512-fyNlBNWJNL6EQUnJyAunEzBzRcwR2cEHtZXBi2pndHPOJ/wpOf3wbS+/Oh+kYYS5sKowQcs0LFwMSl6Y2Xeqkw==",
       "requires": {
         "jquery": ">=1.11"
       }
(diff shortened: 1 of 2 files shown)


lfdebrux:
> It appears later versions use different javascript which requires a further interaction after entering a value e.g. click on dropdown value or press return key for the input to be accepted. It is unclear why this was implemented or if it is intentional. I can make an enter keypress to get this to work using Firefox/geckodriver however, I was unable to get it to work using PhantomJS.

Is this possibly related to https://github.com/alphagov/accessible-autocomplete/issues/472

lfdebrux:
What's to stop us updating the functional tests to hit enter?

gidsg:
> What's to stop us updating the functional tests to hit enter?

@lfdebrux I tried to hit enter, i.e. as described in https://stackoverflow.com/questions/10866946/how-do-i-simulate-hitting-enter-in-an-input-field-with-capybara-and-chromedriver
it worked in geckodriver/firefox but not in PhantomJS so I suspect the PhantomJS driver either can't understand the javascript being used or doesn't support the enter command.

lfdebrux:
> > What's to stop us updating the functional tests to hit enter?
> 
> @lfdebrux I tried to hit enter, i.e. as described in https://stackoverflow.com/questions/10866946/how-do-i-simulate-hitting-enter-in-an-input-field-with-capybara-and-chromedriver
> it worked in geckodriver/firefox but not in PhantomJS so I suspect the PhantomJS driver either can't understand the javascript being used or doesn't support the enter command.

Did you try sending the character '\n'? We do that in https://github.com/alphagov/digitalmarketplace-functional-tests/blob/6cacebd4ac4a8dc517ac1dab60ee3bd70b339550/features/smoulder-tests/public/live_reload.feature#L10

gidsg:
> Did you try sending the character \n?

@lfdebrux Yes, I tried that, it didn't work. I suspect it is because either the javascript either isn't understood by PhantomJS or there is some sort of race condition where it makes the input before the javascript has fully initialised. 
screenshot_2021-02-24-14-45-32 979

lfdebrux:
> @lfdebrux Yes, I tried that, it didn't work. I suspect it is because either the javascript either isn't understood by PhantomJS or there is some sort of race condition where it makes the input before the javascript has fully initialised.

Okay cool 👍

lfdebrux (approved):

PR #1366, opened 2021-03-04
Use govukRadios for declaration reuse question

bjgill:
Part of our work to use GOV.UK components instead of our custom ones to improve accessibility.

Switch to use a standard `RadioField`. Previously, we treated it as a checkbox and allowed users to submit without choosing an option. Users are now forced to choose an option, which should reduce the chance of them accidentally not copying a previous declaration and created a lot of unnecessary work.

Also change the page title and header to be the question. This allows us to remove the question as a separate header, which looked a bit odd.

Then add a friendly error message. Make the question text into a question advice hint so that the error link scrolls correctly.


lfdebrux:
### Screenshot (with error messages)

image

lfdebrux:
@NoraGDS how does the content on this page look to you? We've added an error message and moved the question to be the page heading.

app/templates/frameworks/reuse_declaration.html
```diff
@@ -31,21 +29,43 @@
...
         <input type="hidden" name="old_framework_slug" value="{{ old_framework.slug }}"/>
 
-        {{ form.reuse }}
+        {{ govukRadios({
+          "classes": "govuk-radios--inline",
+          "idPrefix": "input-" + form.reuse.name,
+          "name": form.reuse.name,
+          "hint": {
+            "html": question_advice,
+            "classes": "dm-question-advice",
```
domoscargin, line 52: Are we using this anywhere but on Briefs Responses to override the default grey colouring? https://github.com/search?q=dm-question-advice&type=code

domoscargin: (ie: does it do anything here?)

bjgill: Curious. briefs-frontend has [the override](https://github.com/alphagov/digitalmarketplace-brief-responses-frontend/blob/d9fe9616f598e33fd5e86d28bc47e6664ce77985/app/assets/scss/overrides/_hint.scss), but not supplier-frontend.

You're quite right. It has no effect. I removed line 52 and nothing changed. I'll remove it, since it seems a bit pointless.

domoscargin (approved):
LGTM!

app/templates/frameworks/reuse_declaration.html
lfdebrux: 🤔 so why isn't it grey?

domoscargin: I think Description and Question Advice fields get parsed into HTML with govuk classes. So the elements within a complex formatted hint have things like `govuk-body` which set the text colour and override the grey hint text.

It's not ideal.

lfdebrux: I seee.... confusion abounds

domoscargin: Especially since right now hints are contained within a `<span>`, so having block level elements inside that is technically invalid HTML.

govuk-frontend v3 changes this to a `<div>`

lfdebrux: 😬

lfdebrux (changes requested):
After discussing with @NoraGDS I think we should change the page layout back to having the question directly above the input.

I will make the change.

lfdebrux (approved):

lfdebrux:
### Screenshot

The reuse declaration page after this PR looks the same as it did before.

Screenshot 2021-03-09 at 09 35 34

PR #1368, opened 2021-03-04
Load old manifests lazily for faster app startup

bjgill:
We've added [the ability to lazily load manifests](https://github.com/alphagov/digitalmarketplace-content-loader/pull/137). Use this for old frameworks to dramatically reduce startup time. This should only affect a small number of user requests, where they'll be about 1s slower. See the previous experiment for numbers: #1348

Keep the newest frameworks eagerly loaded. This is because I think these frameworks receive substantial traffic. So by eagerly loading we avoid slow user requests that are otherwise highly likely. We also want to load these frameworks during app startup so that the app fails to start if the manifests are malformed. We don't typically change old framework data so they're at much lower risk of being broken.

If this works and doesn't break things, we can roll out these changes to the other frontend apps that load manifests as well.

### DIFF
--- a/app/main/__init__.py
+++ b/app/main/__init__.py
@@ -17,28 +17,48 @@ _local = Local()
 
 
 def _load_dos(master_cl):
-    master_cl.load_manifest('digital-outcomes-and-specialists', 'declaration', 'declaration')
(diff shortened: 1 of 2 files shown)


lfdebrux (approved):
Works for me! 🙌

PR #1369, opened 2021-03-05
Remove uses of secondary action link template

lfdebrux:
Ticket: https://trello.com/c/MBn0LIOB/852-2-remove-secondary-action-link-component-from-supplier-frontend

We don't want to use a template for these simple links.

Also remove some unneeded CSS styles.


domoscargin (approved):
LGTM

PR #1371, opened 2021-03-08
Replace notification banner on 'make a declaration' journey

alex9smith:
Replace the toolkit banner with the [banner component](https://github.com/alphagov/digitalmarketplace-govuk-frontend/tree/master/src/digitalmarketplace/components/banner) from the digitalmarketplace-govuk-frontend.

I have run the functional tests against this change and the [relevant scenario](https://github.com/alphagov/digitalmarketplace-functional-tests/blob/master/features/supplier/supplier_applies_to_a_framework.feature#L42) still passes as we don't check for the banner as part of that test.

https://trello.com/c/CQlT1KmX/853-2-replace-notification-banner-on-make-a-supplier-declaration-journey

Previous banner:
image

New banner:
image

### DIFF
--- a/app/templates/frameworks/edit_declaration_section.html
+++ b/app/templates/frameworks/edit_declaration_section.html
@@ -33,12 +33,9 @@
 
 {% block mainContent %}
   {% if name_of_framework_that_section_has_been_prefilled_from %}
-    {% with
-       message = "Answers on this page are from an earlier declaration and need review.",
-       type = "information"
-    %}
-      {% include "toolkit/notification-banner.html" %}
-    {% endwith %}
+    {{ dmBanner({
+        "title": "Answers on this page are from an earlier declaration and need review.",
+    }) }}
   {% endif %}
 
   <form method="post" enctype="multipart/form-data" class="supplier-declaration" action="#" {# remove any fragment identifier as validation messages are at the top #}>
--- a/app/templates/_base_page.html
+++ b/app/templates/_base_page.html
@@ -15,6 +15,7 @@
 {% from "digitalmarketplace/components/header/macro.njk" import dmHeader %}
 {% from "digitalmarketplace/components/attachment/macro.njk" import dmAttachment %}
 {% from "digitalmarketplace/components/footer/macro.njk" import dmFooter %}
+{% from "digitalmarketplace/components/banner/macro.njk" import dmBanner %}
 
 {% set assetPath = '/suppliers/static' %}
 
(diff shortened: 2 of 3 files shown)


app/templates/frameworks/edit_declaration_section.html
```diff
@@ -33,12 +33,9 @@
 
 {% block mainContent %}
   {% if name_of_framework_that_section_has_been_prefilled_from %}
-    {% with
-       message = "Answers on this page are from an earlier declaration and need review.",
-       type = "information"
```
alex9smith, line 38: Is the `type` important? We didn't keep it in a [similar PR](https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/1056) for the buyer-frontend

gidsg:
PR is probably a bit late for this comment, but did we consider using the new Design System Notification Banner instead? https://design-system.service.gov.uk/components/notification-banner/

alex9smith:
> did we consider using the new Design System Notification Banner instead?

I believe (but am not sure) that we are using the new banner for messages that require a specific action (eg. error summaries) and this banner for those which are more informational (such as this one which doesn't link to an item further down the page). @lfdebrux @domoscargin am I close to being right here?

domoscargin:
We should definitely start replacing our custom banner with the Design System Notification Banner, but we can't at the moment because it requires migration to govuk-frontend v3.10.0+

Ideally at some point we'll replace our custom banner across all apps with the Notification Banner.

app/templates/frameworks/edit_declaration_section.html
domoscargin: No, the banner component doesn't use type. Our alert component uses it for styling success messages green, etc.

lfdebrux:
> Ideally at some point we'll replace our custom banner across all apps with the Notification Banner, but I don't think we've made the call as to whether we should start replacing them individually where possible (the inconsistency could be an issue).

Yeah, like @domoscargin says we knew this is an issue we just haven't considered it properly. When this ticket was written the Design System banner didn't exist :) We're also in a similar situation with our cookie banner and our breadcrumbs.

My feeling is we should avoid being inconsistent, and it's better to focus on migrating all apps before revisiting old work, but I'm open to being persuaded otherwise.

alex9smith:
Good points, not sure how I forget about supplier frontend not being on govuk-frontend v3 yet. My gut feeling is also that consistency is important - we already have enough differences in the various FE packages between our apps that keeping track of intentional inconsistencies sounds like a nightmare.

gidsg (approved):
Cool, changes make sense to me 👍

PR #1373, opened 2021-03-10
Update DUNS number validation for digitalmarketplace-utils 57.0.0

lfdebrux:
Closes https://trello.com/c/52Xz6MJM/2184-users-are-able-to-create-supplier-accounts-with-invalid-duns-numbers

This PR updates the supplier frontend to use digitalmarketplace-utils 57.0.0 with changes in alphagov/digitalmarketplace-utils#603.

This will hopefully fix our smoulder tests so that invalid DUNS numbers cannot be used to create a supplier account.

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -7,6 +7,9 @@ from flask_login import current_user
 from dmapiclient.audit import AuditTypes
 from dmcontent.content_loader import ContentNotFoundError
 from dmutils.dates import update_framework_with_formatted_dates
+from dmutils.direct_plus_client import (
+    DirectPlusError, DUNSNumberNotFound, DUNSNumberInvalid,
+)
 from dmutils.email import send_user_account_email
 from dmutils.email.dm_mailchimp import DMMailChimpClient
 from dmutils.flask import timed_render_template as render_template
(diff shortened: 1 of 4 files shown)


gidsg (approved):
LGTM 🦆 👍

app/main/views/suppliers.py
```diff
@@ -533,16 +536,15 @@ def duns_number():
             # Otherwise the number is good, add the 'primaryName' to the session and continue
             try:
                 organization = direct_plus_client.get_organization_by_duns_number(form.duns_number.data)
-                if organization is not None:
-                    company_name = organization['primaryName']
-            except (KeyError, ValueError):
+                company_name = organization['primaryName']
+            except (DUNSNumberInvalid, DUNSNumberNotFound):
+                form.duns_number.errors = ["DUNS number not found"]
+            except (DirectPlusError, KeyError, ValueError):
```
bjgill, line 542:
```suggestion
            except (DirectPlusError, KeyError):
```

Is there anything able to raise a `ValueError` any more?

bjgill (commented):
Looks good - the logic is now much clearer. Just one question about exceptions.

app/main/views/suppliers.py
lfdebrux: Yes, `get_organization_by_duns_number()` can still raise a `ValueError` when calling `response.json()`

https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/direct_plus_client.py#L132

I guess we could propagate it all the way to a 500 if we wanted to though, by removing ValueError.

bjgill: Very true - since it wasn't in the doc comment, I missed that it could still throw `KeyError` and `ValueError`.

In that case, I think the behaviour here is correct. It's more that I would like the Direct+ client to either document those exceptions or turn them into `DirectPlusError`s. Sorry - missed that when reviewing the previous PR.

bjgill (approved):
:rocket:

PR #1374, opened 2021-03-11
Return 200 response to GET /suppliers/mailing-list

bjgill:
Currently, the status code for a GET request to /suppliers/mailing-list is always 400. Originally, it returned a 200. However, during refactoring (450bca893bb53401945c0401fd2c0e1bb5f343d6), it was changed to a 400. I can see no justification for the change, and suspect it was a mistake.

Change back to a 200 for GETs, since this seems more correct.


lfdebrux:
I agree, this looks like a mistake. Good catch.

lfdebrux (approved):

PR #1375, opened 2021-03-12
Replace declaration overview summary table with govukSummaryList

domoscargin:
https://trello.com/c/IzTERgai/857-2-replace-your-declaration-overview-summary-table-with-govuk-summary-list

## Changes:
* Summary table replaced with summary list
  * Each entry has an individual "Edit" link
  * Other behaviour stays the same

## Issues:
* All those Edit links look a bit busy.
* An existing issue: we use URI fragments to anchor to a specific question in the links from this summary list, but there's no other confirmation of where the page is focused - ideally the question we linked to would be highlighted on the next page.

### Old summary table
Old summary table, showing some sections filled in, and the lack of "Edit" actions next to individual row items

### New summary list
New summary list, showing some sections filled in, and "Edit" links for both the overall section, and individual entries within that section.

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -45,7 +45,7 @@ from ..helpers.frameworks import (
     get_statuses_for_lot,
     get_supplier_framework_info,
     get_supplier_on_framework_from_info,
-    get_supplier_registered_name_from_declaration,
+    get_supplier_registered_name_from_declaration, question_references,
     register_interest_in_framework,
     return_supplier_framework_info_if_on_framework_or_abort,
     returned_agreement_email_recipients,
@@ -548,6 +548,13 @@ def framework_supplier_declaration_overview(framework_slug):
                     ) if framework["status"] == "open" and sections_errors else None
                 )
             )
+
+        # We need to parse key text for question references. This isn't optimal, but references only apply to
+        # this app, so holding off on centralising this logic to the Content Loader.
+        def label_question_reference(row):
+            row["key"]["text"] = question_references(row.get("key", {}).get("text", ""), section.get_question)
+            return row
+        section.summary_list = list(map(label_question_reference, section.summary_list))
 
     return render_template(
         "frameworks/declaration_overview.html",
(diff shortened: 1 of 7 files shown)


domoscargin:
Converting to draft as I've just noticed an issue with the table generating some question references

lfdebrux:
### My review checklist

I've:

- [ ] looked at the code
- [ ] tested locally

app/main/views/frameworks.py
```diff
@@ -529,6 +530,32 @@ def framework_supplier_declaration_overview(framework_slug):
...
+                    ) if framework["status"] == "open" and sections_errors else None
+                )
+            )
+
+            # We need to parse key text for question references. This isn't optimal, but references only apply to
+            # this app, so holding off on centralising this logic to the Content Loader.
+            def label_question_reference(row):
+                row["key"]["text"] = question_references(row.get("key", {}).get("text", ""), section.get_question)
+                return row
+            section.summary_list = list(map(label_question_reference, section.summary_list))
```
lfdebrux, line 557: Currently this iterates through the list every time a question is added, did you mean to dedent this one level so it runs once after the for loop?

app/templates/frameworks/declaration_overview.html
```diff
@@ -72,56 +72,28 @@ <h1 class="govuk-heading-l">Your declaration overview</h1>
...
+          {{ section.name }}
+          {% if section.editable %}
+            {% if framework.status == 'open' %}
+              <span class="dm-section-action-link">
+                <a class="govuk-link" href="{{ url_for('.framework_supplier_declaration_edit', framework_slug=framework.slug, section_id=section.id)}}">
+                  Edit
+                  <span class="govuk-visually-hidden">{{ section.name }}</span>
+                </a>
+              </span>
+            {% endif %}
```
lfdebrux, line 87: Interesting, what's the advantage of putting the edit link in the heading like this?

app/main/views/frameworks.py
domoscargin: Oof. Well, the good news is that it was loading speedily even though I hadn't caught this, so now it should be even quicker...

Thanks for spotting.

app/templates/frameworks/declaration_overview.html
domoscargin: Oh gosh, none, other than to allow quick prototyping while developing and then miss when pushing. Sorry about this, will tweak now.

lfdebrux (approved):

lfdebrux:
Looks like a good improvement to me, we can deal with the other issues later!

PR #1395, opened 2021-03-25
Replace text boxes on confirm company details journey

domoscargin:
This replaces our Frontend Toolkit text fields with GOV.UK Design System equivalents for the Confirm Company Details journey.

https://trello.com/c/gQM1Rikz/895-2-replace-text-boxes-with-govuk-frontend-text-input-component-in-confirm-company-details-journey

https://trello.com/c/duej3OKP/894-1-replace-large-text-boxes-with-govuk-frontend-charactercount-component-in-confirm-company-details-journey

PR #1396, opened 2021-03-26
Replace radio on supplier trading status page

bjgill:

We've added a helper method in `CompanyTradingStatusForm` to get the items in the right format for `govukRadios`. We're not sure this is the right way to do this - we'll refactor as necessary when we come to replace the other radios in the supplier details journey.

### DIFF
--- a/app/main/forms/suppliers.py
+++ b/app/main/forms/suppliers.py
@@ -252,8 +252,16 @@ class CompanyTradingStatusForm(FlaskForm):
         },
     ]
 
-    trading_status = DMRadioField(
+    trading_status = RadioField(
         "What’s your trading status?",
-        hint="This information will be used to find out about the types of companies on frameworks.",
         validators=[InputRequired(message="You must choose a trading status.")],
-        options=OPTIONS)
+        choices=[(option["value"], option["label"]) for option in OPTIONS],
+        id="input-trading_status-1"  # TODO: change to input-reuse when on govuk-frontend~3
+    )
+
+    def get_items(self):
+        return [
+            {'value': value, 'text': text, 'checked': checked}
+            for (value, text, checked)
+            in self.trading_status.iter_choices()
+        ]
--- a/app/templates/suppliers/edit_supplier_trading_status.html
+++ b/app/templates/suppliers/edit_supplier_trading_status.html
@@ -1,7 +1,9 @@
 {% extends "_base_page.html" %}
 
+{% set page_name = "Trading status" %}
+
 {% block pageTitle %}
-  Trading status – Digital Marketplace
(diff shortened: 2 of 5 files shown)


app/main/forms/suppliers.py
```diff
@@ -252,8 +252,16 @@ class CompanyTradingStatusForm(FlaskForm):
...
+        choices=[(option["value"], option["label"]) for option in OPTIONS],
+        id="input-trading_status-1"  # TODO: change to input-reuse when on govuk-frontend~3
+    )
+
+    def get_items(self):
+        return [
+            {'value': value, 'text': text, 'checked': checked}
+            for (value, text, checked)
+            in self.trading_status.iter_choices()
+        ]
```
lfdebrux, line 267: In other places (for instance [briefs frontend](https://github.com/alphagov/digitalmarketplace-briefs-frontend/blob/master/app/main/forms/award_or_cancel.py)) we've added the items as an attribute on the field in the form constructor, this is perfectly reasonable too though.

alex9smith:
Rebased to bring in @domoscargin 's (much neater) test updates from #1395

app/templates/suppliers/edit_supplier_trading_status.html
```diff
@@ -31,12 +33,27 @@
 <div class="single-question-page">
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
-      <h1 class="govuk-heading-l">What&rsquo;s your trading status?</h1>
-
       <form method="POST" action="{{ url_for('.edit_supplier_trading_status') }}">
```
domoscargin, line 36: We should add `novalidate` to forms to avoid issues with browser-specific validation. https://design-system.service.gov.uk/patterns/validation/#turn-off-html5-validation

app/templates/suppliers/edit_supplier_trading_status.html
```diff
@@ -31,12 +33,27 @@
 <div class="single-question-page">
```
domoscargin, line 33: We should be able to get rid of this wrapping div now that everything on this page is using `govuk-frontend` - this is an old class for spacing our frontend toolkit pages.

It may affect spacing a bit, but that's expected.

app/main/forms/suppliers.py
bjgill: I think the difference between this and your example is that we need to handle `checked`, which tracks which radio choice is pre-checked. As far as Alex and I could tell, `iter_choices` is the only thing that gives us that information, hence this helper method being necessary.

We're planning to revisit when we come to do the other two radios in this journey, when it should be clearer what the right way to do this is.

domoscargin (approved):
LGTM

bjgill:
Thanks. Spacing all looks sensible, so I'm going to merge

image

PR #1398, opened 2021-03-29
Update content for create a supplier account journey

lfdebrux:
Ticket: https://trello.com/c/Ajssa0oq/890-3-update-content-for-create-a-supplier-account-journey

### DIFF
--- a/app/templates/suppliers/create_your_account_complete.html
+++ b/app/templates/suppliers/create_your_account_complete.html
@@ -31,11 +31,9 @@
     <div class="govuk-grid-column-two-thirds">
       <h1 class="govuk-heading-l">Activate your account</h1>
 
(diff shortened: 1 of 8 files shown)


app/templates/suppliers/create_your_account_complete.html
```diff
@@ -32,13 +32,8 @@
       <h1 class="govuk-heading-l">Activate your account</h1>
 
       <div class="dmspeak">
```
domoscargin, line 34: Can we get rid of this while we're here or does it look out of place with other pages on the journey?

lfdebrux: Oh I missed that, I will get rid of it

domoscargin (approved):
👌

PR #1400, opened 2021-03-30
Remove unused template imports

lfdebrux:
Noticed these while trying to list all outstanding frontend changes in this app.

### DIFF
--- a/app/templates/suppliers/create_your_account.html
+++ b/app/templates/suppliers/create_your_account.html
@@ -1,5 +1,4 @@
 {% extends "_base_page.html" %}
-{% import "toolkit/summary-table.html" as summary %}
 
 {% block pageTitle %}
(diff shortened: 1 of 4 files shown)


app/templates/suppliers/create_your_account.html
```diff
@@ -1,5 +1,4 @@
 {% extends "_base_page.html" %}
-{% import "toolkit/summary-table.html" as summary %}
```
domoscargin, line 2: Weird, I'm not sure this page ever used this import.

domoscargin (approved):
LGTM - thanks for catching

app/templates/suppliers/create_your_account.html
lfdebrux: It is weird

PR #1401, opened 2021-03-30
Replace radio on organisation size page 

bjgill:

Follows the example of #1396

---
Old:

image

New:

image


bjgill:
Had a chat on Slack with @lfdebrux, who pointed out that the `govuk_options` helper might be what I want. From quick experimentation, it does indeed seem to work. I'll add a new commit to switch across.

bjgill:
And that's done. The tests pass for me locally, and I've updated the PR description with a screenshot.

PR #1403, opened 2021-03-31
Ask for company registration number with govukRadios 

bjgill:

As part of moving to Design System 3, we need to replace our use of toolkit/forms/selection-buttons.html with govukRadios from the design system. This should make this page more accessible.

This question is quite complex, so the template is quite lengthy. However, quite a bit of that is just caused by the govukRadios being a bit more verbose and having more braces.

Similar to #1396 and #1401, though this question doesn't use `DMRadioField`, so this PR only needs to make template changes.

The overall diff is not very helpful - recommend reviewing commit by commit instead.

---

Old:
image

New:
image

### DIFF
--- a/app/templates/suppliers/edit_company_registration_number.html
+++ b/app/templates/suppliers/edit_company_registration_number.html
@@ -31,42 +31,71 @@
 <div class="single-question-page">
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
-      <h1 class="govuk-heading-l">Are you registered with Companies House?</h1>
-
-      <form method="POST" action="{{ url_for('.edit_supplier_registration_number') }}">
+      <form method="POST" action="{{ url_for('.edit_supplier_registration_number') }}" novalidate>
         <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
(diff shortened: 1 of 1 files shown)


app/templates/suppliers/edit_company_registration_number.html
```diff
@@ -20,61 +22,88 @@
...
+
+      {% set companiesHouseNumberHtml %}
+      {{ govukInput({
+        "id": form.companies_house_number.id,
+        "name": form.companies_house_number.name,
+        "label": {
+          "text": "Companies House number"
+        },
+        "hint": {
+          "html": "<a class='govuk-link' href='https://beta.companieshouse.gov.uk/search/companies'>Find your 8 digit Companies House number</a>"|safe,
```
lfdebrux, line 46: I realise this is a new feature, but I think this link should probably open in a new tab

lfdebrux (approved):
Looks good to me! 👍

app/templates/suppliers/edit_company_registration_number.html
bjgill: I'll put that in a tiny follow-up PR

PR #1407, opened 2021-04-07
Remove unused WTForms forms

lfdebrux:
Found this while populating the frontend backlog.


gidsg (approved):
Good spot 🕵️ 
We must have missed this when doing clean-up post-esignature.
👍

PR #1409, opened 2021-04-07
Add support for localstack S3 subdomains

bjgill:
Trello: https://trello.com/c/XoWXM8lC/85-run-dmrunner-without-credentials

This matches what we use for the real S3 - the old way of doing it in the path is deprecated.

By using localstack S3 subdomains, we now get file upload working correctly. This is because we no longer lose the bucket name when we add the document path, which was the previous behaviour.

This affects supplier-frontend only when run locally in developer mode. It depends on having a recent version of localstack - see https://github.com/alphagov/digitalmarketplace-runner/pull/194. This change means that the functional tests tagged as '@file-upload' now succeed when run locally.

Also, an unrelated small secret baseline update.

### DIFF
--- a/config.py
+++ b/config.py
@@ -139,7 +139,7 @@ class Development(Config):
     DM_API_AUTH_TOKEN = "myToken"
 
     DM_S3_ENDPOINT_URL = (  # use envvars to set this, defaults to using AWS
-        f"http://localhost:{os.environ['DM_S3_ENDPOINT_PORT']}"
+        f"http://s3.localhost.localstack.cloud:{os.environ['DM_S3_ENDPOINT_PORT']}"
         if os.getenv("DM_S3_ENDPOINT_PORT") else None
     )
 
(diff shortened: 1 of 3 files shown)


bjgill:
Once this is merged, I'll apply the same change to admin-frontend, which is the other app that does this.

config.py
```diff
@@ -139,7 +139,7 @@ class Development(Config):
     DM_API_AUTH_TOKEN = "myToken"
 
     DM_S3_ENDPOINT_URL = (  # use envvars to set this, defaults to using AWS
-        f"http://localhost:{os.environ['DM_S3_ENDPOINT_PORT']}"
+        f"http://s3.localhost.localstack.cloud:{os.environ['DM_S3_ENDPOINT_PORT']}"
         if os.getenv("DM_S3_ENDPOINT_PORT") else None
```
lfdebrux, line 143: I think this is only used to set DM_ASSETS_URL, so I'm not sure why the two are different in this commit?

bjgill: The S3 client in dmutils peaks at it: https://github.com/alphagov/digitalmarketplace-utils/pull/580/files

Though you're right that this is surprising behaviour. I'll take a look and see if I can move the logic into supplier-frontend

lfdebrux: Ah good point 🤦

bjgill: Updated. Once we've added this logic here and to admin frontend, we can remove it from dmutils.

lfdebrux (approved):

PR #1410, opened 2021-04-08
change content loader init to use more inclusive language

gidsg:
https://trello.com/c/DrVWMkcZ/2169-rename-default-branches-on-our-repos-from-master-to-main


lfdebrux (approved):

bjgill (approved):
Good catch

PR #1286, opened 2020-11-30
Add attachment component to Frameworks dashboard and updates pages

domoscargin:
https://trello.com/c/tUd7Ck2w/275-5-replace-document-links-with-attachment-component-in-frontend-apps

## What
This PR uses the new Attachment component to display document downloads on these two pages:

`suppliers/frameworks/<framework>`
`suppliers/frameworks/<framework>/updates`

## Why
We currently don't offer a way to acquire more accessible versions of our documents. This makes it easier for users to request them, and provides a way to show more metadata about files once we have a way to extract that information on the backend.

## Issues
* We don't currently provide much metadata beyond file type and last modified date
* On the updates page, we're currently using the file extension to pass in type, when we should ideally be using MIME types like `application/pdf`
* Using a Details component (for the contact details) within a table creates all sorts of issues, so I've had to move the directions for accessible documents out of the tables - this at least makes the table cells less repetitive
* It's not the most intuitive task in the world to force these pages into different states to inspect changes - perhaps we can improve this
* There are still a few digitalmarketplace-frontend-toolkit elements on this page - these felt out of scope, but I did replace the summary-table as part of this PR

### Before
Screenshot 2020-11-30 at 17 34 37
### After
Screenshot 2020-11-30 at 17 23 24

### Before
Screenshot 2020-11-30 at 17 34 06
## After
Screenshot 2020-11-30 at 17 24 40

PR #1427, opened 2021-05-27
Fix tests for updates table

bjgill:
Trello: https://trello.com/c/tnbrDQrf/960-1-investigate-broken-tests-in-supplier-frontend-1-day

Previously, this test was broken in two ways:
1. The CSS styles used in the xpaths were no longer in use due to https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1286/
2. The test didn't fail when the xpath search found no matching tables.

So I needed to fix this by updating the xpaths to match the current naming. I then updated the tests to fail if it ever fails to find the two tables - for clarifications and communications.

The two affected tests seemed very similar, so I consolidated them into one.

### DIFF
--- a/tests/app/main/test_frameworks.py
+++ b/tests/app/main/test_frameworks.py
@@ -3572,10 +3572,10 @@ class TestFrameworkUpdatesPage(BaseApplicationTest):
         doc = html.fromstring(response.get_data(as_text=True))
         self._assert_page_title_and_table_headings(doc)
 
-        [communications_table, clarifications_table] = doc.xpath('//table[contains(@class, "govuk-table")]')
+        [communications_table, clarifications_table] = doc.xpath('//table[@data-name="updates-documents"]')
 
         def assert_updates_table_contains_files(table, files):
-            item_rows = table.findall('.//tr[@class="govuk-table__row"]')
+            item_rows = table.findall('.//tr')
             assert len(item_rows) == len(files)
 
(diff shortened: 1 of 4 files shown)


tests/app/main/test_frameworks.py
```diff
@@ -3608,17 +3572,15 @@ def test_names_with_the_section_name_in_them_will_display_correctly(self, s3):
         doc = html.fromstring(response.get_data(as_text=True))
         self._assert_page_title_and_table_headings(doc)
 
-        tables = doc.xpath('//div[contains(@class, "updates-document-tables")]/table')
+        [communications_table, clarifications_table] = doc.xpath('//table[contains(@class, "govuk-table")]')
```
gidsg, line 3575: I guess this may be a bit brittle as it will find the first table on the page, so maybe better to add a `data-name` or similar specific attribute that would be unique to this table

gidsg (approved):
Looks good (aside from minor optional comment)

tests/app/main/test_frameworks.py
bjgill: I made the xpath more specific, so it should now be less brittle.

bjgill: Actually, looking at it some more, I think a `data-name` is more elegant

gidsg (approved):
💯 👍

PR #1428, opened 2021-05-28
Create a service journey: replace "alerts with buttons" with "confirm action" page pattern

domoscargin:
https://trello.com/c/4qVEhIoC/985-3-replace-alerts-with-buttons-with-confirm-action-page-pattern-in-create-a-service-journey

## What

We want to replace uses of banners with buttons with a page to confirm the action, similar to the GOV.UK Frontend [confirm delete page pattern](http://govuk-frontend-review.herokuapp.com/full-page-examples/confirm-delete).

## Issues
Functional tests pass for this change (both in the "live" state when the supplier application tests are ignored, and the supplier application tests themselves in the "open" state), but I'm reviewing them to see if they accurately account for these new journeys.

## How to see the new pattern in action
1. Open at least one framework, log in as a supplier and begin the framework application.
2. On the framework application dashboard, click "add, edit and complete services"
3. Follow the steps below for each journey

### Remove a draft service
1. If you do not have any draft services, create one.
2. Click "Edit" next a draft service.
3. On the draft service page, scroll down and click "Remove draft service"
Screenshot 2021-05-26 at 11 29 33
4. You'll be redirected to the new confirmation page
Screenshot 2021-05-28 at 11 10 49

### Remove a service subsection
1. If you do not have any draft services, create one (Digital Specialists is a simple lot to use)
2. Click "Edit" next to a draft service.
3. On the draft service page, click "Add" next to a section and save data.
4. On the draft service page, click "Remove" next to the section
Screenshot 2021-05-27 at 14 36 56
5. You'll be redirected to the new confirmation page
Screenshot 2021-05-28 at 11 11 04

### Copy all lot services from a previous framework
1. Add a service in a lot the supplier has previously submitted to a framework (this only applies to multi-service lots, such as the G-Cloud lots)
2. On the services page, click "View and add your services from [previous framework]"
Screenshot 2021-05-28 at 15 25 10
3. Click "Add all your services"
Screenshot 2021-05-28 at 15 26 05
4. You'll be redirected to the new confirmation page
Screenshot 2021-05-28 at 11 08 55

PR #1429, opened 2021-05-28
Remove redundant checks and API requests

bjgill:
The `@EnsureApplicationCompanyDetailsHaveBeenConfirmed` annotation already ensures that suppliers have registered interest in a framework and returns a 404 status if not.

So the `if not get_supplier_framework_info(data_api_client, framework_slug):` checks are entirely redundant, since it is imossible for them ever to fail. So we can safely remove them all to remove some unnecessary API requests.

Add a test for this behaviour to the annotation tests. This means that we can safely remove the tests for this behaviour from the individual endpoints.

The history of this is that the annotation was added later (c29a4ee2597b81ce8062b385105479efb5379721). It looks as if they didn't notice that they'd made this check redundant at the time.

Discovered during https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1428#discussion_r641648384

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -27,7 +27,6 @@ from ..helpers.services import (
 from ..helpers.frameworks import (
     get_framework_and_lot_or_404,
     get_declaration_status,
-    get_supplier_framework_info,
     get_framework_or_404,
     get_framework_or_500,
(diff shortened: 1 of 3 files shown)


alex9smith (approved):
Good spot!

PR #1435, opened 2021-06-02
Commonise getting draft service

bjgill:
There are a number of endpoints that all need to get a draft service in the same way. So it makes sense to create a helper to cut down on unnecessary duplication.


alex9smith (approved):
Nice!

PR #1437, opened 2021-06-03
Replace radio in service reuse page

bjgill:
https://trello.com/c/JiuMuQNd/986-1-replace-radios-in-reuse-service-page

I made a change in behaviour. Previously the page would default to 'no' if you didn't select a response. I couldn't work out how to keep this behaviour with govukradios, and couldn't see any other pages that did this. So it's now mandatory to provide a response on this page, in common with all other similar pages.

Old:

image

New:

image

### DIFF
--- a/app/main/forms/frameworks.py
+++ b/app/main/forms/frameworks.py
@@ -3,7 +3,6 @@ from wtforms import HiddenField, RadioField
 from wtforms.validators import DataRequired, Length, InputRequired
 
 from dmutils.forms.fields import DMBooleanField, DMStripWhitespaceStringField, DMRadioField
-from dmutils.forms.widgets import DMSelectionButtonBase
 
 
 class AcceptAgreementVariationForm(FlaskForm):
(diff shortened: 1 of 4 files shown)


app/main/forms/frameworks.py
```diff
@@ -30,21 +29,21 @@ class ReuseDeclarationForm(FlaskForm):
...
 class OneServiceLimitCopyServiceForm(FlaskForm):
-
-    copy_service = DMBooleanField(
-        "Do you want to reuse your previous {lot_name} service?",
-        false_values=("False", "false", ""),
-        widget=DMSelectionButtonBase(type="boolean"),
+    copy_service = RadioField(
+        id="input-copy_service-1",  # TODO: change to input-copy_service when on govuk-frontend~3
+        choices=[("yes", "Yes"), ("no", "No")],
+        validators=[InputRequired("Select yes if you want to reuse your previous service")],
```
domoscargin, line 35: Odd that it wasn't compulsory to select a response on a radio. Definitely better this way.

domoscargin (approved):
LGTM

PR #1438, opened 2021-06-04
Fix previous services form attributes typo

domoscargin:
Spotted a typo in #1437 where the `novalidate` attribute was enclosed in the `action` string - this fixes it.

### DIFF
--- a/app/templates/services/previous_services.html
+++ b/app/templates/services/previous_services.html
@@ -43,7 +43,7 @@
 
     <div class="govuk-grid-row">
       <div class="govuk-grid-column-two-thirds">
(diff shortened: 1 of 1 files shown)


gidsg (approved):

bjgill:
Good catch - thanks! Presumably this is something the functional tests would have caught when we opened DOS6?

domoscargin:
> Good catch - thanks! Presumably this is something the functional tests would have caught when we opened DOS6?

Yes, the `supplier_applies_to_a_framework` feature would've failed on this.

bjgill:
Thanks - I'll add this as another piece of evidence for why we need the ability to run the functional tests against all framework states

PR #1439, opened 2021-06-04
Use govukTable for list of supplier services

bjgill:
https://trello.com/c/xcKuSRij/984-2-replace-summary-tables-of-services

We were previously using toolkit/summary-table.html. We want to replace all uses of the toolkit with design system components. This is part of our migration to Design System 3 to improve accessibility.

No changes in behaviour. Some formatting changes, but none that look major.

Old:

image

New (the first service has a funny name - I was testing that it was still being escaped properly):

image

### DIFF
--- a/app/templates/services/list_services.html
+++ b/app/templates/services/list_services.html
@@ -1,5 +1,4 @@
 {% extends "_base_page.html" %}
-{% import "toolkit/summary-table.html" as summary %}
 
 {% block pageTitle %}
   Current services – Digital Marketplace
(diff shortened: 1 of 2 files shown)


app/templates/services/list_services.html
```diff
@@ -30,35 +29,34 @@ <h1 class="govuk-heading-l">Your {{ framework.name }} services</h1>
...
-        {% endif %}
-      {% endcall %}
-    {% endcall %}
-  {% endcall %}
+    {{ govukTable({
+      'head': [
+        {'text': "Name", "classes": "govuk-!-width-one-half "},
+        {'text': "Lot"},
+        {
+          'text': "",
```
bjgill, line 53: Here, I'm just copying what came before. Should we replace this with `"Status"` instead for accessibility?

domoscargin: Good call. "Lot" is not necessarily a great  heading either - I got the sense that people who regularly work with DOS would know that really well, but other users might not. We decided to keep "Lot" when this came up last.

domoscargin (approved):
This looks good to me. Good catch on the unicode stuff.

PR #1442, opened 2021-06-08
Fix list of live services for DOS

bjgill:
In https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1439, I accidentally removed some logic that is needed for the DOS version of this table to render correctly. This is because some DOS lots are limited to one service, so the service has no name.

Fix the logic to match the previous logic, and then add a test for this scenario

image


domoscargin (approved):
🎉  This works for me and passes local tests.

PR #1448, opened 2021-06-14
Switch to sourcemaps from gulp

bjgill:
https://trello.com/c/zjkM8Utr/8-3-upgrade-off-gulp-sourcemaps-v3

As of Gulp 4, sourcemap support is built into Gulp. Switch over to this new method (https://gulpjs.com/docs/en/api/src#sourcemaps).

This will help us remove a dependency on gulp-sourcemaps, which is pinning a version of postcss with a known vulnerability.

I've tested this locally, and it doesn't seem to break anything - the sourcemap is still present in the same location.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -6,7 +6,6 @@ const filelog = require('gulp-filelog')
 const include = require('gulp-include')
 const colours = require('colors/safe')
 const path = require('path')
-const sourcemaps = require('gulp-sourcemaps')
 
 // Paths
 let environment = 'development'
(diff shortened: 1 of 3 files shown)


gidsg:
Seems like we need to remove `gulp-sourcemaps` from dependencies and delete this line: https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/0ff7d1689b59fc447387b82f85f9acc308279b17/gulpfile.js#L9

alex9smith (approved):
LGTM

PR #1451, opened 2021-06-17
Replace summary table on draft services page

gidsg:
Replace toolkit summary table on draft services page with govuk-frontend v3 elements.
https://trello.com/c/xcKuSRij/24-2-replace-summary-tables-of-services

### DIFF
--- a/app/templates/frameworks/services.html
+++ b/app/templates/frameworks/services.html
@@ -1,5 +1,4 @@
 {% extends "_base_page.html" %}
-{% import "toolkit/summary-table.html" as summary %}
 {% import "macros/submission.html" as submission %}
 
 {% block pageTitle %}
@@ -39,7 +38,10 @@
   {% endif %}
 
   <h1 class="govuk-heading-l">{{ lot.name }} services</h1>
-
+{{ govukButton({
+      "text": "Add a service",
+      "href": url_for(".start_new_draft_service", framework_slug=framework.slug, lot_slug=lot.slug)
+ }) }}
   {% if framework.status == 'pending' %}
     <div class="summary-item-lede">
       <div class="govuk-grid-row">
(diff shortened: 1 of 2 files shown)


app/templates/frameworks/services.html
```diff
@@ -65,91 +64,115 @@ <h2 class="summary-item-heading">{{ framework.name }} is closed for applications
...
+</p>
 
+    {% if not drafts %}
+      {{"You haven’t added any services yet." if framework.status == 'open' else "You didn’t add any services."}}
+    {% endif %}
+    {{ govukTable({
+      'head': [
+        {'text': "Service name"},
+        {'text': "Progress"},
+        {'text': "Make a copy"},
```
gidsg, line 105: These used to be visually hidden, not sure if we want to replicate that?

app/templates/frameworks/services.html
```diff
@@ -65,91 +64,115 @@ <h2 class="summary-item-heading">{{ framework.name }} is closed for applications
...
+        {% set row = ns.completed_rows.append(
+        [
+          {"html": service_link},
+          {"html": make_copy_button}
+        ]
+    ) %}
+        {{ govukTable({
+      'head': [
+        {'text': "Service name"},
+        {'text': "Make a copy"}
```
gidsg, line 158: Because there are only two items in this table it does not visually align with the draft table as before.
## Before
Screenshot 2021-06-17 at 15 32 39

## After
Screenshot 2021-06-17 at 15 31 41

app/templates/frameworks/services.html
```diff
@@ -65,91 +64,115 @@ <h2 class="summary-item-heading">{{ framework.name }} is closed for applications
...
-    {% endcall %}
-  {% endcall %}
+          submission.unanswered_optional_text(draft.unanswered_required, draft.unanswered_optional))},
+          {"html": make_copy_button},
+      ]
+    ) %}
+{% endfor %}
+<p class="summary-item-top-level-action">
+  <a href="{{ url_for(".start_new_draft_service", framework_slug=framework.slug, lot_slug=lot.slug) }}" class="summary-change-link">Add a service</a>
+</p>
```
gidsg, line 96: I just copied this markup to replicate what the previous macro was doing, not sure if we want to keep this or replace with something else like a button?

app/templates/frameworks/services.html
gidsg: ## Before 
Screenshot 2021-06-17 at 15 32 39
## After
Screenshot 2021-06-17 at 15 38 38

app/templates/frameworks/services.html
```diff
@@ -65,91 +64,115 @@ <h2 class="summary-item-heading">{{ framework.name }} is closed for applications
...
+    }) }}
+    {% endfor %}
+    {% elif framework.status == 'pending' or 'stanstill' %}
+    <ul class="govuk-list">
+        {% for complete_draft in complete_drafts  %}
+            <li>
+                <a class="govuk-link" href="{{ url_for('.view_service_submission', framework_slug=framework.slug, lot_slug=complete_draft.lot, service_id=draft.id) }}">{{ complete_draft.serviceName }}</a>
+            </li>
+        {% endfor %}
+      </ul>
```
gidsg, line 171: I have replaced the table in the pending/standstill state with a simple list as there is only one row. Not sure if we want to replicate the previous style?

## Before
Screenshot 2021-06-17 at 15 48 38


## After
Screenshot 2021-06-17 at 15 47 37

app/templates/frameworks/services.html
domoscargin: The last time we had a table like this, we hid "Make a copy", since it's repeated by all the buttons: https://github.com/alphagov/digitalmarketplace-briefs-frontend/blob/afedd090199322b0dfe1545b824e10e03ae7899e/app/templates/buyers/dashboard.html#L41-L44

app/templates/frameworks/services.html
```diff
@@ -65,91 +64,115 @@ <h2 class="summary-item-heading">{{ framework.name }} is closed for applications
...
+      ]
+    ) %}
+{% endfor %}
+<p class="summary-item-top-level-action">
+  <a href="{{ url_for(".start_new_draft_service", framework_slug=framework.slug, lot_slug=lot.slug) }}" class="summary-change-link">Add a service</a>
+</p>
 
+    {% if not drafts %}
+      {{"You haven’t added any services yet." if framework.status == 'open' else "You didn’t add any services."}}
+    {% endif %}
```
gidsg, line 100: This text used to appear in the table, I've moved it to plain text above the table. 

## Before 
Screenshot 2021-06-17 at 15 52 01

## After
Screenshot 2021-06-17 at 15 52 36

app/templates/frameworks/services.html
```diff
@@ -65,91 +64,115 @@ <h2 class="summary-item-heading">{{ framework.name }} is closed for applications
...
+   {% if framework.status == 'open' %}
+    {% for complete_draft in complete_drafts %}
+    {% set service_link %}
+    <a class="govuk-link" href="{{ url_for('.view_service_submission', framework_slug=framework.slug, lot_slug=draft.lot, service_id=draft.id) }}">{{ complete_draft.serviceName }}</a>
+    {% endset %}
+    {% set make_copy_button %}
+    <form method="post" action="{{ url_for('.copy_draft_service', framework_slug=framework.slug, lot_slug=complete_draft.lot, service_id=complete_draft.id) }}">
+        <input type="hidden" name="csrf_token" value="{{ csrf_token_value or csrf_token() }}" />
+        {{ govukButton({
+          "text": "Make a copy",
```
domoscargin, line 144:
```suggestion
          "html": 'Make a copy <span class="govuk-visually-hidden">of {{ complete_draft.serviceName }}</span>',
```

This will help screenreaders not just reel off "Make a copy" repeatedly.

gidsg:
@domoscargin I've added notes and screenshots of the design changes I've spotted, could you review them?

Note: You'll need to change framework state locally to see all the views.

app/templates/frameworks/services.html
domoscargin: How do we feel about right-aligning the buttons?

We've done that previously with a class: https://github.com/alphagov/digitalmarketplace-briefs-frontend/blob/afedd090199322b0dfe1545b824e10e03ae7899e/app/templates/buyers/dashboard.html#L61

app/templates/frameworks/services.html
domoscargin: I think it makes sense to have a button to add services, since when you're on this page in the right framework state, that's your main action.

Depends on whether we think we should tie that action to the Drafts Services table - I don't think we need to.

app/templates/frameworks/services.html
domoscargin: I think we should jettison the table in this case as well - just keep the plain text message, since there's no point displaying an empty table.

app/templates/frameworks/services.html
```diff
@@ -65,91 +64,115 @@ <h2 class="summary-item-heading">{{ framework.name }} is closed for applications
...
+        {{ govukTable({
+      'head': [
+        {'text': "Service name"},
+        {'text': "Make a copy"}
+      ],
+      'rows': ns.completed_rows,
+      "classes": "dm-table govuk-!-margin-bottom-6",
+    }) }}
+    {% endfor %}
+    {% elif framework.status == 'pending' or 'stanstill' %}
```
domoscargin, line 164:
```suggestion
    {% elif framework.status == 'pending' or 'standstill' %}
```

app/templates/frameworks/services.html
```diff
@@ -65,91 +64,115 @@ <h2 class="summary-item-heading">{{ framework.name }} is closed for applications
...
+  {% set ns = namespace(draft_rows = []) %}
+  {% for draft in drafts %}
+    {% set service_link %}
+    <a class="govuk-link" href="{{ url_for('.view_service_submission', framework_slug=framework.slug, lot_slug=draft.lot, service_id=draft.id) }}">{{ draft.serviceName }}</a>
+    {% endset %}
+    {% set make_copy_button %}
+    <form method="post" action="{{ url_for('.copy_draft_service', framework_slug=framework.slug, lot_slug=draft.lot, service_id=draft.id) }}">
+      <input type="hidden" name="csrf_token" value="{{ csrf_token_value or csrf_token() }}" />
+    {{ govukButton({
+      "text": "Make a copy",
```
domoscargin, line 78:
```suggestion
      "html": 'Make a copy <span class="govuk-visually-hidden">of {{ draft.serviceName }}</span>',
```

app/templates/frameworks/services.html
```diff
@@ -1,5 +1,4 @@
 {% extends "_base_page.html" %}
-{% import "toolkit/summary-table.html" as summary %}
```
domoscargin, line 2: 🎉

app/templates/frameworks/services.html
domoscargin: No, I think this makes sense.

Anywhere we list services (in tables or otherwise), the page looks ridiculous if the user has tons of services, but that's a problem for another day.

app/templates/frameworks/services.html
```diff
@@ -65,91 +64,115 @@ <h2 class="summary-item-heading">{{ framework.name }} is closed for applications
     </p>
   {% endif %}
 
-  {{ summary.heading("Draft services") }}
+  <h2 class="govuk-heading-m">Draft services</h2>
```
domoscargin, line 67: It'd be good to use [table `<caption>`](https://design-system.service.gov.uk/components/table/#table-captions), but we don't have a clean mechanism to do that AND hide the table if there's no data.

We've been doing this if/else thing in lots of places: https://github.com/alphagov/digitalmarketplace-briefs-frontend/blob/afedd090199322b0dfe1545b824e10e03ae7899e/app/templates/buyers/dashboard.html#L33-L39

But I'm concerned about a mix of heading tags and table captions for same-level content if there's multiple tables which might be empty, so I think headings are probably best if they directly precede the table.

app/templates/frameworks/services.html
gidsg: I had to do this a different way to make Jinja happy, but should work as intended now after 37a37a3

app/templates/frameworks/services.html
gidsg: @domoscargin thanks for the suggestion, done now in 5646f92

app/templates/frameworks/services.html
gidsg: changed to button in 9011699

app/templates/frameworks/services.html
gidsg: fixed as suggested in ff4b27d

gidsg:
@domoscargin I believe I have addressed all the comments now, please may you re-review?

app/templates/frameworks/services.html
```diff
@@ -65,91 +64,127 @@ <h2 class="summary-item-heading">{{ framework.name }} is closed for applications
...
+    </form>
+    {% endset %}
+        {% set row = ns.completed_rows.append(
+        [
+          {"html": service_link},
+          {"html": make_copy_button,
+          "classes": 'dm-app-align-right'}
+        ]
+    ) %}
+        {{ govukTable({
```
domoscargin, line 167: Need to move this out of the for loop.

app/templates/frameworks/services.html
domoscargin: Could we move the CTA directly below the `<h1>`, or do you think it's better placed where it is?

domoscargin:
A couple of small comments, but otherwise this is looks good to me.

We've done some work on the functional tests in the past to parse either a govukSummaryList or our old summary table, so hopefully they're generic enough to handle these changes (though the relevant tests don't actually run unless there's an open framework).

Local tests pass for me.

app/templates/frameworks/services.html
gidsg: good spot 👀   Fixed in 3f4fd50.

app/templates/frameworks/services.html
gidsg: Have implemented this suggestion in 78d57f0

domoscargin (approved):
Nice!

bjgill:
Small thing from me - would it be possible to explain in the commit message why we're making this change? That might help future developers.

gidsg:
@bjgill OK, I added an extended commit message in 1cca79f.

bjgill (approved):

PR #1455, opened 2021-06-21
replace summary table on previous services page with govuk frontend v3 table

gidsg:
https://trello.com/c/xcKuSRij/24-2-replace-summary-tables-of-services
We are replacing Digital Marketplace toolkit components with GOVUK frontend v3 (Design System) components as they have been accessibility tested.

## Before
Screenshot 2021-06-21 at 12 14 07

## After
Screenshot 2021-06-21 at 12 17 05

### DIFF
--- a/app/templates/services/previous_services.html
+++ b/app/templates/services/previous_services.html
@@ -84,7 +84,7 @@
 
 <h1 class="govuk-heading-l">Previous {{ lot.name|lower }} services</h1>
     {{ govukButton({
-    'text': 'Add all your services',
+    'text': 'Add all your {} services'.format(source_framework.name),
     'href': url_for('.confirm_copy_all_previous_services', framework_slug=framework.slug, lot_slug=lot.slug),
     "attributes": {
         "data-name": 'add-all-services'
--- a/app/templates/services/previous_services.html
+++ b/app/templates/services/previous_services.html
@@ -124,7 +124,7 @@
       {% endfor %}
       {{ govukTable({
         'caption': "Your services from {}".format(source_framework.name),
-        'captionClasses': "govuk-table__caption--l",
+        'captionClasses': "govuk-table__caption--m",
         'head': [
         {'text': "Service name"},
         {'html': '<span class="govuk-visually-hidden">Add to ' + framework.name + '</span>'}
--- a/app/templates/services/previous_services.html
+++ b/app/templates/services/previous_services.html
@@ -1,6 +1,4 @@
 {% extends "_base_page.html" %}
-{% import "toolkit/summary-table.html" as summary %}
-{% import "macros/submission.html" as submission %}
 
 {% block pageTitle %}
   Copy your {{ framework.name }} services – Digital Marketplace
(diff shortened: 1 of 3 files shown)


app/templates/services/previous_services.html
```diff
@@ -84,41 +82,53 @@
...
     {% include "partials/service_warning.html" %}
 
-    <h1 class="govuk-heading-l">Previous {{ lot.name|lower }} services</h1>
-
-    {{ summary.heading("Your services from {}".format(source_framework.name)) }}
-    {% if previous_services_still_to_copy|length > 1 %}
-      {{ summary.top_link("Add all your services", url_for('.confirm_copy_all_previous_services', framework_slug=framework.slug, lot_slug=lot.slug)) }}
+<h1 class="govuk-heading-l">Previous {{ lot.name|lower }} services</h1>
+    {{ govukButton({
+    'text': 'Add all your services',
```
domoscargin, line 87: I wonder if this is clear enough or whether we want it to read "Add all your {source_framework} services".

Might be a long button for "Add all your Digital Outcomes & Specialists 4 services", but I think that could only ever apply on User Research Studios, since the others all have a single limit.

app/templates/services/previous_services.html
```diff
@@ -84,41 +82,53 @@
...
+          {% set row = ns.service_rows.append(
+          [
+          {'html': service_link},
+          {'html': add_button}
+          ]
+
+          ) %}
+      {% endfor %}
+      {{ govukTable({
+        'head': [
```
domoscargin, line 126: Since this page is basically just a table, could we move the `<h2>` to the table caption? This will just more tightly couple the table with its heading.

```
'caption': "Your services from {}".format(source_framework.name),
'captionClasses': 'govuk-table__caption--m'
```

app/templates/services/previous_services.html
```diff
@@ -84,41 +82,53 @@
...
+      {% set ns = namespace(service_rows = []) %}
+      {% for service in previous_services_still_to_copy %}
+          {% set service_link %}
+            <a href="{{ url_for(".edit_service", framework_slug=source_framework.slug, service_id=service.id) }}">{{ service.serviceName }}</a>
+          {% endset %}
+          {% set add_button %}
+            <form method="post" action="{{ url_for('.copy_previous_service', framework_slug=framework.slug, lot_slug=service.lot, service_id=service.id) }}">
+                <input type="hidden" name="csrf_token" value="{{ csrf_token_value or csrf_token() }}" />
+                {{ govukButton({
+                          "text": "Add",
```
domoscargin, line 105:
```suggestion
                          "html": 'Add<span class="govuk-visually-hidden"> {} to {}</span>'.format(service.serviceName, framework.name),
```

app/templates/services/previous_services.html
```diff
@@ -1,6 +1,4 @@
 {% extends "_base_page.html" %}
-{% import "toolkit/summary-table.html" as summary %}
-{% import "macros/submission.html" as submission %}
 
 {% block pageTitle %}
   Copy your {{ framework.name }} services – Digital Marketplace
```
domoscargin, line 4: Is this an accurate title? It should probably be "Copy your {{ source_framework.name }} services"

app/templates/services/previous_services.html
gidsg: implemented in 4c3dceb

app/templates/services/previous_services.html
gidsg: good spot 👀   fixed in 1769ba7

app/templates/services/previous_services.html
```diff
@@ -84,41 +82,55 @@
...
+          [
+          {'html': service_link},
+          {'html': add_button}
+          ]
+
+          ) %}
+      {% endfor %}
+      {{ govukTable({
+        'caption': "Your services from {}".format(source_framework.name),
+        'captionClasses': "govuk-table__caption--l",
```
domoscargin, line 127: Didn't realise these modifiers were only added in v3.12 of govuk-frontend, so we'll either have to revert back to the `<h2>` or do something like https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1456

domoscargin (approved):
This looks great!

app/templates/services/previous_services.html
gidsg: Not sure if that makes sense as you're only copying from one lot?

`Add all your {source_framework} {lot.name} services` ?

domoscargin: Hmm, I think we should keep this as it is now. We can see if this impacts anything once we open these frameworks for real - my unproven thought is that it won't.

gidsg: OK, I've reverted this now in 3ebdd88.

PR #1463, opened 2021-07-08
remove unused variable

gidsg:
Spotted this unused variable during the course of another story.
There are no references to it in the rest of the codebase.


daniele-occhipinti (approved):
Good to keep the code as clean as possible!

PR #1707, opened 2023-11-08
Add build for native AWS container image support

cmurton-bg:
This PR adds Dockerfiles to build the container images used for DMP 1.0 as part of the migration to AWS.

Changes include:
- A /healthcheck endpoint is added
- Support for a config environment called 'native-aws'
- A new `Makefile` target called `build` which will, when run on a host which has the referenced base images available, create HTTP and WSGI images necessary for the Fargate tasks to run this application.


tim-s-ccs (approved):

cmurton-bg:
@tim-s-ccs should we be concerned with the failing `flake8` test or am I good to merge? It doesn't appear related to my work.

